PR #41, opened 2026-02-26
Add project scaffolding

mkolesnik:
Add minimal scaffolding to get started:
- MultiClusterMesh CRD (pkg/apis/mesh/v1alpha1)
- Build system (Makefile, Dockerfile)
- Main entry point with cobra CLI
- Tool dependencies (tools/go.mod)
- Go modules and dependencies

PR #42, opened 2026-03-25
Implement a basic controller to install the Sail/OSSM operator

mkolesnik:
This is the basic implementation for the controller, making sure to install the operator when a MultiClusterMesh CR is created, and remove the operator when it's deleted.

Right now the PR is a draft as it's being actively worked on, expect more features and cases to be added in this PR.

Depends-On: #41


jewertow (approved):
I have tested this change on ACM-enabled clusters and I can confirm that creating `MultiClusterMesh` results in installing OSSM operator in all clusters, but when I delete the `MultiClusterMesh` there remain orphaned `ManifestWork` resources. I think this is because we don't have finalizers.

I'm OK with merging it as is and addressing it in follow-up PRs.

jewertow (changes requested):
Just blocking the automatic merge

PR #49, opened 2026-04-23
Initialize controller-runtime logger to suppress warning

jewertow:
controller-runtime requires explicit logger initialization via ctrl.SetLogger(). Without it, the framework prints a warning with stack trace whenever its internal components attempt to log.


yxun (approved):

yxun:
lgtm

PR #51, opened 2026-04-24
Implement distributing certificates

jewertow:
**Summary:**

Implements basic certificate distribution:
- adds initial support for distributing intermediate CA certificates to managed clusters
- cert-manager integration remains optional - clusters without issuer configuration work as before
- `Certificate` resources are created in the `MultiClusterMesh` namespace with naming pattern `cacerts-<cluster-name>`
- secrets are distributed to managed clusters via `ManifestWork` resources

**Limitations:**

- `ManifestWork` reconciliation does not handle `Secret`/`Issuer` removal/modification
- only the namespace-scoped `Issuer` is supported (no `ClusterIssuer` support)
- does not handle cleanup of child resources

These limitations will be addressed in follow-up PRs.


pkg/hub/mesh/controller.go
```diff
@@ -383,3 +407,222 @@ func (r *Reconciler) applyOperatorDefaults(config meshv1alpha1.OperatorConfig, i
...
+			Namespace: controlPlaneNamespace,
+		},
+		Type: corev1.SecretTypeTLS,
+		Data: secret.Data,
+	}
+
+	return &workv1.ManifestWork{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      ManifestWorkNameCacerts,
+			Namespace: clusterName,
```
mkolesnik, line 684:
```suggestion
			Namespace: clusterName,
			Labels: map[string]string{
				LabelMeshName:      mesh.Name,
				LabelMeshNamespace: mesh.Namespace,
			},
```

Should also label this MW so that it gets deleted when the mesh is deleted

pkg/hub/mesh/controller.go
```diff
@@ -383,3 +407,222 @@ func (r *Reconciler) applyOperatorDefaults(config meshv1alpha1.OperatorConfig, i
...
+			Workload: workv1.ManifestsTemplate{
+				Manifests: []workv1.Manifest{{
+					RawExtension: runtime.RawExtension{Object: cacertsSecret},
+				}},
+			},
+		},
+	}
+}
+
+// updateCacertsManifestWorkIfNeeded updates the ManifestWork if the secret data has changed
```
mkolesnik, line 697: This is called and described as "update if needed", but it always tries to update.
Did you mean to add a check here that the secret has changed?

pkg/hub/mesh/controller.go
```diff
@@ -383,3 +407,222 @@ func (r *Reconciler) applyOperatorDefaults(config meshv1alpha1.OperatorConfig, i
...
+}
+
+// getCacertsName returns the name for the certificate and secret for a specific cluster
+func getCacertsName(clusterName string) string {
+	return fmt.Sprintf("cacerts-%s", clusterName)
+}
+
+// ensureCertificatesCreated creates Certificate resources for each cluster in the mesh
+func (r *Reconciler) ensureCertificatesCreated(ctx context.Context, mesh *meshv1alpha1.MultiClusterMesh, clusters []clusterv1.ManagedCluster) error {
+	if mesh.Spec.Security.Trust.CertManager.IssuerRef.Name == "" {
```
mkolesnik: You already checked it on L#151, right before calling this func.
Is it necessary?

pkg/hub/mesh/controller.go
```diff
@@ -383,3 +407,222 @@ func (r *Reconciler) applyOperatorDefaults(config meshv1alpha1.OperatorConfig, i
...
+	}
+
+	klog.Infof("Successfully created Certificate %s/%s", mesh.Namespace, certName)
+	return nil
+}
+
+// ensureCacertsDistributed creates ManifestWorks to distribute cacerts secrets to clusters
+func (r *Reconciler) ensureCacertsDistributed(ctx context.Context, mesh *meshv1alpha1.MultiClusterMesh, clusters []clusterv1.ManagedCluster) error {
+	issuerRef := mesh.Spec.Security.Trust.CertManager.IssuerRef
+	if issuerRef.Name == "" {
```
mkolesnik: You already checked it on L#151, right before calling this func.
Is it necessary?

pkg/hub/mesh/controller.go
```diff
@@ -383,3 +407,222 @@ func (r *Reconciler) applyOperatorDefaults(config meshv1alpha1.OperatorConfig, i
...
+	}
+
+	klog.Infof("Creating Certificate %s/%s for cluster %s", mesh.Namespace, certName, cluster.Name)
+
+	cert = &certmanagerv1.Certificate{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      certName,
+			Namespace: mesh.Namespace,
+			Labels: map[string]string{
+				LabelManagedBy:     "multicluster-mesh-addon",
```
mkolesnik: nit: value should be a const

pkg/hub/mesh/controller.go
```diff
@@ -383,3 +407,222 @@ func (r *Reconciler) applyOperatorDefaults(config meshv1alpha1.OperatorConfig, i
...
+				LabelMeshName:      mesh.Name,
+				LabelMeshNamespace: mesh.Namespace,
+				LabelClusterName:   cluster.Name,
+			},
+		},
+		Spec: certmanagerv1.CertificateSpec{
+			SecretName: certName,
+			SecretTemplate: &certmanagerv1.CertificateSecretTemplate{
+				Labels: map[string]string{
+					LabelManagedBy:     "multicluster-mesh-addon",
```
mkolesnik: nit: value should be a const

pkg/hub/mesh/controller.go
```diff
@@ -383,3 +407,222 @@ func (r *Reconciler) applyOperatorDefaults(config meshv1alpha1.OperatorConfig, i
...
+					LabelMeshName:      mesh.Name,
+					LabelMeshNamespace: mesh.Namespace,
+					LabelClusterName:   cluster.Name,
+				},
+			},
+			Duration:    &metav1.Duration{Duration: 1440 * time.Hour},
+			RenewBefore: &metav1.Duration{Duration: 360 * time.Hour},
```
mkolesnik: nit: perhaps we should define const for "Day"?
It would be easier to read.

pkg/hub/mesh/controller.go
```diff
@@ -76,6 +84,7 @@ func RegisterController(mgr manager.Manager) error {
 
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&meshv1alpha1.MultiClusterMesh{}).
+		Watches(&corev1.Secret{}, handler.EnqueueRequestsFromMapFunc(reconciler.mapSecretToMesh)).
```
mkolesnik: Should we use a predicate when watching the secrets?

I reckon there could be unrelated secrets which we won't want to listen to and trigger for each one.

pkg/hub/mesh/controller.go
```diff
@@ -383,3 +407,222 @@ func (r *Reconciler) applyOperatorDefaults(config meshv1alpha1.OperatorConfig, i
...
+		return nil
+	}
+
+	meshName := secret.Labels[LabelMeshName]
+	meshNamespace := secret.Labels[LabelMeshNamespace]
+	if meshName == "" || meshNamespace == "" {
+		return nil
+	}
+
+	klog.V(4).Infof("Secret %s/%s changed, triggering reconcile for mesh %s/%s",
```
mkolesnik: nit: Technically this triggers for create and delete as well, so the log "changed" isn't quite accurate.

test/integration/controller_test.go
```diff
@@ -245,6 +245,104 @@ var _ = Describe("MultiClusterMesh Controller", func() {
...
+
+			// Verify the ManifestWork contains a secret
+			Expect(work.Spec.Workload.Manifests).To(HaveLen(1))
+
+			secret := &corev1.Secret{}
+			Expect(unmarshalManifest(work.Spec.Workload.Manifests[0], secret)).To(Succeed())
+			Expect(secret.Name).To(Equal("cacerts"))
+			Expect(secret.Namespace).To(Equal("istio-system"))
+			Expect(secret.Data).To(HaveKey("tls.crt"))
+			Expect(secret.Data).To(HaveKey("tls.key"))
```
mkolesnik: Shouldn't we also check that it has "ca.crt"?

Also, do we want to make sure it's a `SecretTypeTLS`?

test/integration/controller_test.go
```diff
@@ -245,6 +245,104 @@ var _ = Describe("MultiClusterMesh Controller", func() {
 		})
 	})
 
+	Context("Certificate distribution", func() {
+		It("should create ManifestWork when cacerts secret is created", func() {
+			util.CreateK8sManagedCluster(ctx, k8sClient, clusterName, testClusterSet)
+			util.CreateMultiClusterMeshWithCertManager(ctx, k8sClient, meshName, testNs, testClusterSet, "mesh-issuer")
+
+			// Wait for operator ManifestWork to be created first
+			expectSailManifestWork(clusterName)
```
mkolesnik: I reckon you can put these in a `BeforeEach` since every test seems to need them..

For the one that has 2 clusters, just create an additional one in the test itself.

test/integration/controller_test.go
```diff
@@ -245,6 +245,104 @@ var _ = Describe("MultiClusterMesh Controller", func() {
...
+			initialResourceVersion := work.ResourceVersion
+
+			// Update the secret
+			secret := &corev1.Secret{}
+			Expect(k8sClient.Get(ctx, types.NamespacedName{
+				Name:      fmt.Sprintf("cacerts-%s", clusterName),
+				Namespace: testNs,
+			}, secret)).To(Succeed())
+
+			secret.Data["tls.crt"] = []byte("updated-cert-data")
```
mkolesnik: Don't we want to check the data itself has been updated?

test/integration/controller_test.go
```diff
@@ -245,6 +245,104 @@ var _ = Describe("MultiClusterMesh Controller", func() {
...
+			util.CreateMultiClusterMeshWithCertManager(ctx, k8sClient, meshName, testNs, testClusterSet, "mesh-issuer")
+
+			// Wait for operator ManifestWork to be created first
+			expectSailManifestWork(clusterName)
+
+			// Simulate cert-manager creating the cacerts secret
+			util.CreateCacertsSecret(ctx, k8sClient, testNs, clusterName, meshName, testNs)
+
+			// Verify cacerts ManifestWork is created
+			work := expectManifestWork(meshcontroller.ManifestWorkNameCacerts, clusterName)
```
mkolesnik: nit: `expectManifestWork(meshcontroller.ManifestWorkNameCacerts, clusterName)` can be extracted to a function since it repeats in all the cert tests

pkg/hub/mesh/controller.go
jewertow: Yeah, it's not necessary... Sorry.

Removed.

pkg/hub/mesh/controller.go
jewertow: Removed

pkg/hub/mesh/controller.go
jewertow: Done

pkg/hub/mesh/controller.go
jewertow: Done

pkg/hub/mesh/controller.go
jewertow: Yes, we definitely should use a predicate. Thank you for catching it.

Done.

pkg/hub/mesh/controller.go
jewertow: Done

pkg/hub/mesh/controller.go
jewertow: Thanks for catching it.

Done.

pkg/hub/mesh/controller.go
jewertow: > Did you mean to add a check here that the secret has changed?

Yes, I added missing checks. I will address label reconciliation in another PR that will also address watching ManifestWork with proper predicate.

pkg/hub/mesh/controller.go
jewertow: JFYI: I didn't care much about it, because I think we will need to rethink these settings before making it production-ready.

Done.

test/integration/controller_test.go
jewertow: Honestly, this less readable to me. Also, these tests follow existing convention, so it would be better to change it consitently for all tests in a follow-up PR.

test/integration/controller_test.go
jewertow: Sorry, but I do not understand what you suggest. Could you clarify what function you would like to see?

test/integration/controller_test.go
jewertow: Done

test/integration/controller_test.go
jewertow: The test verified resource version - it seemed sufficient to me, but I added the data assertion as you want.

mkolesnik: Thanks,
I agree that resourse version indicates some update happened, but it might've been some other field.

test/integration/controller_test.go
mkolesnik: Since it repeats in all tests, there could be a `expectCacertsManifestWork(clusterName)` func.

I know it doesn't save much but since the MW name is constant it makes some sense.

I leave it up to you to add it if you find it better

pkg/hub/mesh/controller.go
```diff
@@ -59,6 +72,8 @@ const (
 
 var (
 	MissingClaimRequeueDelay = 30 * time.Second
+
+	Day = 24 * time.Hour
```
mkolesnik, line 74: nit: Should be a const? Or do we expect to set it in tests or something?

mkolesnik (commented):
Thanks, overall it looks good.
I'd like to get an additional review from @sridhargaddam if he can

sridhargaddam:
> Thanks, overall it looks good. I'd like to get an additional review from @sridhargaddam if he can

I'm on it, @mkolesnik. Will share any review comments soon.

pkg/hub/mesh/controller.go
sridhargaddam: +1 we can make the Day as a const

pkg/hub/mesh/controller.go
```diff
@@ -383,3 +419,214 @@ func (r *Reconciler) applyOperatorDefaults(config meshv1alpha1.OperatorConfig, i
...
+	}
+
+	if !errors.IsNotFound(err) {
+		return fmt.Errorf("failed to get Certificate: %w", err)
+	}
+
+	klog.Infof("Creating Certificate %s/%s for cluster %s", mesh.Namespace, certName, cluster.Name)
+
+	cert = &certmanagerv1.Certificate{
+		ObjectMeta: metav1.ObjectMeta{
```
sridhargaddam, line 577: Do we need `OwnerReference` on this Certificate object?

pkg/hub/mesh/controller.go
```diff
@@ -383,3 +419,214 @@ func (r *Reconciler) applyOperatorDefaults(config meshv1alpha1.OperatorConfig, i
...
+			Name:      certName,
+			Namespace: mesh.Namespace,
+			Labels: map[string]string{
+				LabelManagedBy:     ControllerName,
+				LabelMeshName:      mesh.Name,
+				LabelMeshNamespace: mesh.Namespace,
+				LabelClusterName:   cluster.Name,
+			},
```
sridhargaddam: The same labels are used even for SecretTemplate (L493). Probably we can extract them to an helper API.

test/util/k8s.go
```diff
@@ -23,3 +26,26 @@ func CreateNamespace(ctx context.Context, k8sClient client.Client, name string)
...
 }
+
+// CreateCacertsSecret creates a TLS secret that simulates what cert-manager would create.
+func CreateCacertsSecret(ctx context.Context, k8sClient client.Client, namespace, clusterName, meshName, meshNamespace string) {
+	secret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      fmt.Sprintf("cacerts-%s", clusterName),
+			Namespace: namespace,
+			Labels: map[string]string{
+				meshcontroller.LabelManagedBy:     "multicluster-mesh-addon",
```
sridhargaddam: nit: 

```suggestion
				meshcontroller.LabelManagedBy:     meshcontroller.ControllerName,
```

pkg/hub/mesh/controller.go
```diff
@@ -383,3 +419,214 @@ func (r *Reconciler) applyOperatorDefaults(config meshv1alpha1.OperatorConfig, i
...
+		},
+	}
+}
+
+// updateCacertsManifestWorkIfNeeded updates the ManifestWork if the secret data has changed
+func (r *Reconciler) updateCacertsManifestWorkIfNeeded(ctx context.Context, mesh *meshv1alpha1.MultiClusterMesh, work *workv1.ManifestWork, secret *corev1.Secret) error {
+	newWork := r.buildCacertsManifestWork(mesh, work.Namespace, secret)
+
+	if cmp.Equal(work.Spec, newWork.Spec) {
```
sridhargaddam: Please check if we can safely use cmp.Equal to compare the secrets. If not, we should try to unmarshal them and use reflect.DeepEqual.

pkg/apis/mesh/v1alpha1/types.go
```diff
@@ -19,6 +19,19 @@ type MultiClusterMesh struct {
 	Status MultiClusterMeshStatus `json:"status,omitempty"`
 }
 
+// GetTrustDomain returns the trust domain for this mesh.
+func (m *MultiClusterMesh) GetTrustDomain() string {
+	return m.Name
```
sridhargaddam: I vaguely remember the trustDomain name to have a suffix of ".svc" in some doc

sridhargaddam (commented):
Initial pass...

pkg/hub/mesh/controller.go
jewertow: Yes, we can safely use `cmp.Equal` in this case, but to avoid this kind of concerns I replaced it with `reflect.DeepEqual` as we do not need any customization.

pkg/apis/mesh/v1alpha1/types.go
jewertow: That was likely just a placeholder in the documentation. I see no technical justification for appending `.svc` to the trust domain, and frankly, I am strongly opposed to the idea. It adds unnecessary complexity for no reason.

test/util/k8s.go
jewertow: Thanks, done.

pkg/hub/mesh/controller.go
jewertow: Done

pkg/hub/mesh/controller.go
jewertow: Done

pkg/hub/mesh/controller.go
jewertow: Yeah, we definitely need it. The only reason why I didn't add it is we were planning to make small PRs and work iteratively - I also said in the PR that this is an initial implementation.

pkg/hub/mesh/controller.go
```diff
@@ -383,3 +419,213 @@ func (r *Reconciler) applyOperatorDefaults(config meshv1alpha1.OperatorConfig, i
...
+			},
+		},
+	}
+}
+
+// updateCacertsManifestWorkIfNeeded updates the ManifestWork if the secret data has changed
+func (r *Reconciler) updateCacertsManifestWorkIfNeeded(ctx context.Context, mesh *meshv1alpha1.MultiClusterMesh, work *workv1.ManifestWork, secret *corev1.Secret) error {
+	newWork := r.buildCacertsManifestWork(mesh, work.Namespace, secret)
+
+	if reflect.DeepEqual(work.Spec, newWork.Spec) {
```
sridhargaddam: When I checked with Claude, this is what it says...

`buildCacertsManifestWork` creates a ManifestWork with RawExtension.Object populated, while the existing work read from the API has RawExtension.Raw populated (Object is nil). reflect.DeepEqual will see these as different every time, causing an unnecessary update on every reconcile. We should unmarshall before calling reflect.DeepEqual

  existingSecret := &corev1.Secret{} 
  if err := json.Unmarshal(work.Spec.Workload.Manifests[0].Raw, existingSecret); err != nil {
      return fmt.Errorf("failed to unmarshal existing manifest: %w", rr)
  }                                                                  
  
  if reflect.DeepEqual(existingSecret.Data, secret.Data) { 
      return nil                                                     
  }

sridhargaddam (commented):
One small observation, otherwise LGTM.

pkg/hub/mesh/controller.go
jewertow: I didn't notice any unnecessary update when running it on OCP.

test/integration/controller_test.go
jewertow: Ok, I refactored all tests according to you recent changes

pkg/hub/mesh/controller.go
```diff
@@ -86,6 +100,12 @@ func RegisterController(mgr manager.Manager) error {
 			&clusterv1beta2.ManagedClusterSet{},
 			handler.EnqueueRequestsFromMapFunc(reconciler.findMeshesForClusterSet),
 		).
+		Watches(&corev1.Secret{},
+			handler.EnqueueRequestsFromMapFunc(reconciler.mapSecretToMesh),
+			builder.WithPredicates(predicate.NewPredicateFuncs(func(obj client.Object) bool {
+				return obj.GetLabels()[LabelMeshName] != "" && obj.GetLabels()[LabelMeshNamespace] != ""
+			})),
+		).
```
nrfox, line 112: There can easily be lots of secrets in a cluster which can eat up a lot of resources. I think predicates only filter what secrets cause a workitem to be enqueued for this controller but it doesn't affect the informers at all. You can set some [cache options](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.24.1/pkg/cache#ByObject) when you create the manager which should limit the amount of secrets that are watched/cached.

pkg/hub/mesh/controller.go
```diff
@@ -467,3 +502,213 @@ func (r *Reconciler) applyOperatorDefaults(config meshv1alpha1.OperatorConfig, i
 
 	return config
 }
+
+// mapSecretToMesh maps a Secret to the MultiClusterMesh that owns it
+func (r *Reconciler) mapSecretToMesh(_ context.Context, obj client.Object) []reconcile.Request {
+	secret, ok := obj.(*corev1.Secret)
+	if !ok {
+		return nil
+	}
```
nrfox, line 525: This function has a [type parameter of any](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.24.1/pkg/handler#TypedMapFunc) so you can pass it the `*corev1.Secret` directly. You don't need the `client.Object` interface unless it could be any kube object.
```suggestion
func (r *Reconciler) mapSecretToMesh(_ context.Context, secret *corev1.Secret) []reconcile.Request {
```

pkg/hub/mesh/controller.go
```diff
@@ -467,3 +502,213 @@ func (r *Reconciler) applyOperatorDefaults(config meshv1alpha1.OperatorConfig, i
...
+			Namespace: mesh.Namespace,
+			Labels:    certificateLabels(mesh, cluster.Name),
+		},
+		Spec: certmanagerv1.CertificateSpec{
+			SecretName: certName,
+			SecretTemplate: &certmanagerv1.CertificateSecretTemplate{
+				Labels: certificateLabels(mesh, cluster.Name),
+			},
+			Duration:    &metav1.Duration{Duration: 60 * Day},
+			RenewBefore: &metav1.Duration{Duration: 15 * Day},
```
nrfox, line 588: Will these be configurable eventually?

pkg/hub/mesh/controller.go
```diff
@@ -467,3 +502,213 @@ func (r *Reconciler) applyOperatorDefaults(config meshv1alpha1.OperatorConfig, i
...
+			IsCA:        true,
+			Usages: []certmanagerv1.KeyUsage{
+				certmanagerv1.UsageDigitalSignature,
+				certmanagerv1.UsageKeyEncipherment,
+				certmanagerv1.UsageCertSign,
+			},
+			IssuerRef: cmmeta.IssuerReference{
+				Name:  mesh.Spec.Security.Trust.CertManager.IssuerRef.Name,
+				Kind:  "Issuer",
+				Group: "cert-manager.io",
```
nrfox, line 599: These seem like they should come from the spec. Add TODO?

pkg/hub/mesh/controller.go
```diff
@@ -467,3 +502,213 @@ func (r *Reconciler) applyOperatorDefaults(config meshv1alpha1.OperatorConfig, i
...
+	for _, cluster := range clusters {
+		if err := r.ensureCertificateForCluster(ctx, mesh, &cluster); err != nil {
+			return fmt.Errorf("failed to ensure certificate for cluster %s: %w", cluster.Name, err)
+		}
+	}
+	return nil
+}
+
+// ensureCertificateForCluster creates a Certificate resource for a specific cluster
+func (r *Reconciler) ensureCertificateForCluster(ctx context.Context, mesh *meshv1alpha1.MultiClusterMesh, cluster *clusterv1.ManagedCluster) error {
```
nrfox, line 557: A lot of this follow the same get/create/update pattern. You might simplify it but using server side apply.

pkg/hub/mesh/controller.go
```diff
@@ -86,6 +100,12 @@ func RegisterController(mgr manager.Manager) error {
 			&clusterv1beta2.ManagedClusterSet{},
 			handler.EnqueueRequestsFromMapFunc(reconciler.findMeshesForClusterSet),
 		).
+		Watches(&corev1.Secret{},
```
nrfox, line 107: Should we also watch for `ManifestWork` changes?

pkg/hub/mesh/controller.go
jewertow: I think that cache-level label filtering would affect ManifestWork reconciliation in case of a user modification, e.g. the user manually removes the label. If I had cache-level filtering, the reconciler couldn't read the secret, because `r.Get` would return Not Found.

pkg/hub/mesh/controller.go
jewertow: Yes, we should watch for `ManifestWork`, but this is out of the scope of this PR, because we also use `ManifestWork` for `Subscription` resources, so this is more general problem.

pkg/hub/mesh/controller.go
jewertow: This suggestion does not apply to the controller-runtime 0.23, which we use.

pkg/hub/mesh/controller.go
jewertow: Thanks, I will consider SSA when I address Certificate reconciliation in follow-up PRs. At this point, it would be worse, as we create the Certificate and forget about it, so requesting a patch on every reconciliation makes no sense, but I will definitely consider it in the next iteration.

pkg/hub/mesh/controller.go
jewertow: This is a very good question, and I believe we definitely need to allow some customization, but I think the design doc would be better place to discuss it.

pkg/hub/mesh/controller.go
jewertow: As you can see in the PR description, we only support the namespace-scoped `Issuer`. `ClusterIssuer` will be addressed in the future.

pkg/hub/mesh/controller.go
```diff
@@ -40,11 +46,17 @@ const (
...
 
 	FinalizerName = "mesh.open-cluster-management.io/finalizer"
 
-	ManagedByLabel = "app.kubernetes.io/managed-by"
-	ManagedByValue = "multicluster-mesh-addon"
+	ManagedByLabel     = "app.kubernetes.io/managed-by"
+	ManagedByValue     = "multicluster-mesh-addon"
+	LabelMeshName      = "mesh.open-cluster-management.io/mesh-name"
+	LabelMeshNamespace = "mesh.open-cluster-management.io/mesh-namespace"
+	LabelClusterName   = "mesh.open-cluster-management.io/cluster-name"
```
```suggestion
	ClusterNameLabel   = "mesh.open-cluster-management.io/cluster-name"
	ManagedByLabel     = "app.kubernetes.io/managed-by"
	ManagedByValue     = "multicluster-mesh-addon"
	MeshNameLabel      = "mesh.open-cluster-management.io/mesh-name"
	MeshNamespaceLabel = "mesh.open-cluster-management.io/mesh-namespace"
```

For consistency?

pkg/hub/mesh/controller.go
mkolesnik: I'm working on a change for watching MW, it should reconcile any MW managed by the addon.

There is a case to add watch for Certificates created by us, but I think this can be done in a follow up PR

pkg/hub/mesh/controller.go
mkolesnik: We could do a version bump in a follow up PR, if there are no conflicts with the other libraries we use..

pkg/hub/mesh/controller.go
mkolesnik: Agree with @jewertow - We can discuss this but for now I think it's ok to go with hard-coded values for these

pkg/hub/mesh/controller.go
mkolesnik: RN we decided to go with namespaced issuer for this, we can always add cluster issuer support in the future without breaking the API

pkg/hub/mesh/controller.go
```diff
@@ -467,3 +502,213 @@ func (r *Reconciler) applyOperatorDefaults(config meshv1alpha1.OperatorConfig, i
...
+			Labels:    certificateLabels(mesh, cluster.Name),
+		},
+		Spec: certmanagerv1.CertificateSpec{
+			SecretName: certName,
+			SecretTemplate: &certmanagerv1.CertificateSecretTemplate{
+				Labels: certificateLabels(mesh, cluster.Name),
+			},
+			Duration:    &metav1.Duration{Duration: 60 * Day},
+			RenewBefore: &metav1.Duration{Duration: 15 * Day},
+			CommonName:  "Intermediate Istio CA",
```
mkolesnik, line 589: I know you found some issues with CN, perhaps its a good idea to put a TODO comment here so it's not forgotten about

pkg/hub/mesh/controller.go
```diff
@@ -467,3 +502,213 @@ func (r *Reconciler) applyOperatorDefaults(config meshv1alpha1.OperatorConfig, i
...
+		Type: corev1.SecretTypeTLS,
+		Data: secret.Data,
+	}
+
+	return &workv1.ManifestWork{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      ManifestWorkNameCacerts,
+			Namespace: clusterName,
+			Labels: map[string]string{
+				LabelMeshName:      mesh.Name,
```
mkolesnik: Should also add the ManagedByLabel here?

Maybe just call `certificateLabels` (could be renamed to `meshOwnedLabels` perhaps?)

pkg/hub/mesh/controller.go
mkolesnik: Sridhar-Claude is correct, we even have `unmarshalManifest` in the tests for that, perhaps move it to a general package and use here as well?

You can observe it with the following integ test:
```golang
It("should populate Raw (not Object) when reading ManifestWork from API", func() {
      util.CreateK8sManagedCluster(ctx, k8sClient, clusterName, testClusterSet)
      util.CreateMultiClusterMesh(ctx, k8sClient, meshName, testNs, testClusterSet,
  meshv1alpha1.OperatorConfig{})

      work := expectOperatorManifestWork(clusterName)

      manifest := work.Spec.Workload.Manifests[0]
      Expect(manifest.RawExtension.Raw).NotTo(BeNil(), "Raw should be populated when read from API")
      Expect(manifest.RawExtension.Object).To(BeNil(), "Object should be nil when read from API")
  })
```

test/integration/controller_test.go
```diff
@@ -322,6 +323,80 @@ var _ = Describe("MultiClusterMesh Controller", func() {
 		})
 	})
 
+	Context("Certificate distribution", func() {
+		When("cert-manager issuer is configured", func() {
```
mkolesnik, line 328: Actually these whould chec that `Certificate` resources get created, no?

Another context should check when we simulate CA secrets created by cert-manager, which would be these tests

pkg/hub/mesh/controller.go
```diff
@@ -40,11 +46,17 @@ const (
 	DefaultChannel = "stable"
 
 	OperatorManifestWorkName = "multicluster-mesh-operator"
+	ManifestWorkNameCacerts  = "multicluster-mesh-cacerts"
```
mkolesnik, line 52: This would cause two meshes targeting same cluster to "fight" over the resource, even though each should get its own MW..

pkg/hub/mesh/controller.go
jewertow: Done

pkg/hub/mesh/controller.go
jewertow: Done

pkg/hub/mesh/controller.go
jewertow: Thanks for challenging me to understand it better. I read the docs, run the test and seems that you're both right. Still don't know why I didn't observe that when testing on OCP. Anyway... I changed it according to your suggestion.

test/integration/controller_test.go
jewertow: > Actually these whould chec that Certificate resources get created, no?

Yes, sure. Thanks for catching that! Vibe coding is hard...

> Another context should check when we simulate CA secrets created by cert-manager,

Simulating cert-manager is just creating the secret in the API. Why would you add another context for that?

pkg/hub/mesh/controller.go
jewertow: Will create an issue for better visibility.

pkg/hub/mesh/controller.go
jewertow: Multi-tenancy, right... Let me address it in a dedicated PR, because that may open another long discussion. Doing this fully correctly will require calculating hash and including readable name and namespace for debugging, and also keeping the length limit. We might also consider reusing the same ManifestResource for different tenats in the same cluster.

pkg/hub/mesh/controller.go
jewertow: https://github.com/stolostron/multicluster-mesh-addon/issues/65

pkg/hub/mesh/controller.go
jewertow: https://github.com/stolostron/multicluster-mesh-addon/issues/66

pkg/hub/mesh/controller.go
jewertow: https://github.com/stolostron/multicluster-mesh-addon/issues/67

pkg/hub/mesh/controller.go
nrfox: Applying the labels should be handled by reconciliation of `ManifestWork` objects but also if the user is manually removing labels then they are probably breaking things. Feel free to just add a `TODO`.

pkg/hub/mesh/controller.go
nrfox: It's been there since v0.18: https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.23.0/pkg/handler#TypedMapFunc

nrfox: For all my feedback feel free to just add `TODO`

mkolesnik: We have other mapper functions as well, I suggest to handle it in a follow up which will also cover those

jewertow: Ok, sorry, but I still cannot simply change `EnqueueRequestsFromMapFunc` to `TypedEnqueueRequestsFromMapFunc` . You can see here what changes it involves and why I didn't want to include it in this PR: https://github.com/stolostron/multicluster-mesh-addon/pull/68. If that can be done simpler please continue discussion in that PR.

test/integration/controller_test.go
mkolesnik: Yes I meant for test code organization this part can be split in its own context. I'm fine with doing that in a follow up PR

mkolesnik (commented):
Looks good to me overall.
Please make sure to run the make targets manually as we don't have CI yet.

If all's OK, I'm fine with this going in if other reviewers are finished with their review.

pkg/hub/mesh/controller.go
jewertow: https://github.com/stolostron/multicluster-mesh-addon/issues/69

pkg/hub/mesh/controller.go
jewertow: https://github.com/stolostron/multicluster-mesh-addon/issues/71

pkg/hub/mesh/controller.go
jewertow: https://github.com/stolostron/multicluster-mesh-addon/issues/72

pkg/hub/mesh/controller.go
jewertow: > they are probably breaking things

Yes, that's a breaking modification, and therefore it should be recovered by the controller. If we added the suggested cache-level filtering, the controller would not be able to recover from that state. So why would you want to do it in this case?

test/integration/controller_test.go
jewertow: So you suggest to have 2 contexts: first for checking that Certificate resource is created by the controller, and second for checking that the controller creates ManifestWork when Secret exists - do I understand it correctly? If so, why the current flow is wrong if it checks all of the above, but in one context representing the entire flow?

mkolesnik: From an integration test perspective I think it's a better separation of concerns.

The integration tests are white-box so they know what is expected to happen, and as such make more sense to test the different parts in higher resolution (e.g. more tests == better) so that if any part fails it's easier to pinpoint why.

I think an approach of "test the entire flow" is more suitable for E2E tests where you feed the input mesh resource and expect cacert secret to appear on the target cluster, you don't really care how the controller goes about implementing that distribution.

As I said, I'm fine with doing it in a follow up to unblock this, but I think a more granular approach in integration tests is better for quickly determining bugs.

pkg/hub/mesh/controller.go
jewertow: I create an issue and I think we should discuss it further there: https://github.com/stolostron/multicluster-mesh-addon/issues/79

test/integration/controller_test.go
jewertow: Ok, thank you for explaining your expectations.

pkg/hub/mesh/controller.go
```diff
@@ -483,3 +516,215 @@ func (r *Reconciler) applyOperatorDefaults(config meshv1alpha1.OperatorConfig, i
...
+	cacertsSecret := &corev1.Secret{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "v1",
+			Kind:       "Secret",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      CacertsSecretName,
+			Namespace: mesh.GetControlPlaneNamespace(),
+		},
+		Type: corev1.SecretTypeTLS,
```
sridhargaddam, line 677: If we create a ManifestWork with SecretTypeTLS and the same is created on the ManagedCluster, how does it work with Istio? AFAICT, the Istio's plugin CA config requires ca-cert.pem, ca-key.pem,  cert-chain.pem, and root-cert.pem.

jewertow: Istio supports both formats.

sridhargaddam (approved):
LGTM, thanks @jewertow

PR #52, opened 2026-04-26
Add Finalizer support to deletion

mkolesnik:
This ensures the controller cleans up the created resources before the MultiClusterMesh CR is deleted.

### DIFF
--- a/pkg/hub/mesh/controller.go
+++ b/pkg/hub/mesh/controller.go
@@ -18,6 +18,7 @@ import (
 	workv1 "open-cluster-management.io/api/work/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
@@ -40,6 +41,8 @@ const (
 
 	ManifestWorkNameOSSM = "multicluster-mesh-operator-ossm"
 	ManifestWorkNameSail = "multicluster-mesh-operator-sail"
+
+	FinalizerName = "mesh.open-cluster-management.io/finalizer"
 
 	clusterClaimProduct = "product.open-cluster-management.io"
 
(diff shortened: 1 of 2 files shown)


pkg/hub/mesh/controller.go
```diff
@@ -132,10 +144,15 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
+	if !controllerutil.ContainsFinalizer(mesh, FinalizerName) {
+		klog.V(4).Infof("MultiClusterMesh %s/%s has no finalizer, nothing to clean up", mesh.Namespace, mesh.Name)
+		return reconcile.Result{}, nil
+	}
+
 	klog.Infof("Handling deletion for MultiClusterMesh %s/%s", mesh.Namespace, mesh.Name)
 	clusters, err := r.getClustersFromSet(ctx, mesh.Spec.ClusterSet)
-	if err != nil {
-		return reconcile.Result{}, fmt.Errorf("failed to get clusters for cleanup: %w", err)
+	if err != nil && !errors.IsNotFound(err) {
```
jewertow: I think this condition may cause leaving orphaned resources in the following scenario:
1. Create a cluster set.
2. Create a MultiClusterMesh with the cluster set.
3. Delete the cluster set.
4. Delete the MultiClusterMesh.

If I correctly understand this function the ManifestWork resources will be left, because when the cluster set is not found it skips clusters cleanup and jumps to removing finalizer.

Could you add a test case for this scenario?

mkolesnik: Good catch, I'll change the logic so that we track the created ManifestWorks directly.
This way we're always deleting them in case the mesh resource is deleted, regardless of what happens with the ClusterSet.

mkolesnik: I'll add a listener to ClusterSet later so that the controller also reacts to cases when it gets deleted immediately, so this case shouldn't happen, but tracking by labels will be more robust anyways and will avoid potential race conditions.

pkg/hub/mesh/controller.go
```diff
@@ -132,40 +147,41 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
-			}
-			return reconcile.Result{}, fmt.Errorf("failed to get ManifestWork %s/%s: %w", cluster.Name, workName, err)
-		}
+	if err := r.List(ctx, workList, labelSelector); err != nil {
+		return reconcile.Result{}, fmt.Errorf("failed to list ManifestWorks for cleanup: %w", err)
+	}
 
+	klog.V(4).Infof("Found %d ManifestWorks to clean up for MultiClusterMesh %s/%s", len(workList.Items), mesh.Namespace, mesh.Name)
+	for i := range workList.Items {
+		work := &workList.Items[i]
```
jewertow, line 168: nit: couldn't it be simpler?
```suggestion
	for _, work := range workList.Items {
```

jewertow (approved):

PR #55, opened 2026-04-29
[Draft] Implement ManagedServiceAccount resources reconciliation and tests

yxun:
This PR has been broken into the following small PRs. All following comments will be addressed in the small PRs.
I will close this original one when all the small PRs are merged.

https://github.com/stolostron/multicluster-mesh-addon/pull/126
https://github.com/stolostron/multicluster-mesh-addon/pull/127
https://github.com/stolostron/multicluster-mesh-addon/pull/130
https://github.com/stolostron/multicluster-mesh-addon/pull/189
https://github.com/stolostron/multicluster-mesh-addon/pull/215

~This PR adds implementations of ManagedServiceAccount reconciliation and tests in the controller~

~It implements creation and cleanup of the ManagedServiceAccount related resources for each managed cluster in the ClusterSet.~ 

~It builds istio remote Secret resources and creates ManifestWork resources for distributing those Secrets so that istiod control plane can discover remote endpoints from remote cluster API server(s)~

References:
- https://docs.redhat.com/en/documentation/red_hat_advanced_cluster_management_for_kubernetes/2.16/html/clusters/cluster_mce_overview#managed-serviceaccount-addon
- https://github.com/istio/istio/blob/master/istioctl/pkg/multicluster/remote_secret.go


pkg/hub/mesh/controller.go
```diff
@@ -23,6 +23,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	meshv1alpha1 "github.com/stolostron/multicluster-mesh-addon/pkg/apis/mesh/v1alpha1"
+	msav1alpha1 "open-cluster-management.io/managed-serviceaccount/apis/authentication/v1alpha1"
```
mkolesnik: We should use v1beta1.

I know you were worried about the `rotation.enabled` field being deprecated, but the `rotation.validity` is still exposed AFAIK so there's no reason to use the older API.

pkg/hub/mesh/controller.go
```diff
@@ -383,3 +390,31 @@ func (r *Reconciler) applyOperatorDefaults(config meshv1alpha1.OperatorConfig, i
 
 	return config
 }
+
+// CreateManagedServiceAccount creates a ManagedServiceAccount resource for a cluster in the ClusterSet
+func (r *Reconciler) CreateManagedServiceAccount(ctx context.Context, cluster clusterv1.ManagedCluster, mesh *meshv1alpha1.MultiClusterMesh) error {
+	klog.Infof("Creating a ManagedServiceAccount for a managed cluster: %s", cluster.Name)
+	validity, err := time.ParseDuration(mesh.Spec.Security.Discovery.TokenValidity) // this is derived from the MultiClusterMesh spec.security.discovery.tokenValidity
```
mkolesnik: Hmm right now in our API the `TokenValidity` field is defined as a sting that 
```
	// Supports hours (h), days (d), weeks (w), or months (m)
```

This parsing won't work well, but I wonder if we should just change the type of `TokenValidity` to be `metav1.Duration`?

pkg/hub/mesh/controller.go
```diff
@@ -383,3 +390,31 @@ func (r *Reconciler) applyOperatorDefaults(config meshv1alpha1.OperatorConfig, i
...
+	if err != nil {
+		return fmt.Errorf("failed to parse validity duration: %w", err)
+	}
+
+	msa := &msav1alpha1.ManagedServiceAccount{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      fmt.Sprintf("%s-%s", cluster.Name, mesh.Name), // Naming convention: <cluster-name>-<mesh-name>
+			Namespace: cluster.Name,
```
mkolesnik: You should add labels to the resource so it can be properly deleted.
If you're adding deletion later, then I'm fine with not adding them right now.

pkg/hub/mesh/controller.go
```diff
@@ -383,3 +390,31 @@ func (r *Reconciler) applyOperatorDefaults(config meshv1alpha1.OperatorConfig, i
...
+		},
+		Spec: msav1alpha1.ManagedServiceAccountSpec{
+			Rotation: msav1alpha1.ManagedServiceAccountRotation{
+				Enabled:  false,
```
mkolesnik: Why `false`?

pkg/hub/mesh/controller.go
```diff
@@ -141,6 +140,14 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		return reconcile.Result{RequeueAfter: MissingClaimRequeueDelay}, nil
 	}
 
+	// Create ManagedServiceAccount resources for each managed cluster
+	for _, cluster := range clusters {
+		if err := r.CreateManagedServiceAccount(ctx, cluster, mesh); err != nil {
```
mkolesnik: What if it already exists?
I reckon this'll fail?

test/integration/external-crds/ocm/authentication.open-cluster-management.io_managedserviceaccounts.yaml
```diff
@@ -0,0 +1,314 @@
+---
```
mkolesnik: This filed shouldn't be checked in

test/integration/suite_test.go
```diff
@@ -57,6 +58,7 @@ var _ = BeforeSuite(func() {
 		CRDDirectoryPaths: []string{
 			filepath.Join("..", "..", "config", "crd"),                      // Custom MultiClusterMesh CRD
 			filepath.Join("..", "..", "test", "integration", "crds", "ocm"), // OCM CRDs
+			filepath.Join("..", "..", "test", "integration", "external-crds", "ocm"), // OCM ManagedServiceAccount CRD
```
mkolesnik: I reckon we can just keep them in the same directory as other OCM test crds, or at most in a subdirectory there

test/integration/controller_test.go
```diff
@@ -193,6 +196,28 @@ var _ = Describe("MultiClusterMesh Controller", func() {
...
+			time.Sleep(2 * time.Second)
+
+			list := &unstructured.UnstructuredList{}
+			list.SetGroupVersionKind(schema.GroupVersionKind{
+				Group:   "authentication.open-cluster-management.io",
+				Kind:    "ManagedServiceAccount",
+				Version: "v1alpha1",
+			})
+
+			Expect(k8sClient.List(context.Background(), list, client.InNamespace(cluster1))).To(Succeed())
```
mkolesnik: You should add test that validity gets passed if set by user

mkolesnik (commented):
I see this only handles creating MSA but not updating or deleting it.
Is this planned to be handled later, or will you be updating this PR?

pkg/hub/mesh/controller.go
yxun: I have updated this with the v1beta1 version.

yxun:
> I see this only handles creating MSA but not updating or deleting it. Is this planned to be handled later, or will you be updating this PR?

I have updated the part of the controller code. If the cluster has an existing ManagedServiceAccount resource with the naming convention , it will skip creating or updating it.

I can add a TODO and will implement updating existing resources in a follow up PR.

pkg/hub/mesh/controller.go
yxun: Yes, the CreateManagedServiceAccount call would fail if the resource already exists. I have updated this part by checking an existing resource first.

test/integration/external-crds/ocm/authentication.open-cluster-management.io_managedserviceaccounts.yaml
yxun: I have updated the Makefile target and removed this additional crd yaml file in this PR.

test/integration/suite_test.go
yxun: right, the Makefile update-test-crds target is pulling that crd yaml in the same ocm test crds directory now.

pkg/hub/mesh/controller.go
yxun: I think so. Changing that TokenValidity type to be metav1.Duration will help aligning it with the following struct field.
https://pkg.go.dev/open-cluster-management.io/managed-serviceaccount@v0.10.0/apis/authentication/v1beta1#ManagedServiceAccountRotation

I will update this in the mesh/v1alpha1 types.

pkg/hub/mesh/controller.go
yxun: > // Enabled prescribes whether the ServiceAccount token will be rotated before it expires.
	// Deprecated: All ServiceAccount tokens will be rotated before they expire regardless of this field.

I thought we don't want to rotate that ServiceAccount token. Anyway, I can set that to true or not set it. The default value is true.

pkg/apis/mesh/v1alpha1/types.go
```diff
@@ -116,9 +116,10 @@ type IssuerReference struct {
 type DiscoveryConfig struct {
 	// TokenValidity defines how long discovery tokens are valid
 	// Supports hours (h), days (d), weeks (w), or months (m)
+	// It is a wrapper around time.Duration which supports correct marshaling to YAML and JSON
 	// +optional
 	// +kubebuilder:default="1m"
```
mkolesnik: Shouldn't default be updated?

(1m == 1 month, but feel free to use a different default if you think its more sane)

pkg/hub/mesh/controller.go
```diff
@@ -26,6 +28,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	meshv1alpha1 "github.com/stolostron/multicluster-mesh-addon/pkg/apis/mesh/v1alpha1"
+	msav1v1beta1 "open-cluster-management.io/managed-serviceaccount/apis/authentication/v1beta1"
```
```suggestion
	msav1beta1 "open-cluster-management.io/managed-serviceaccount/apis/authentication/v1beta1"
```

Did you mean this?

pkg/hub/mesh/controller.go
```diff
@@ -156,6 +159,26 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		return reconcile.Result{}, err
 	}
 
+	// Create ManagedServiceAccount resources for each managed cluster
+	msaList := &unstructured.UnstructuredList{}
```
mkolesnik: Why use unstructured? 

You should register the scheme (in main.go) and use the typed structure directly

pkg/hub/mesh/controller.go
```diff
@@ -156,6 +159,26 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
+	msaList.SetGroupVersionKind(schema.GroupVersionKind{
+		Group:   "authentication.open-cluster-management.io",
+		Kind:    "ManagedServiceAccount",
+		Version: "v1beta1",
+	})
+
+	for _, cluster := range clusters {
+		if err := r.List(ctx, msaList, client.InNamespace(cluster.Name),
```
mkolesnik: Why not simply use `client.Get`?

e.g.
```
  if err := r.Get(ctx, types.NamespacedName{Name: msaName, Namespace: cluster.Name}, existing) ...
```

pkg/hub/mesh/controller.go
```diff
@@ -476,3 +499,31 @@ func (r *Reconciler) applyOperatorDefaults(config meshv1alpha1.OperatorConfig, i
 
 	return config
 }
+
+// CreateManagedServiceAccount creates a ManagedServiceAccount resource for a cluster in the ClusterSet
+func (r *Reconciler) CreateManagedServiceAccount(ctx context.Context, cluster clusterv1.ManagedCluster, mesh *meshv1alpha1.MultiClusterMesh) error {
```
```suggestion
func (r *Reconciler) createManagedServiceAccount(ctx context.Context, cluster clusterv1.ManagedCluster, mesh *meshv1alpha1.MultiClusterMesh) error {
```

No reason to make it public

pkg/hub/mesh/controller.go
```diff
@@ -476,3 +499,31 @@ func (r *Reconciler) applyOperatorDefaults(config meshv1alpha1.OperatorConfig, i
...
+
+	msa := &msav1v1beta1.ManagedServiceAccount{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      fmt.Sprintf("%s-%s", cluster.Name, mesh.Name), // Naming convention: <cluster-name>-<mesh-name>
+			Namespace: cluster.Name,
+			Labels:    map[string]string{ManagedByLabel: ManagedByValue},
+		},
+		Spec: msav1v1beta1.ManagedServiceAccountSpec{
+			Rotation: msav1v1beta1.ManagedServiceAccountRotation{
+				// Field Enabled default=true. It prescribes the ServiceAccount token will be rotated before it expires.
```
```suggestion
```

I'm not sure this commment adds any useful info

test/integration/suite_test.go
```diff
@@ -24,6 +24,7 @@ import (
 
 	meshv1alpha1 "github.com/stolostron/multicluster-mesh-addon/pkg/apis/mesh/v1alpha1"
 	meshcontroller "github.com/stolostron/multicluster-mesh-addon/pkg/hub/mesh"
+	msav1v1beta1 "open-cluster-management.io/managed-serviceaccount/apis/authentication/v1beta1"
```
mkolesnik:
```suggestion
	msav1beta1 "open-cluster-management.io/managed-serviceaccount/apis/authentication/v1beta1"
```

Here also

test/util/ocm.go
```diff
@@ -8,6 +8,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	clusterv1 "open-cluster-management.io/api/cluster/v1"
 	clusterv1beta2 "open-cluster-management.io/api/cluster/v1beta2"
+	msav1v1beta1 "open-cluster-management.io/managed-serviceaccount/apis/authentication/v1beta1"
```
mkolesnik:
```suggestion
	msav1beta1 "open-cluster-management.io/managed-serviceaccount/apis/authentication/v1beta1"
```

Here also

test/integration/controller_test.go
```diff
@@ -225,6 +229,59 @@ var _ = Describe("MultiClusterMesh Controller", func() {
...
+		It("should create ManagedServiceAccount resources for each cluster in the ClusterSet", func() {
+			cluster1 := util.UniqueName("cluster")
+			cluster2 := util.UniqueName("cluster")
+
+			util.CreateK8sManagedCluster(ctx, k8sClient, cluster1, testClusterSet)
+			util.CreateK8sManagedCluster(ctx, k8sClient, cluster2, testClusterSet)
+			util.CreateMultiClusterMesh(ctx, k8sClient, meshName, testNs, testClusterSet, meshv1alpha1.OperatorConfig{})
+			awaitReconcileFinished()
+
+			list := &unstructured.UnstructuredList{}
```
mkolesnik: Here also, no need to use unstructured, especially since you registered the type in the test schema

test/integration/controller_test.go
```diff
@@ -225,6 +229,59 @@ var _ = Describe("MultiClusterMesh Controller", func() {
...
+			list.SetGroupVersionKind(schema.GroupVersionKind{
+				Group:   "authentication.open-cluster-management.io",
+				Kind:    "ManagedServiceAccount",
+				Version: "v1beta1",
+			})
+
+			specData := map[string]interface{}{
+				"rotation": map[string]interface{}{
+					"validity": "24h0m",
```
mkolesnik: You set it but didn't check it.

Also I think setting custom fields is better checked in a separate test (`It`)

pkg/hub/mesh/controller.go
```diff
@@ -476,3 +499,31 @@ func (r *Reconciler) applyOperatorDefaults(config meshv1alpha1.OperatorConfig, i
...
 	return config
 }
+
+// CreateManagedServiceAccount creates a ManagedServiceAccount resource for a cluster in the ClusterSet
+func (r *Reconciler) CreateManagedServiceAccount(ctx context.Context, cluster clusterv1.ManagedCluster, mesh *meshv1alpha1.MultiClusterMesh) error {
+	klog.Infof("Creating a ManagedServiceAccount for a managed cluster: %s", cluster.Name)
+
+	msa := &msav1v1beta1.ManagedServiceAccount{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      fmt.Sprintf("%s-%s", cluster.Name, mesh.Name), // Naming convention: <cluster-name>-<mesh-name>
```
mkolesnik: I think a better convention would be something like `<mesh-namespace>-<mesh-name>-istio-reader`

Either way `cluster name` is stuttering since the ns is already the cluster name in this case.

We can further discuss naming convention on #72 as well.

test/integration/controller_test.go
```diff
@@ -225,6 +229,59 @@ var _ = Describe("MultiClusterMesh Controller", func() {
...
+				},
+			}
+			unstructured.SetNestedField(list.Object, specData, "spec")
+
+			Expect(k8sClient.List(context.Background(), list, client.InNamespace(cluster1))).To(Succeed())
+			Expect(k8sClient.List(context.Background(), list, client.InNamespace(cluster2))).To(Succeed())
+		})
+
+		When("the ManagedServiceAccount resource exists", func() {
```
mkolesnik: I don't think this makes sense as an integration test.

We don't expect MSA resource to be created by the user externally, which is what this test implies.

We expect them to exist in concurrent reconciliation of the same mesh, in which case the correct tests should test that mesh is updated.

test/integration/controller_test.go
```diff
@@ -225,6 +229,59 @@ var _ = Describe("MultiClusterMesh Controller", func() {
...
+
+			specData := map[string]interface{}{
+				"rotation": map[string]interface{}{
+					"validity": "24h0m",
+				},
+			}
+			unstructured.SetNestedField(list.Object, specData, "spec")
+
+			Expect(k8sClient.List(context.Background(), list, client.InNamespace(cluster1))).To(Succeed())
+			Expect(k8sClient.List(context.Background(), list, client.InNamespace(cluster2))).To(Succeed())
```
mkolesnik: These don't really test that the MSA were created as expected.

For all we know, they might've been created by something else, or by the controller but not as we expected them

pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -0,0 +1,134 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/klog/v2"
+	addonv1beta1 "open-cluster-management.io/api/addon/v1beta1"
+	clusterv1 "open-cluster-management.io/api/cluster/v1"
+	msav1beta1 "open-cluster-management.io/managed-serviceaccount/apis/authentication/v1beta1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// deployManagedServiceAccountAddon deploys the ManagedServiceAccount add-on for each cluster in the ClusterSet
+func (r *Reconciler) deployManagedServiceAccountAddon(ctx context.Context, clusters []clusterv1.ManagedCluster) error {
```
mkolesnik: I'm not sure this concern should be handled by us.

This is an external concern for the cluster admin to handle, we just need to create MSA resources.
If we start installing other addons, it would be similar to an operator installing other operators.

Makefile
```diff
@@ -118,10 +118,22 @@ update-test-crds: ## Update test CRDs from OCM API and cert-manager dependencies
 	fi; \
 	mkdir -p $(TEST_CRD_DIR)/ocm; \
 	echo "Copying CRDs from $$OCM_API_PATH..."; \
+	cp -v $$OCM_API_PATH/addon/v1beta1/*.crd.yaml $(TEST_CRD_DIR)/ocm/ 2>/dev/null || true; \
```
mkolesnik: This shouldn't be part of our concerns..

Makefile
```diff
@@ -118,10 +118,22 @@ update-test-crds: ## Update test CRDs from OCM API and cert-manager dependencies
...
+	@OCP_MSA_PATH=$$(go list -m -f '{{.Dir}}' open-cluster-management.io/managed-serviceaccount 2>/dev/null); \
+	if [ -z "$$OCP_MSA_PATH" ]; then \
+		echo "Error: open-cluster-management.io/managed-serviceaccount not found in go.mod"; \
+		echo "Run: go mod download open-cluster-management.io/managed-serviceaccount"; \
+		exit 1; \
+	fi; \
+	mkdir -p $(TEST_CRD_DIR)/ocm; \
+	echo "Copying CRDs from $$OCP_MSA_PATH..."; \
+	cp -v $$OCP_MSA_PATH/charts/managed-serviceaccount/crds/*.yaml $(TEST_CRD_DIR)/ocm/ 2>/dev/null || true; \
+	echo "Test CRDs updated successfully in $(TEST_CRD_DIR)/ocm/"
```
mkolesnik:
```suggestion
	echo "Test CRDs for MSA updated successfully in $(TEST_CRD_DIR)/ocm/"
```

main.go
```diff
@@ -46,6 +46,8 @@ func init() {
 	utilruntime.Must(operatorsv1.AddToScheme(runtimeScheme))
 	utilruntime.Must(operatorsv1alpha1.AddToScheme(runtimeScheme))
 	utilruntime.Must(certmanagerv1.AddToScheme(runtimeScheme))
+	utilruntime.Must(msav1beta1.AddToScheme(runtimeScheme))
+	utilruntime.Must(addonv1beta1.Install(runtimeScheme))
```
mkolesnik: This shouldn't be part of our concerns..

pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -0,0 +1,134 @@
...
+	if len(clusters) == 0 {
+		klog.Info("The ClusterSet has no managed cluster")
+		return nil
+	}
+
+	klog.Info("Creating ManagedServiceAccount resources for each managed cluster in the ClusterSet")
+	for _, cluster := range clusters {
+		existing := &msav1beta1.ManagedServiceAccount{}
+		// Naming convention: <cluster-name>-<mesh-name>-istio-reader
+		msaName := fmt.Sprintf("%s-%s-istio-reader", cluster.Name, mesh.Name)
```
mkolesnik: The cluster name is redundant since the MSA lives in the cluster namespace.
Also we should discuss better naming convention on #72.

test/integration/controller_test.go
```diff
@@ -241,6 +242,27 @@ var _ = Describe("MultiClusterMesh Controller", func() {
...
+			util.CreateOCPManagedCluster(ctx, k8sClient, cluster3, testClusterSet, meshcontroller.ProductOCP)
+
+			util.CreateMultiClusterMesh(ctx, k8sClient, meshName, testNs, testClusterSet, meshv1alpha1.OperatorConfig{})
+			awaitReconcileFinished()
+
+			expectManagedServiceAccount(fmt.Sprintf("%s-%s-istio-reader", cluster2, meshName), cluster2)
+			expectManagedServiceAccount(fmt.Sprintf("%s-%s-istio-reader", cluster3, meshName), cluster3)
+		})
+
+		It("should create a ManagedServiceAccount resource with the token validity field", func() {
```
mkolesnik: This test creates the MSA externally rather than through the controller.
It should create a mesh with a custom `tokenValidity`, wait for reconciliation, then verify the created MSA has the expected `rotation.validity`.

test/integration/controller_test.go
```diff
@@ -604,6 +638,49 @@ func expectCacertsSecret(work *workv1.ManifestWork) {
...
 
+func expectManagedServiceAccount(name, namespace string) *msav1beta1.ManagedServiceAccount {
+	msa := &msav1beta1.ManagedServiceAccount{}
+	Eventually(func() error {
+		return k8sClient.Get(ctx, types.NamespacedName{
+			Name:      name,
+			Namespace: namespace,
+		}, msa)
+	}).Should(Succeed())
+	Expect(msa.ObjectMeta.Labels).To(HaveKeyWithValue(meshcontroller.ManagedByLabel, meshcontroller.ManagedByValue))
```
mkolesnik: This should be outside the expect..
It should return the created MSA and specific content checks should be done by the tests themselves

pkg/hub/mesh/controller.go
```diff
@@ -801,46 +824,63 @@ func (r *Reconciler) ensureCertificateForCluster(ctx context.Context, mesh *mesh
...
 		}
 	}
 	return nil
 }
 
-// ensureCacertsManifestWork creates a ManifestWork to distribute the cacerts secret to a cluster
-func (r *Reconciler) ensureCacertsManifestWork(ctx context.Context, mesh *meshv1alpha1.MultiClusterMesh, cluster *clusterv1.ManagedCluster) error {
-	secretName := getCacertsName(cluster.Name)
+// ensureSecretManifestWork creates a ManifestWork to distribute the secret to a cluster
+func (r *Reconciler) ensureSecretManifestWork(ctx context.Context, secretName string, mesh *meshv1alpha1.MultiClusterMesh, cluster *clusterv1.ManagedCluster) error {
```
mkolesnik: This function uses `buildCacertsManifestWork` which hardcodes `CacertsSecretName` and `SecretTypeTLS`, so MSA secrets would end up named "cacerts" with type TLS on the spoke.
You should extract a common function that creates a `ManifestWork` for secrets, parameterized by name/type/namespace (maybe just send it a `Secret` to create if it's easier), then call it with the right values for "cacerts" and "MSA" respectively.

As a future improvement, consider putting all secrets for a cluster into a single `ManifestWork` (using sdk-go's [`WorkBuilder`](https://github.com/open-cluster-management-io/sdk-go/tree/main/pkg/apis/work/v1#workbuilder)) instead of one MW per secret.
More efficient, fewer API calls.
But that should probably be done as a follow-up PR.

pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -0,0 +1,134 @@
...
+				klog.Infof("Successfully created a ManagedServiceAccount %s/%s", cluster.Name, msaName)
+			}
+		}
+	}
+
+	return nil
+}
+
+// deleteManagedServiceAccounts deletes ManagedServiceAccount resources managed by multicluster-mesh-addon
+func (r *Reconciler) deleteManagedServiceAccounts(ctx context.Context) error {
```
mkolesnik: This deletes all MSAs with the managed-by label, not just the ones for the mesh being deleted. If two meshes target overlapping clusters, deleting one mesh cleans up the other's MSAs.
Scope the deletion to the current mesh, or use the `getClustersNeededByAnyMesh()` pattern like `cleanupManifestWorks` does.

Please add integ tests to avoid such problems in the future.

pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -0,0 +1,134 @@
...
+
+	klog.Info("Creating ManagedServiceAccount resources for each managed cluster in the ClusterSet")
+	for _, cluster := range clusters {
+		existing := &msav1beta1.ManagedServiceAccount{}
+		// Naming convention: <cluster-name>-<mesh-name>-istio-reader
+		msaName := fmt.Sprintf("%s-%s-istio-reader", cluster.Name, mesh.Name)
+		if err := r.Get(ctx, types.NamespacedName{Name: msaName, Namespace: cluster.Name}, existing); err == nil {
+			klog.Infof("Cluster %s has an existing ManagedServiceAccount resource %s, skipping createManagedServiceAccount", cluster.Name, msaName)
+			continue
+		} else {
```
mkolesnik: The `else` here runs for any Get error, not just NotFound.
A timeout or RBAC error would silently fall through to Create.

```suggestion
		if err := r.Get(ctx, types.NamespacedName{Name: msaName, Namespace: cluster.Name}, existing); err == nil {
			klog.Infof("Cluster %s has an existing ManagedServiceAccount resource %s, skipping createManagedServiceAccount", cluster.Name, msaName)
			continue
		} else if !apierrors.IsNotFound(err) {
			return fmt.Errorf("failed to check ManagedServiceAccount %s/%s: %w", cluster.Name, msaName, err)
		}
```

pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+
+	klog.Info("Creating ManagedServiceAccount resources for each managed cluster in the ClusterSet")
+	for _, cluster := range clusters {
+		existing := &msav1beta1.ManagedServiceAccount{}
+		// Naming convention: <cluster-name>-<mesh-name>-istio-reader
+		msaName := fmt.Sprintf("%s-%s-istio-reader", cluster.Name, mesh.Name)
+		if err := r.Get(ctx, types.NamespacedName{Name: msaName, Namespace: cluster.Name}, existing); err == nil {
+			klog.Infof("Cluster %s has an existing ManagedServiceAccount resource %s, skipping createManagedServiceAccount", cluster.Name, msaName)
+			continue
```
mkolesnik, line 42: If the MSA already exists, it's skipped. If the user changes `tokenValidity` on the mesh, existing MSAs keep the old value.
Fine for this PR, but please file a follow-up for updating existing MSAs when spec changes.

pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -0,0 +1,134 @@
...
+	addonv1beta1 "open-cluster-management.io/api/addon/v1beta1"
+	clusterv1 "open-cluster-management.io/api/cluster/v1"
+	msav1beta1 "open-cluster-management.io/managed-serviceaccount/apis/authentication/v1beta1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// deployManagedServiceAccountAddon deploys the ManagedServiceAccount add-on for each cluster in the ClusterSet
+func (r *Reconciler) deployManagedServiceAccountAddon(ctx context.Context, clusters []clusterv1.ManagedCluster) error {
+	if len(clusters) == 0 {
+		klog.Info("The ClusterSet has no managed cluster")
```
mkolesnik: nit: fires on every reconciliation for empty ClusterSets.

```suggestion
		klog.V(4).Info("The ClusterSet has no managed cluster")
```

pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -0,0 +1,134 @@
...
+		}
+	}
+
+	return nil
+}
+
+// createManagedServiceAccounts creates ManagedServiceAccount resources for each cluster in the ClusterSet
+func (r *Reconciler) createManagedServiceAccounts(ctx context.Context, mesh *meshv1alpha1.MultiClusterMesh, clusters []clusterv1.ManagedCluster) error {
+	if len(clusters) == 0 {
+		klog.Info("The ClusterSet has no managed cluster")
```
mkolesnik: nit: fires on every reconciliation for empty ClusterSets.

```suggestion
		klog.V(4).Info("The ClusterSet has no managed cluster")
```

pkg/hub/mesh/controller.go
```diff
@@ -346,6 +364,11 @@ func (r *Reconciler) handleDeletion(ctx context.Context, mesh *meshv1alpha1.Mult
 		return reconcile.Result{}, fmt.Errorf("failed to cleanup ManifestWorks: %w", err)
 	}
 
+	klog.Infof("Handling deletion for ManagedServiceAccount resources managed mesh %s/%s", mesh.Namespace, mesh.Name)
```
mkolesnik: nit: typo

```suggestion
	klog.Infof("Handling deletion for ManagedServiceAccount resources for mesh %s/%s", mesh.Namespace, mesh.Name)
```

test/integration/controller_test.go
```diff
@@ -604,6 +638,49 @@ func expectCacertsSecret(work *workv1.ManifestWork) {
...
+	// Test the following msa.Status fields after running additional controller(s)
+	// Expect(msa.Status.TokenSecretRef.Name).To(Equal(msa.Name))
+	// expire := msa.Status.TokenSecretRef.LastRefreshTimestamp.Add(msa.Spec.Rotation.Validity.Duration)
+	// Expect(msa.Status.ExpirationTimestamp.Time).To(Equal(expire))
+	return msa
+}
+
+// TODO: Run additional controller(s) from
+// https://github.com/open-cluster-management-io/managed-serviceaccount/tree/main/pkg/controllers/event
+func expectMsaSecret(msa *msav1beta1.ManagedServiceAccount) {
```
mkolesnik: This isn't used nor needed right now.
If it's ever needed lets add it then.

test/integration/controller_test.go
```diff
@@ -241,6 +242,27 @@ var _ = Describe("MultiClusterMesh Controller", func() {
 			expectFinalizer(meshName, testNs)
 		})
 
+		It("should create ManagedServiceAccount resources for each OCP ManagedCluster in the ClusterSet", func() {
```
mkolesnik: The test name says "for each OCP ManagedCluster" but `createManagedServiceAccounts` doesn't filter by platform. It creates MSAs for ALL clusters in the ClusterSet.
Either the code should check the product claim, or the test name should be updated and a separate test should verify behavior for non-OCP clusters.

test/integration/controller_test.go
```diff
@@ -389,6 +410,19 @@ var _ = Describe("MultiClusterMesh Controller", func() {
 
 			util.DeleteResource(ctx, k8sClient, &meshv1alpha1.MultiClusterMesh{}, meshName, testNs)
 		})
+
+		It("should cleanup related ManagedServiceAccount resources", func() {
```
mkolesnik: This test creates one mesh, deletes it, and asserts all MSAs are gone. It would pass even with the scoping bug.
A proper test should create two meshes with overlapping clusters, delete one, and verify only that mesh's MSAs are removed while the other's remain.

pkg/hub/mesh/controller.go
```diff
@@ -265,6 +266,23 @@ func (r *Reconciler) doReconcile(ctx context.Context, mesh *meshv1alpha1.MultiCl
 		}
 	}
 
+	// Deploy managed-serviceaccount addon for each cluster
```
mkolesnik: This call should be removed along with `deployManagedServiceAccountAddon` (see comment on endpoint_discovery.go:20).

test/integration/controller_test.go
mkolesnik, line 1: The MSA tests are missing lifecycle scenarios that the operator ManifestWork tests cover. Please add:
- Controller-created MSA has the correct default validity (360h from mesh spec)
- Cluster added to ClusterSet gets its MSA created
- Cluster removed from ClusterSet gets MSA cleaned up
- `Consistently` negative test: cluster without ClusterSet label doesn't get an MSA
- Cleanup with two meshes and overlapping clusters (the current test would pass even with the scoping bug)

pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -0,0 +1,134 @@
...
+	for _, cluster := range clusters {
+		existing := &msav1beta1.ManagedServiceAccount{}
+		// Naming convention: <cluster-name>-<mesh-name>-istio-reader
+		msaName := fmt.Sprintf("%s-%s-istio-reader", cluster.Name, mesh.Name)
+		if err := r.Get(ctx, types.NamespacedName{Name: msaName, Namespace: cluster.Name}, existing); err == nil {
+			klog.Infof("Cluster %s has an existing ManagedServiceAccount resource %s, skipping createManagedServiceAccount", cluster.Name, msaName)
+			continue
+		} else {
+			duration := mesh.Spec.Security.Discovery.TokenValidity
+			// A duration less than 10 minutes causes an error when applying a ManagedServiceAccount CR
```
mkolesnik: Silently clamping the duration hides a misconfiguration. If the user sets 5 minutes, they should know it was rejected, not have it quietly bumped to 10.

Options:
- Add a `+kubebuilder:validation:Format` or webhook to reject values < 10min at admission
- Set a status condition (e.g., `InvalidTokenValidity`) so it's visible in `kubectl describe`
- At minimum, return an error instead of silently correcting

The 10-minute minimum should also be documented on the `TokenValidity` field in types.go.

pkg/hub/mesh/controller.go
```diff
@@ -50,8 +50,9 @@ const (
 
 	DefaultChannel = "stable"
 
-	OperatorManifestWorkName = "multicluster-mesh-operator"
-	ManifestWorkNameCacerts  = "multicluster-mesh-cacerts"
+	OperatorManifestWorkName   = "multicluster-mesh-operator"
+	ManifestWorkNameCacerts    = "multicluster-mesh-cacerts"
+	ManifestWorkNameMsaSecrets = "multicluster-mesh-msa-secrets"
```
mkolesnik: `ManifestWorkNameMsaSecrets` is a fixed name per cluster, but the design allows multiple meshes to target the same ClusterSet (different `controlPlane.namespace`). Two meshes would collide on this ManifestWork name in each cluster namespace.
The name should include the mesh identity, similar to how the MSA itself is named per-mesh.
Note: `ManifestWorkNameCacerts` has the same issue.

mkolesnik (commented):
Please rebase onto current main.
In particular, #87 switched ManifestWork handling to `WorkApplier` and #83 scoped cleanup to a single ClusterSet.
After rebasing, please use `r.workApplier.Apply()` for the MSA ManifestWork path as well.

pkg/hub/mesh/endpoint_discovery.go
yxun: ok, I will remove that addon method and crd. This addon is installed by default on OpenShift ocm hub cluster. I was thinking about adding other kube integration tests by deploying that. I will skip that addon check.

pkg/hub/mesh/endpoint_discovery.go
yxun: https://github.com/stolostron/multicluster-mesh-addon/issues/120

yxun:
This PR has been broken into the following four small PRs. All following comments will be addressed in the small PRs.
I will close this original one when all the small PRs are merged.
- https://github.com/stolostron/multicluster-mesh-addon/pull/126
- https://github.com/stolostron/multicluster-mesh-addon/pull/127
- https://github.com/stolostron/multicluster-mesh-addon/pull/130

 https://github.com/stolostron/multicluster-mesh-addon/pull/ ? (pending for resolving pr 130 first)

mkolesnik:
closing this in favor of the incremental approach which is already ongoing

PR #56, opened 2026-05-04
Add listener for ManagedCluster changes

mkolesnik:
This makes sure that when ManagedCluster changes we reconcile the meshes properly.

### DIFF
--- a/pkg/hub/mesh/controller.go
+++ b/pkg/hub/mesh/controller.go
@@ -3,7 +3,6 @@ package mesh
 import (
 	"context"
 	"fmt"
-	"time"
 
 	operatorsv1 "github.com/operator-framework/api/pkg/operators/v1"
 	operatorsv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
@@ -19,6 +18,7 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
@@ -47,6 +47,7 @@ const (
 	LabelMeshName      = "mesh.open-cluster-management.io/mesh-name"
 	LabelMeshNamespace = "mesh.open-cluster-management.io/mesh-namespace"
 
+	ClusterSetLabel     = "cluster.open-cluster-management.io/clusterset"
 	clusterClaimProduct = "product.open-cluster-management.io"
 
 	// Product claim values from github.com/stolostron/multicloud-operators-foundation/pkg/klusterlet/clusterclaim
@@ -55,10 +56,6 @@ const (
 	ProductARO  = "ARO"
 	ProductROKS = "ROKS"
 	ProductOSD  = "OpenShiftDedicated"
-)
-
-var (
-	MissingClaimRequeueDelay = 30 * time.Second
 )
 
 // Reconciler reconciles MultiClusterMesh resources
@@ -76,6 +73,10 @@ func RegisterController(mgr manager.Manager) error {
 
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&meshv1alpha1.MultiClusterMesh{}).
+		Watches(
+			&clusterv1.ManagedCluster{},
+			handler.EnqueueRequestsFromMapFunc(reconciler.findMeshesForCluster),
+		).
 		Complete(reconciler)
 }
 
@@ -121,12 +122,9 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 
 	klog.Infof("Found %d clusters in set %s", len(clusters), mesh.Spec.ClusterSet)
 
-	// Install the operator on each cluster
-	missingClaims := false
 	for _, cluster := range clusters {
 		if getProductClaim(&cluster) == "" {
-			klog.V(4).Infof("Cluster %s missing product claim (needed for platform detection), skipping (will requeue)", cluster.Name)
-			missingClaims = true
+			klog.V(4).Infof("Cluster %s missing product claim (needed for platform detection), skipping", cluster.Name)
 			continue
 		}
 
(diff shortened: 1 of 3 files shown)


pkg/hub/mesh/controller.go
```diff
@@ -136,15 +134,46 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
+	}
+
+	meshList := &meshv1alpha1.MultiClusterMeshList{}
+	if err := r.List(ctx, meshList); err != nil {
+		klog.Errorf("Failed to list MultiClusterMeshes when handling cluster %s: %v", cluster.Name, err)
+		return
+	}
+
+	for _, mesh := range meshList.Items {
+		if mesh.Spec.ClusterSet == clusterSetName {
```
jewertow: We could avoid manually filtering items if you add a field indexer for `spec.clusterSet` (`mgr.GetFieldIndexer().IndexField()`).

pkg/hub/mesh/controller.go
```diff
@@ -218,6 +246,44 @@ func (r *Reconciler) ensureOperatorInstalled(ctx context.Context, mesh *meshv1al
...
+	}
+
+	// List ALL ManifestWorks that were created foir this mesh, some of which might be orphaned
+	workList := &workv1.ManifestWorkList{}
+	if err := r.List(ctx, workList, meshLabelSelector(mesh)); err != nil {
+		return fmt.Errorf("failed to list ManifestWorks: %w", err)
+	}
+
+	for i := range workList.Items {
+		work := &workList.Items[i]
```
jewertow: nit:
```suggestion
	for _, work := range workList.Items {
```

pkg/hub/mesh/controller.go
```diff
@@ -218,6 +246,44 @@ func (r *Reconciler) ensureOperatorInstalled(ctx context.Context, mesh *meshv1al
...
+		clusterName := work.Namespace
+
+		// Skip any clusters that are still in the set the mesh is referencing
+		if expectedClusters[clusterName] {
```
jewertow: What do you think about this?
```suggestion
		if slices.ContainsFunc(clusters, func(c clusterv1.ManagedCluster) bool {
            return c.Name == someName
        }) {
```
You wouldn't need to iterate over clusters to create the map.

pkg/hub/mesh/controller.go
```diff
@@ -246,6 +312,10 @@ func getOperatorManifestWorkName(isOCP bool) string {
 func (r *Reconciler) getClustersFromSet(ctx context.Context, clusterSetName string) ([]clusterv1.ManagedCluster, error) {
 	clusterSet := &clusterv1beta2.ManagedClusterSet{}
 	if err := r.Get(ctx, types.NamespacedName{Name: clusterSetName}, clusterSet); err != nil {
+		if errors.IsNotFound(err) {
+			klog.V(4).Infof("ManagedClusterSet %s not found, will reconcile when it is created", clusterSetName)
```
jewertow: This message is little confusing, because in this case user will see:
```
ManagedClusterSet test not found, will reconcile when it is created
Successfully reconciled MultiClusterMesh foo/bar
```
So I would suggest just saying that the cluster set was not found. Also, "will reconcile when it is created" is obvious in the context of k8s controllers.
```suggestion
			klog.V(4).Infof("ManagedClusterSet %s not found", clusterSetName)
```

pkg/hub/mesh/controller.go
mkolesnik: It's a bit more inefficient like that, since we'd need to (possibly) go over the entire slice for each MW.
With the map, we go over once and then the lookup is very efficient, so depending on amount of MW*clusters your proposed way, although simpler, can be much less efficient.

test/integration/controller_test.go
```diff
@@ -164,35 +164,93 @@ var _ = Describe("MultiClusterMesh Controller", func() {
...
+			})
+
+			It("should cleanup ManifestWork when the cluster is deleted", func() {
+				cluster := &clusterv1.ManagedCluster{}
+				Expect(k8sClient.Get(ctx, types.NamespacedName{Name: clusterName}, cluster)).To(Succeed())
+				Expect(k8sClient.Delete(ctx, cluster)).To(Succeed())
+				expectResourceDeleted(&clusterv1.ManagedCluster{}, testClusterSet, "")
+
+				awaitReconcileFinished()
+				expectNoManifestWorks()
```
jewertow: The previous test case verifies that ManifestWork was deleted, and it's not created back in the meantime, so this assertion does not really verify the cleanup.

test/integration/controller_test.go
```diff
@@ -298,6 +356,23 @@ func expectResourceDeleted(obj client.Object, name, namespace string) {
 	}).Should(BeTrue())
 }
 
+func awaitReconcileFinished() {
+	time.Sleep(1 * time.Second)
```
jewertow, line 348: Couldn't we use `Eventually()` in `expectNoManifestWorks()` to avoid sleeps?

test/integration/controller_test.go
mkolesnik: They test two different scenarios, so I'm not sure what you mean?

Each test creates a mesh + k8s cluster (look at BeforeEach on L#216), and makes sure the MW for the cluster is created.

I think it makes sense to test both, but perhaps I'm missing something in your comment?

test/integration/controller_test.go
mkolesnik: You're right, I'll change it where it's possible to use, otherwise we'll use `Consistently`

pkg/hub/mesh/controller.go
```diff
@@ -69,13 +66,23 @@ type Reconciler struct {
...
+	}
+
 	reconciler := &Reconciler{
 		Client: mgr.GetClient(),
 		Scheme: mgr.GetScheme(),
 	}
 
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&meshv1alpha1.MultiClusterMesh{}).
+		Watches(
```
sridhargaddam, line 82: Should the controller watch even the `ManagedClusterSet` as well to handle the scenario where `ManagedClusterSet` is deleted while the `MultiClusterMesh` still exists?

pkg/hub/mesh/controller.go
```diff
@@ -218,6 +248,41 @@ func (r *Reconciler) ensureOperatorInstalled(ctx context.Context, mesh *meshv1al
 	return nil
 }
 
+func (r *Reconciler) cleanupOrphanedManifestWorks(ctx context.Context, mesh *meshv1alpha1.MultiClusterMesh, clusters []clusterv1.ManagedCluster) error {
+	expectedClusters := make(map[string]bool)
+	for _, cluster := range clusters {
+		expectedClusters[cluster.Name] = true
+	}
+
+	// List ALL ManifestWorks that were created foir this mesh, some of which might be orphaned
```
sridhargaddam:
```suggestion
	// List ALL ManifestWorks that were created for this mesh, some of which might be orphaned
```

test/integration/controller_test.go
```diff
@@ -298,6 +344,35 @@ func expectResourceDeleted(obj client.Object, name, namespace string) {
...
+}
+
+func updateClusterSetLabel(clusterName, newClusterSet string) {
+	cluster := &clusterv1.ManagedCluster{}
+	Expect(k8sClient.Get(ctx, types.NamespacedName{Name: clusterName}, cluster)).To(Succeed())
+	cluster.Labels[meshcontroller.ClusterSetLabel] = newClusterSet
+	Expect(k8sClient.Update(ctx, cluster)).To(Succeed())
+}
+
+// expectNoManifestWorks makes sure that no ManifestWorks are created, checking consistenly
```
sridhargaddam:
```suggestion
// expectNoManifestWorks makes sure that no ManifestWorks are created, checking consistently
```

pkg/hub/mesh/controller.go
mkolesnik: Yes but I'm doing it iteratively so that'll be part of a future PR

mkolesnik (commented):
Thanks for the reviews.
I'm working on a revised approach that'll handle the operator singleton better, so maybe hold off on merging this since it also changes how the introduced behavior works.
Will post the new PR (based on this one) ASAP so that it can be evaluated

pkg/hub/mesh/controller.go
sridhargaddam: Okay Mike, thanks.

mkolesnik:
Closed in favor of #58

PR #58, opened 2026-05-11
Add listener for ManagedCluster changes with Singleton approach

mkolesnik:
This is an improvement over PR #56 which uses a singleton approach for the operator installation

Since there can be only one Sail/OSSM operatr installed, we shouldn't associate the ManifestWork with a specific mesh but rather with the addon.
The new code makes sure that when necessary, the ManifestWork gets removed and doesn't get removes if it's cluster is still being referenced by a mesh.

### DIFF
--- a/pkg/hub/mesh/controller.go
+++ b/pkg/hub/mesh/controller.go
@@ -44,8 +44,8 @@ const (
 
 	FinalizerName = "mesh.open-cluster-management.io/finalizer"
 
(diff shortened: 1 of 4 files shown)


mkolesnik:
Updated with a commit to use a single `ManifestWork` instead of one for OSSM and one for Sail.

The commit is pretty simple and small, and makes sense to have as part of this PR.

jewertow (approved):

PR #60, opened 2026-05-13
Add ManagedClusterSet listener

mkolesnik:
The listener makes sure that meshes get reconciled when ManagedClusterSets are created/updated/deleted.


sridhargaddam (approved):
LGTM, thanks

PR #61, opened 2026-05-13
Add predicate to ManagedClusterSet watcher

sridhargaddam:
The ManagedClusterSet watcher triggers reconciliation on every
event, including status-only and metadata-only updates. The
reconciler only reads Spec.ClusterSelector.SelectorType from
ManagedClusterSet, so these extra reconciliations are unnecessary.


mkolesnik (approved):
lgtm

PR #62, opened 2026-05-13
Add some basic documentation

mkolesnik:
Add `AGENTS.md` (and `CLAUDE.md`) referencing it.

Also add a high level design doc that outlines the design of the addon.

### DIFF
--- a/docs/design.md
+++ b/docs/design.md
@@ -1,0 +1,215 @@
+# Design
+
+## Table of Contents
+
+- [Overview](#overview)
+- [Architecture](#architecture)
+- [Scope](#scope)
+- [Supported Topologies](#supported-topologies)
+- [Custom Resource](#custom-resource)
+- [Cluster Selection and Multi-Tenancy](#cluster-selection-and-multi-tenancy)
+- [Operator Lifecycle](#operator-lifecycle)
+- [Trust Distribution](#trust-distribution)
+- [Endpoint Discovery](#endpoint-discovery)
+- [Phased Approach](#phased-approach)
+
+## Overview
+
+The OCM Service Mesh Add-on automates multi-cluster Istio service mesh setup via [OCM]. It manages the `MultiClusterMesh` custom resource on the hub cluster to orchestrate three concerns across managed clusters:
+
+1. **Operator Lifecycle** - Installing and managing the [Sail]/OSSM operator
+2. **Trust Distribution** - Establishing mTLS trust via [cert-manager]
+3. **Endpoint Discovery** - Exchanging discovery credentials via [ManagedServiceAccount]
+
+Without this add-on, multi-cluster mesh setup is a manual process involving certificate management, O(N^2) secret exchanges, and per-cluster operator configuration.
+
+## Architecture
+
+The add-on follows OCM's hub-and-spoke model:
+
+- **Hub**: Manages `MultiClusterMesh` resources, creates [ManifestWorks][ManifestWork], orchestrates cert-manager and ManagedServiceAccount
+- **Spoke** (managed clusters): Receives ManifestWorks from the hub, runs the Sail/OSSM operator and Istio control plane
+
+```mermaid
+flowchart TD
+    subgraph Hub Cluster
+        mesh([MultiClusterMesh]) --> addon[Mesh Add-on]
+        addon --> cert(["Certificate
+        (per cluster)"])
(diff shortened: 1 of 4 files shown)


docs/design.md
```diff
@@ -0,0 +1,215 @@
...
+2. **Trust Distribution** - Establishing mTLS trust via [cert-manager]
+3. **Endpoint Discovery** - Exchanging discovery credentials via [ManagedServiceAccount]
+
+Without this add-on, multi-cluster mesh setup is a manual process involving certificate management, O(N^2) secret exchanges, and per-cluster operator configuration.
+
+## Architecture
+
+The add-on follows OCM's hub-and-spoke model:
+
+- **Hub**: Manages `MultiClusterMesh` resources, creates [ManifestWorks][ManifestWork], orchestrates cert-manager and ManagedServiceAccount
```
FilipB: Might be a bit confusing? `MultiClusterMesh` resource is managed by Service Mesh Add-on controller. Hub is more general term.

docs/design.md
```diff
@@ -0,0 +1,215 @@
...
+- **Spoke** (managed clusters): Receives ManifestWorks from the hub, runs the Sail/OSSM operator and Istio control plane
+
+```mermaid
+flowchart TD
+    subgraph Hub Cluster
+        mesh([MultiClusterMesh]) --> addon[Mesh Add-on]
```
FilipB, line 36: Not sure if we should call it Mesh add-on or Mesh controller. If we want to call it add-on, we should probably create ClusterManagementAddOn resource for proper ACM integration. On the other hand  we don't need to capture here current implementation state but target state which is to be a proper ACM add-on. So maybe ignore this comment:)

docs/design.md
```diff
@@ -0,0 +1,215 @@
...
+        (cacerts)"])
+        agent --> remotesecret(["Secret
+        (remote secret, per peer)"])
+    end
+
+    mw_operator --> agent
+    mw_cacerts --> agent
+    mw_remote --> agent
+
+    subgraph Legend
```
FilipB: Would it make sense to differentiate ACM resources (ManifestWork), OSSM resources, cert-manager resources and general OCP resources?

docs/design.md
```diff
@@ -0,0 +1,215 @@
...
+
+
+## Scope
+
+### What the add-on does (Plumbing)
+
+- Installs the Sail/OSSM operator on managed clusters via OLM
+- Distributes intermediate CA certificates for mTLS trust
+- Exchanges discovery tokens between peer clusters
+- Handles lifecycle events (cluster add/remove, mesh creation/deletion)
```
FilipB, line 103: These are managed clusters and mesh events right? We don't handle OSSM add-on events like enable/disable add-on atm. Maybe we should describe it here.

docs/design.md
```diff
@@ -0,0 +1,215 @@
...
+
+### What the add-on does not do (Configuration)
```
FilipB, line 105: Should we mention kiali also?

docs/design.md
mkolesnik: Yes the design describes the target state.

Also the `ClusterManagementAddOn` resource is an implementation detail, isn't it? :)

docs/design.md
mkolesnik: Updated to make it specific

docs/design.md
mkolesnik: The diagram now color-codes resources by managing component, hope this makes it clearer.
Or did you think of something else?

docs/design.md
mkolesnik: Sure, added it to the "does not do" list and mentioned in Phase 2

docs/design.md
mkolesnik: It's already mentioned in general under 
> - Does not deploy monitoring, observability, or application workloads

Do you think we need to list it explicitly?

docs/design.md
FilipB: Not sure, it's probably design decision how much we want to integrate with ACM. Not important, we will know more after the call with ACM team.

docs/design.md
FilipB: It's probably good enough to mention observability in general.

FilipB (commented):
LGTM, not approving so others can review too.

sridhargaddam (approved):
LGTM, thanks @mkolesnik

PR #63, opened 2026-05-14
Add deployment support with make targets and health probes

FilipB:
Introduces deployment infrastructure for development and testing:
- Split deployment into separate numbered manifests (namespace, serviceaccount, role, rolebinding, deployment)
- Add gen-rbac target to generate RBAC from kubebuilder annotations in controller code
- Add make targets: deploy, undeploy, push for add-on deployment
- Add health and readiness probe endpoints to controller
- Switch from library-go to controller-runtime leader election to fix deployment rollout deadlock (non-leader pods must serve health probes)
- Add namespace detection via POD_NAMESPACE env var with service account fallback
- Change default REGISTRY_BASE from quay.io/stolostron to quay.io/sail-dev
- Update .gitignore to allow config/deploy while ignoring config/crd
- Document deployment process in README


Makefile
```diff
@@ -35,7 +35,7 @@ $(ENVTEST): $(BIN_DIR)
 	@test -s $(ENVTEST) || GOBIN=$(BIN_DIR) go install sigs.k8s.io/controller-runtime/tools/setup-envtest@$(CONTROLLER_RUNTIME_BRANCH)
 
 # Image registry and name
-REGISTRY_BASE ?= quay.io/stolostron
+REGISTRY_BASE ?= quay.io/sail-dev
```
mkolesnik, line 39: Are we sure we want this in sail-dev though?

I think we need to coordinate with ACM folks.

AFAIK submariner addon, for example, does use quay.io/stolostron

Makefile
```diff
@@ -120,6 +120,20 @@ clean: ## Clean build artifacts
 images: ## Build the container image
 	$(CONTAINER_ENGINE) build -t $(IMG) .
 
+.PHONY: push
+push: images ## Push the container image
+	$(CONTAINER_ENGINE) push $(IMG)
+
+.PHONY: deploy
+deploy: push ## Deploy the controller to the cluster
```
mkolesnik: Shouldn't it rely on `gen-crds`?

config/deploy/deployment.yaml
```diff
@@ -0,0 +1,204 @@
+apiVersion: v1
```
mkolesnik: How about we use kustomize and split into "themed" files instead of a monolithic file?
This should also allow more k8s native replacement of fields (such as image)

mkolesnik (commented):
Preliminary review

Makefile
FilipB: It will take some time to get access to be able to push to `quay.io/stolostron` so we might switch later. `quay.io/sail-dev` will unblock us for now. I already have jenkins job which will build and push to `quay.io/sail-dev`

config/deploy/deployment.yaml
FilipB: Yes, we should discuss if we want to use kustomize or helm but I think we should wait after discussion with ACM team. So we can do that in follow up PR

config/deploy/deployment.yaml
```diff
@@ -0,0 +1,204 @@
...
+  name: multicluster-mesh-controller
+  namespace: multicluster-mesh-system
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - pods
+  verbs:
+  - get
+  - list
```
sridhargaddam: Do we need permissions for listing the pods? AFAICT, the controller is not listing any pods.

config/deploy/deployment.yaml
```diff
@@ -0,0 +1,204 @@
...
+      app: multicluster-mesh-controller
+  template:
+    metadata:
+      labels:
+        app: multicluster-mesh-controller
+    spec:
+      serviceAccountName: multicluster-mesh-controller
+      containers:
+      - name: controller
+        image: quay.io/stolostron/multicluster-mesh-addon:latest
```
sridhargaddam: The REGISTRY_BASE is updated to use sail-dev. To avoid confusion, shall we use the same path here?

config/deploy/deployment.yaml
FilipB: Good catch, thank you. It's removed and it works fine.

README.md
```diff
@@ -34,7 +34,66 @@ The addon manages the "plumbing" (trust and connectivity) while users configure
 
 ### Installation
 
-TBD
+#### Development Deployment
+
+To build and deploy the controller to your cluster:
+
```
sridhargaddam, line 47: Please add a pre-requisite section here capturing information like kubeconfig should be pointing to some existing cluster etc.

main.go
```diff
@@ -84,17 +85,21 @@ func newCommand() *cobra.Command {
...
 	probeAddr   string
+	leaderElect bool
 )
 
 func newControllerCommand() *cobra.Command {
-	cmd := controllercmd.
-		NewControllerCommandConfig("multicluster-mesh-addon-controller", version.Get(), runController, clock.RealClock{}).
-		NewCommand()
+	cmdConfig := controllercmd.NewControllerCommandConfig("multicluster-mesh-addon-controller", version.Get(), runController, clock.RealClock{})
+	// Disable library-go leader election - controller-runtime will handle it
```
sridhargaddam: Can you explain or eloborate on why we are disabling library-go leader election?

main.go
```diff
@@ -103,15 +108,16 @@ func runController(ctx context.Context, controllerContext *controllercmd.Control
...
 		Scheme: runtimeScheme,
 		Metrics: metricsserver.Options{
 			BindAddress: metricsAddr,
 		},
-		HealthProbeBindAddress: probeAddr,
-		LeaderElection:         false,
+		HealthProbeBindAddress:  probeAddr,
+		LeaderElection:          leaderElect,
+		LeaderElectionID:        "multicluster-mesh-addon-controller-lock",
+		LeaderElectionNamespace: "multicluster-mesh-system",
```
sridhargaddam: Instead of hardcoding the LeaderElectionNamespace to "multicluster-mesh-system", its better to use the POD_NAMESPACE via the downward API.

sridhargaddam (commented):
Thanks for addressing the comments @FilipB. Some more observations after reviewing the PR...

main.go
FilipB: Thanks. Updated. I followed what we do in sail operator. I think this whole section will change once we understand better integration points with ACM and if we become a default add-on.

README.md
FilipB: Done

main.go
FilipB: Done

sridhargaddam (commented):
LGTM. I'll let @mkolesnik to take another look and approve the PR.

Makefile
mkolesnik: OK, we should add a TODO comment here not to forget to change it

config/deploy/deployment.yaml
```diff
@@ -0,0 +1,202 @@
...
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: multicluster-mesh-controller
+subjects:
+- kind: ServiceAccount
+  name: multicluster-mesh-controller
+  namespace: multicluster-mesh-system
+---
+apiVersion: rbac.authorization.k8s.io/v1
```
mkolesnik: How about generating the RBAC from code?

We already have the RBAC specified in the code, might as well use it.
That way the RBAC in the deployment always matches the code annotations and can't drift.

config/deploy/deployment.yaml
mkolesnik: Sure, let's discuss this and follow up if necessary.

mkolesnik (commented):
Couple small comments, looks good otherwise

config/deploy/deployment.yaml
FilipB: Done

FilipB:
@mkolesnik Added also some reasonable Claude suggestions for improvements into Makefile

Makefile
```diff
@@ -56,8 +59,16 @@ gen-code: ## Generate DeepCopy code
 gen-crds: ## Generate CRD manifests
 	$(CONTROLLER_GEN) crd paths=./pkg/apis/... output:crd:dir=./config/crd
 
+.PHONY: gen-rbac
+gen-rbac: ## Generate RBAC manifests from controller annotations
+	$(CONTROLLER_GEN) rbac:roleName=multicluster-mesh-controller paths=./pkg/hub/mesh output:rbac:dir=./config/deploy
```
mkolesnik: Perhaps use a temp file/dir for this output?
This way you won't have to bother deleting it

mkolesnik (commented):
Thanks, looks good, one small suggestion

Makefile
FilipB: Done

FilipB:
I will squash the commits to have cleaner merge commit msg.

mkolesnik (approved):

PR #64, opened 2026-05-14
Do not treat a terminating ManifestWork as an error

sridhargaddam:
When ensureOperatorInstalled detects a ManifestWork in the process of
being deleted, it returns a plain error, causing the reconciler to repeatedly
log "Failed to ensure mesh operator on cluster ..." during each requeue
cycle, even though the deletion is proceeding normally. This PR fixes that.

### DIFF
--- a/pkg/hub/mesh/controller.go
+++ b/pkg/hub/mesh/controller.go
@@ -3,6 +3,7 @@ package mesh
 import (
 	"context"
 	"fmt"
+	"time"
 
 	operatorsv1 "github.com/operator-framework/api/pkg/operators/v1"
 	operatorsv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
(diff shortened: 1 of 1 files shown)


mkolesnik (commented):
Doesn't it automatically requeue with exponential backoff?

sridhargaddam:
> Doesn't it automatically requeue with exponential backoff?

I think it does. This PR is mostly to fix the erroreneous log message where it says "Failed to ensure mesh operator on cluster ..." during each requeue cycle, **even though the deletion is proceeding normally**.

jewertow (approved):

jewertow:
LGTM from my side, but I let @mkolesnik to review again before merge.

> > Doesn't it automatically requeue with exponential backoff?

> I think it does.

From my experience and from I read now, I can say it does not apply the exponential backoff. This is the controller's responsibility, and `RequeueAfter` is a fixed time.

mkolesnik (approved):

PR #68, opened 2026-05-18
controller: use typed watches

jewertow:


pkg/hub/mesh/controller.go
```diff
@@ -81,14 +81,20 @@ func RegisterController(mgr manager.Manager) error {
 
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&meshv1alpha1.MultiClusterMesh{}).
-		Watches(
-			&clusterv1.ManagedCluster{},
-			handler.EnqueueRequestsFromMapFunc(reconciler.findMeshesForCluster),
+		WatchesRawSource(
```
mkolesnik, line 84: Honestly I think the old one looks better and is more legible :)

This seems to save us the typecast line (actual type safety is indirectly checked in runtime by the integ tests)

jewertow: I agree. I opened this PR to address @nrfox's [request](https://github.com/stolostron/multicluster-mesh-addon/pull/51#discussion_r3241497007), but I am not sure whether we want it, so I keep it as a draft.

PR #70, opened 2026-05-18
Add status reporting

mkolesnik:
Add basic status reporting to report status per cluster and overall. Right now this is a very basic implementation, and will be expanded upon in upcoming future work.

### DIFF
--- a/pkg/hub/mesh/controller.go
+++ b/pkg/hub/mesh/controller.go
@@ -2,6 +2,7 @@ package mesh
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"reflect"
 	"slices"
@@ -13,7 +14,8 @@ import (
 	operatorsv1 "github.com/operator-framework/api/pkg/operators/v1"
 	operatorsv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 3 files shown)


pkg/hub/mesh/controller.go
```diff
@@ -316,14 +337,83 @@ func (r *Reconciler) cleanupOperatorManifestWorks(ctx context.Context) error {
...
+	mesh.Status.ClusterStatus = clusterStatuses
+
+	readyCondition := metav1.Condition{Type: meshv1alpha1.ConditionReady}
+	if allReady {
+		readyCondition.Status = metav1.ConditionTrue
+		readyCondition.Reason = meshv1alpha1.ReasonAllClustersReady
+		readyCondition.Message = "All clusters are ready"
+	} else {
+		readyCondition.Status = metav1.ConditionFalse
+		readyCondition.Reason = meshv1alpha1.ReasonClustersNotReady
```
jewertow, line 420: When `allReady` is false I would expect either `ClustersNotReady=true` or `AllClustersReady=false`.

Logically `NOT (ReasonClustersNotReady)` => `ReasonClustersReady`, right?

pkg/hub/mesh/controller.go
```diff
@@ -316,14 +337,83 @@ func (r *Reconciler) cleanupOperatorManifestWorks(ctx context.Context) error {
...
+		}
+
+		work := &workv1.ManifestWork{}
+		err := r.Get(ctx, types.NamespacedName{
+			Name:      OperatorManifestWorkName,
+			Namespace: cluster.Name,
+		}, work)
+
+		if err == nil {
+			// TODO: Set to actual status when operator installation is confirmed via ManifestWork status feedback
```
jewertow, line 396: I my assumption correct that we have to set `spec.manifestConfigs[].feedbackRules` in the ManifestWork to read the actual status?

pkg/hub/mesh/controller.go
mkolesnik: Yeah this follows the k8s condition convention, the type is positive (`Ready`) and the reason explains why it's in that state. So `Ready=False, Reason=ClustersNotReady` reads as "not ready because clusters aren't ready".

Using a negative type like `ClustersNotReady=True` is discouraged by the k8s API conventions.

pkg/hub/mesh/controller.go
mkolesnik: Correct, this PR only reports that the ManifestWork was created, not that the operator is actually running on the spoke.
The TODO on that line tracks adding `feedbackRules` to check actual installation status.
I wanted to get the status framework in place first, since its useful for more changes, then iterate on the actual feedback.

pkg/hub/mesh/controller.go
jewertow: Ah, right. Makes sense. Thank you for clarification.

jewertow (approved):

mkolesnik:
Had to merge main, tests all pass now (again)

jewertow (approved):

PR #73, opened 2026-05-19
Add listener for ManifestWork

mkolesnik:
The listener will reconcile in case an owned ManifestWork is created/updated/delete.
In case of creation/update the controller is idempotently handling the reconciliation.

### DIFF
--- a/pkg/hub/mesh/controller.go
+++ b/pkg/hub/mesh/controller.go
@@ -110,6 +110,13 @@ func RegisterController(mgr manager.Manager) error {
 				return obj.GetLabels()[MeshNameLabel] != "" && obj.GetLabels()[MeshNamespaceLabel] != ""
 			})),
 		).
+		Watches(
+			&workv1.ManifestWork{},
+			handler.EnqueueRequestsFromMapFunc(reconciler.findMeshesForManifestWork),
+			builder.WithPredicates(predicate.NewPredicateFuncs(func(obj client.Object) bool {
+				return obj.GetLabels()[ManagedByLabel] == ManagedByValue
+			})),
+		).
 		Complete(reconciler)
 }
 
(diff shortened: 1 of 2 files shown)


pkg/hub/mesh/controller.go
```diff
@@ -244,6 +251,17 @@ func (r *Reconciler) findMeshesForClusterSet(ctx context.Context, obj client.Obj
 	return requests
 }
 
+// findMeshesForManifestWork returns a list of all meshes to reconcile following a ManifestWork change
+func (r *Reconciler) findMeshesForManifestWork(ctx context.Context, obj client.Object) []reconcile.Request {
+	cluster := &clusterv1.ManagedCluster{}
+	if err := r.Get(ctx, types.NamespacedName{Name: obj.GetNamespace()}, cluster); err != nil {
```
sridhargaddam, line 257: Should we handle the `errors.IsNotFound` (as an expected case) here to handle cluster teardown scenarios?

mkolesnik: Do you mean you want to treat it differently?

Right now any error will log, including isNotFound, so no real special treatment is needed, unless you want to split the logs (e.g. log at error level if it's any other error besides that).

What do you propose we do with it?

sridhargaddam: > Do you mean you want to treat it differently?

Yes, I was thinking that if its `errors.IsNotFound`, we can simply avoid logging.

mkolesnik: Thinking about this some more, if MC is being deleted the namespace (with all objects) is also being deleted, so I don't think there's much sense in handling this case specifically.

sridhargaddam (approved):

mkolesnik:
@sridhargaddam thanks, there was a race condition in the integ test.
PTAL @ the 2nd commit (cce1d7b3b5986188c33b47375ad56c8ba222ac02) which should fix it

sridhargaddam (approved):

Issue #71, opened 2026-05-18
Add support for ClusterIssuer


jewertow:
Currently, our implementation only supports namespace-scoped `Issuer` resources. There is no architectural restriction preventing us from supporting `ClusterIssuer` as well. Extending support to `ClusterIssuer` will give users greater flexibility, allowing them to define cluster-wide certificate authorities without duplicating issuer configurations across multiple namespaces.

Implementing this requires an update to our API surface to allow users to specify the issuer's kind, alongside a backend architectural change regarding where the target certificates are created.

PR #76, opened 2026-05-20
Add sample YAML files and update README

FilipB:
Create samples directory with five MultiClusterMesh examples demonstrating different use cases: basic minimal config, complete with all fields, OpenShift-specific, version pinning with manual approval, and cert-manager Issuer setup. Update README to reference samples instead of showing full YAML inline.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,35 +104,35 @@ kubectl create namespace multicluster-mesh-system
 
 ### Usage
 
-Create a namespace to host your multi cluster mesh resources.
-Inside the namespace, create an `Issuer` resource that will act as the Root CA.
+#### Quick Start
 
-Create a `MultiClusterMesh` resource on the hub cluster:
-
-```yaml
-apiVersion: mesh.open-cluster-management.io/v1alpha1
-kind: MultiClusterMesh
-metadata:
-  name: prod-mesh
-  namespace: mesh-ns
-spec:
-  clusterSet: finance-prod
-  controlPlane:
-    namespace: istio-system
-  operator:
-    channel: "stable"
-    source: redhat-operators
-  security:
-    trust:
-      certManager:
-        issuerRef:
-          name: mesh-root-issuer
-    discovery:
-      tokenValidity: "1w"
+1. Create a namespace for your mesh resources:
+```bash
+kubectl create namespace mesh-system
 ```
 
+2. Create a cert-manager Issuer to act as the Root CA:
+```bash
+kubectl apply -f samples/cert-manager-issuer.yaml
+```
+
+3. Deploy a basic MultiClusterMesh:
+```bash
+kubectl apply -f samples/basic.yaml
+```
+
+For more configuration options, see the [samples](./samples/) directory:
+
+- **[basic.yaml](./samples/basic.yaml)** - Minimal configuration using defaults
+- **[complete.yaml](./samples/complete.yaml)** - All available fields with documentation
+- **[openshift.yaml](./samples/openshift.yaml)** - OpenShift-specific configuration
+- **[pinned-version.yaml](./samples/pinned-version.yaml)** - Version pinning with manual approval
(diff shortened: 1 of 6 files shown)


sridhargaddam (commented):
I'm fine with the PR. I'll let others in the team to take a look and approve it.

samples/cert-manager-issuer.yaml
```diff
@@ -0,0 +1,20 @@
...
+---
+# Alternative: ClusterIssuer for cluster-wide root CA
+# Note: MultiClusterMesh currently references Issuers, not ClusterIssuers
+
+# apiVersion: cert-manager.io/v1
+# kind: ClusterIssuer
+# metadata:
+#   name: mesh-root-ca
+# spec:
+#   selfSigned: {}
```
jewertow: I would remove it for now, and add back in a dedicated file once its supported.

jewertow (changes requested):

samples/cert-manager-issuer.yaml
FilipB: It will be supported when https://github.com/stolostron/multicluster-mesh-addon/pull/51 is merged or something more is missing?

jewertow: No, that PR adds support only for `Issuer`. `ClusterIssuer` is on the roadmap (#71), but I can't say when we will do it.

README.md
```diff
@@ -104,35 +104,35 @@ kubectl create namespace multicluster-mesh-system
...
+```bash
+kubectl apply -f samples/cert-manager-issuer.yaml
 ```
 
+3. Deploy a basic MultiClusterMesh:
+```bash
+kubectl apply -f samples/basic.yaml
+```
+
+For more configuration options, see the [samples](./samples/) directory:
```
mkolesnik, line 126: I see they all target `default` cluster set.

Do we want to specifically instruct them to set this field?

I feel it's easy to overlook like this, and it's the most basic decision they need to make..

samples/openshift.yaml
```diff
@@ -0,0 +1,35 @@
...
+  controlPlane:
+    namespace: istio-system
+
+  # OpenShift-specific operator configuration
+  operator:
+    # On OpenShift, the operator is typically installed in openshift-operators
+    namespace: openshift-operators
+
+    # Use Red Hat's operator catalog
+    channel: "stable"
```
mkolesnik:
```suggestion
    channel: stable
```

nit: do we need the quotes (also for other field values)?

samples/openshift.yaml
```diff
@@ -0,0 +1,35 @@
...
+
+    # Automatic approval for production environments
+    installPlanApproval: Automatic
+
+  # Security configuration
+  security:
+    trust:
+      certManager:
+        issuerRef:
+          name: openshift-mesh-root-ca
```
mkolesnik: This won't work though since this isn't created by the user (unless such an issuer is expected).

Maybe add a comment above it to clarify this?

Also, there's nothing special for openshift here FWIW

samples/cert-manager-issuer.yaml
FilipB: Done

samples/openshift.yaml
FilipB: It's question about yaml best practices, both options are valid. I will remove them so it's consistent with other fields.

samples/openshift.yaml
FilipB: Thanks. I updated the name to mach the name used in the sample Issuer. OCP specific fields are `namespace`, `source` and `sourceNamespace`. Those are different in the `complete.yaml` sample which is using values for vanilla k8s. I guess it will make better sense when there are more differences. Although I'm not sure if there will be any more differences added in the future.

jewertow (approved):

PR #77, opened 2026-05-20
Fix KIND controller crash due to missing auth-reader RoleBinding

sridhargaddam:
When the controller was deployed on a KIND cluster it was crashlooping with the following error log.

```
  error initializing delegating authentication: unable to load configmap
  based request-header-client-ca-file: configmaps
  "extension-apiserver-authentication" is forbidden
```

The delegating authenticator in k8s.io/apiserver reads the extension-apiserver-authentication ConfigMap from kube-system at startup to bootstrap request-header CA validation. On OpenShift this works out of the box because authenticated users get read access to that ConfigMap by default, but vanilla Kubernetes doesn't grant it.

This PR adds the standard RoleBinding to extension-apiserver-authentication-reader in kube-system so the controller starts cleanly on any cluster.

### DIFF
--- a/config/deploy/07-rolebinding-auth-reader.yaml
+++ b/config/deploy/07-rolebinding-auth-reader.yaml
@@ -1,0 +1,14 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: multicluster-mesh-controller-auth-reader
+  namespace: kube-system
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: extension-apiserver-authentication-reader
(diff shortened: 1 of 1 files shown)


FilipB (commented):
Is this good time to use kustomized or helm to generate correct set of yamls for kind vs. ocp?

sridhargaddam:
> Is this good time to use kustomized or helm to generate correct set of yamls for kind vs. ocp?

Yeah, thats a good idea. I've started looking into the kustomize changes. Can we merge this PR, so that the need for this roleBinding will be clear in the project history and I'll push a Kustomize PR shortly with the necessary changes?

FilipB (approved):

PR #80, opened 2026-05-20
Use Kustomize for platform-specific manifests

sridhargaddam:
This PR replaces the flat config/deploy/ layout a proper Kustomize base + overlays structure. Set PLATFORM=kind or PLATFORM=openshift (default) when deploying.


config/deploy/base/kustomization.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - 00-namespace.yaml
```
FilipB: Nit: files no longer need to be prefixed with numbers as the order `oc apply` will apply them is defined here.

Makefile
```diff
@@ -137,19 +144,17 @@ images: ## Build the container image
...
 	$(CONTAINER_ENGINE) push $(IMG)
 
-# TODO: Use kustomized or helm instead of sed
 .PHONY: deploy
-deploy: gen push ## Deploy the controller to the cluster
+deploy: gen push ## Deploy the controller to the cluster (PLATFORM=openshift|kind)
 	kubectl apply -f config/crd/
-	kubectl apply -f config/deploy/
-	sed 's|image:.*|image: $(IMG)|' config/deploy/06-deployment.yaml | kubectl apply -f -
+	kubectl kustomize config/deploy/overlays/$(PLATFORM) | sed 's|image:.*|image: $(IMG)|' | kubectl apply -f -
```
FilipB: I believe the image could be set without `sed` using just kustomized

Makefile
```diff
@@ -137,19 +150,20 @@ images: ## Build the container image
...
-# TODO: Use kustomized or helm instead of sed
 .PHONY: deploy
-deploy: gen push ## Deploy the controller to the cluster
+deploy: gen push $(KUSTOMIZE) ## Deploy the controller to the cluster (PLATFORM=openshift|kind)
 	kubectl apply -f config/crd/
-	kubectl apply -f config/deploy/
-	sed 's|image:.*|image: $(IMG)|' config/deploy/06-deployment.yaml | kubectl apply -f -
+	cd config/deploy/overlays/$(PLATFORM) && $(KUSTOMIZE) edit set image quay.io/sail-dev/multicluster-mesh-addon=$(IMG)
+	kubectl apply -k config/deploy/overlays/$(PLATFORM)
+	# kustomize edit set image writes the resolved image tag into kustomization.yaml on disk. Lets restore it to original file.
```
FilipB, line 158: Sorry, I didn't realize this will be necessary so now I'm not sure what is better. If simple sed or this. Do we want to always restore it? What about adding a default value for image to kustomization.yaml so it would be updated only with non default image and not restoring at all?

sridhargaddam: > Do we want to always restore it? 

Yes, we have to restore it otherwise it will appear like a modified file in the local codebase.

> What about adding a default value for image to kustomization.yaml so it would be updated only with non default image and not restoring at all?

The default value for the image would be based on the git commit id. So, we cannot have a default value like "latest" or "devel". Even if we keep a default value of "latest", it will be overwritten by kustomize and the file will appear like modified in the git status (effectively means, we still have to call git checkout ...). If you rather prefer the original "sed", I can revert the change done in commit 2 related to this. Please do let me know.

FilipB: Thanks. Let's keep current approach.

FilipB (approved):

PR #81, opened 2026-05-25
Support for a local Kind+OCM dev environment

sridhargaddam:
This PR includes the necessary scripts to create a complete local development environment that provisions 3 Kind clusters, installs OCM and OLM, registers managed clusters, and deploys the addon controller, via the `make dev-env` command. It also supports the following make targets.

**dev-env:**                Provision full dev environment (Kind + OCM + addon)
**create-clusters:**  Creates three Kind clusters (hub, cluster1, cluster2)
**install-olm:**      Install OLM on managed clusters (cluster1, cluster2)
**init-ocm:**         Initialize hub as OCM control plane
**join-clusters:**    Register managed clusters and create a ManagedClusterSet
**deploy-addon:**     Build and deploy addon controller to the hub Kind cluster
**dev-clean:**        Destroy dev clusters and remove .kube/ folder


config/deploy/overlays/kind/kustomization.yaml
```diff
@@ -3,3 +3,11 @@ kind: Kustomization
...
   - ../../base
   - rolebinding-auth-reader.yaml
+patches:
+  - target:
+      kind: Deployment
+      name: multicluster-mesh-controller
+    patch: |
+      - op: replace
+        path: /spec/template/spec/containers/0/imagePullPolicy
+        value: IfNotPresent
```
FilipB, line 16: Why not `Always`?

samples/kind-dev.yaml
```diff
@@ -0,0 +1,8 @@
+---
```
FilipB, line 1: Maybe it should be moved to samples dir so it's aligned with https://github.com/stolostron/multicluster-mesh-addon/pull/76/changes?

hack/dev-env.sh
```diff
@@ -0,0 +1,366 @@
...
+        log "Labeling ${cluster_name} with clusterset=mesh-cluster-set"
+        kubectl --kubeconfig="${HUB_KUBECONFIG}" label managedcluster "${cluster_name}" \
+            cluster.open-cluster-management.io/clusterset=mesh-cluster-set \
+            --overwrite
+    done
+
+    # Create product ClusterClaim on each managed cluster so the OCM registration
+    # agent syncs it to ManagedCluster.status.clusterClaims on the hub.
+    # The addon controller uses this for platform detection.
+    for cluster_name in "${CLUSTER1_NAME}" "${CLUSTER2_NAME}"; do
```
FilipB: Isn't this created automatically? At least it is on openshift.

FilipB (commented):
Could you please update also the Readme?
I hit `too many open files` on Fedora 44 so we should probably also link [this](https://kind.sigs.k8s.io/docs/user/known-issues/#pod-errors-due-to-too-many-open-files) known issue.

config/deploy/overlays/kind/kustomization.yaml
sridhargaddam: `IfNotPresent` is required because images are side-loaded into Kind via "kind load image-archive" and never pushed to a registry. With "Always", the kubelet would try to pull from the registry and fail with `ErrImagePull`.

I've also added a comment so that it helps anyone looking at the code.

hack/dev-env.sh
sridhargaddam: On OCP, its automatically created, but for KIND we have to explicitly create it as the klusterlet foundation agent is not installed on Kind.

config/deploy/overlays/kind/kustomization.yaml
FilipB: Thanks.

hack/dev-env.sh
```diff
@@ -0,0 +1,368 @@
...
+        armv7*)  echo "arm" ;;
+        *)       err "Unsupported architecture: ${arch}" ;;
+    esac
+}
+
+detect_os() {
+    uname | tr '[:upper:]' '[:lower:]'
+}
+
+install_deps() {
```
mkolesnik: I think this makes more sense to put in the Makefile, similar to how Sail does it.
The tools are an "external concern" for the script, it should be focused on setting up the env & etc, not fetching the tools.

e.g. in Sail
```Makefile
$(OPERATOR_SDK): $(LOCALBIN)
        @if test -x $(LOCALBIN)/operator-sdk && ! $(LOCALBIN)/operator-sdk version | grep -q $(OPERATOR_SDK_VERSION); then \
                echo "$(LOCALBIN)/operator-sdk version is not expected $(OPERATOR_SDK_VERSION). Removing it before installing."; \
                rm -rf $(LOCALBIN)/operator-sdk; \
        fi
        @test -s $(LOCALBIN)/operator-sdk || \
        curl -sSLfo $(LOCALBIN)/operator-sdk https://github.com/operator-framework/operator-sdk/releases/download/$(OPERATOR_SDK_VERSION)/operator-sdk_$(OS)_$(ARCH) && \
        chmod +x $(LOCALBIN)/operator-sdk;
```

Also, if you want to save duplication in Makefile, I'd use something like:
```Makefile
  OS=$(shell go env GOOS)
  ARCH=$(shell go env GOARCH)
  define install-binary
        @if ! test -x $(1) || ! $(1) version 2>/dev/null | grep -q $(2); then \
                echo "Installing $(notdir $(1)) $(2)..."; \
                curl -sSL $(3) -o $(1); \
                if file $(1) | grep -q 'gzip'; then \
                        mv $(1) $(1).tar.gz && \
                        tar xzf $(1).tar.gz -C $(dir $(1)) $(notdir $(1)) && \
                        rm $(1).tar.gz; \
                fi; \
                chmod +x $(1); \
        fi
  endef

  $(KIND): | $(BIN_DIR)
        $(call install-binary,$@,$(KIND_VERSION),https://github.com/kubernetes-sigs/kind/releases/download
  D_VERSION)/kind-$(OS)-$(ARCH))

  $(CLUSTERADM): | $(BIN_DIR)
        $(call install-binary,$@,$(CLUSTERADM_VERSION),https://github.com/open-cluster-management-io/clust
  releases/download/$(CLUSTERADM_VERSION)/clusteradm_$(OS)_$(ARCH).tar.gz)
```

Makefile
```diff
@@ -181,3 +179,66 @@ undeploy: gen ## Remove the controller from the cluster (PLATFORM=openshift|kind
...
+HUB_KUBECONFIG := $(DEV_KUBE_DIR)/hub.config
+CLUSTER1_KUBECONFIG := $(DEV_KUBE_DIR)/cluster1.config
+CLUSTER2_KUBECONFIG := $(DEV_KUBE_DIR)/cluster2.config
+
+KIND := $(DEV_BIN_DIR)/kind
+CLUSTERADM := $(DEV_BIN_DIR)/clusteradm
+
+DEV_ENV_SCRIPT := $(CURDIR)/hack/dev-env.sh
+DEV_ENV_VARS := DEV_BIN_DIR=$(DEV_BIN_DIR) DEV_KUBE_DIR=$(DEV_KUBE_DIR) \
+	KIND_VERSION=$(KIND_VERSION) CLUSTERADM_VERSION=$(CLUSTERADM_VERSION) K8S_VERSION=$(K8S_VERSION) \
```
mkolesnik:
```suggestion
  export DEV_BIN_DIR DEV_KUBE_DIR KIND_VERSION CLUSTERADM_VERSION K8S_VERSION OLM_VERSION
```

Any reason not use make's built in variable passing capability?

Makefile
```diff
@@ -181,3 +179,66 @@ undeploy: gen ## Remove the controller from the cluster (PLATFORM=openshift|kind
...
 help: ## Display this help
 	@grep -E '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-20s\033[0m %s\n", $$1, $$2}'
+
+# Dev Environment (local Kind + OCM clusters)
+KIND_VERSION ?= v0.31.0
+CLUSTERADM_VERSION ?= v1.2.0
+K8S_VERSION ?= v1.30.0
+OLM_VERSION ?= v0.42.0
+
+DEV_BIN_DIR := $(CURDIR)/.bin
```
mkolesnik: We already have BIN_DIR why not use that?

Or maybe a sub-directory under that if you really want to separate the toolchain (though I don't see a real reason to do so, it already houses `envtest` for integ tests)

hack/dev-env.sh
```diff
@@ -0,0 +1,368 @@
...
+KIND="${DEV_BIN_DIR}/kind"
+CLUSTERADM="${DEV_BIN_DIR}/clusteradm"
+
+HUB_KUBECONFIG="${DEV_KUBE_DIR}/hub.config"
+CLUSTER1_KUBECONFIG="${DEV_KUBE_DIR}/cluster1.config"
+CLUSTER2_KUBECONFIG="${DEV_KUBE_DIR}/cluster2.config"
+
+HUB_NAME="hub"
+CLUSTER1_NAME="cluster1"
+CLUSTER2_NAME="cluster2"
```
mkolesnik: These should all be passed from the Makefile also

FilipB (commented):
Thank you. LGTM

Makefile
```diff
@@ -181,3 +179,66 @@ undeploy: gen ## Remove the controller from the cluster (PLATFORM=openshift|kind
...
+DEV_KUBE_DIR := $(CURDIR)/.kube
+
+HUB_KUBECONFIG := $(DEV_KUBE_DIR)/hub.config
+CLUSTER1_KUBECONFIG := $(DEV_KUBE_DIR)/cluster1.config
+CLUSTER2_KUBECONFIG := $(DEV_KUBE_DIR)/cluster2.config
```
mkolesnik: These two aren't actually used in the Makefile

Makefile
```diff
@@ -181,3 +179,66 @@ undeploy: gen ## Remove the controller from the cluster (PLATFORM=openshift|kind
...
+	$(CONTAINER_ENGINE) save $(IMG) -o $(DEV_KUBE_DIR)/.addon-image.tar
+	$(KIND) load image-archive $(DEV_KUBE_DIR)/.addon-image.tar --name hub
+	rm -f $(DEV_KUBE_DIR)/.addon-image.tar
+	kubectl --kubeconfig=$(HUB_KUBECONFIG) apply -f config/crd/
+	$(KUSTOMIZE) build config/deploy/overlays/kind | \
+		sed 's|image: quay.io/sail-dev/multicluster-mesh-addon:.*|image: $(IMG)|' | \
+		kubectl --kubeconfig=$(HUB_KUBECONFIG) apply -f -
+	kubectl --kubeconfig=$(HUB_KUBECONFIG) rollout status deployment/multicluster-mesh-controller \
+		-n multicluster-mesh-system --timeout=180s
+	kubectl --kubeconfig=$(HUB_KUBECONFIG) apply -f samples/kind-dev.yaml
```
mkolesnik: I'm wondering if this shouldn't be in the script as well, seeing as it's pretty tightly coupled to how the rest of it works

hack/dev-env.sh
```diff
@@ -0,0 +1,368 @@
...
+
+    local kind_node_image="kindest/node:${K8S_VERSION}"
+
+    for cluster_name in "${HUB_NAME}" "${CLUSTER1_NAME}" "${CLUSTER2_NAME}"; do
+        local kubeconfig
+        case "${cluster_name}" in
+            "${HUB_NAME}")      kubeconfig="${HUB_KUBECONFIG}" ;;
+            "${CLUSTER1_NAME}") kubeconfig="${CLUSTER1_KUBECONFIG}" ;;
+            "${CLUSTER2_NAME}") kubeconfig="${CLUSTER2_KUBECONFIG}" ;;
+        esac
```
mkolesnik: This pattern repeats a few times, I think it'll ber clearer as a func:

```bash
  kubeconfig_for() {
      echo "${DEV_KUBE_DIR}/${1}.config"
  }
```

hack/dev-env.sh
```diff
@@ -0,0 +1,368 @@
...
+metadata:
+  name: klusterlet-work-olm
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: klusterlet-work-olm
+subjects:
+  - kind: ServiceAccount
+    name: klusterlet-work-sa
+    namespace: open-cluster-management-agent
```
mkolesnik: Seeing as this is entirely static, it would be better to store in a static yaml file (which can be edited/linted more easily)

hack/dev-env.sh
```diff
@@ -0,0 +1,368 @@
...
+        deployment/cluster-manager -n open-cluster-management --timeout=120s
+}
+
+join_clusters() {
+    if [[ ! -f "${DEV_KUBE_DIR}/.ocm-join-cmd" ]]; then
+        err "OCM join command not found. Run 'make init-ocm' first."
+    fi
+
+    local join_cmd
+    join_cmd="$(cat "${DEV_KUBE_DIR}/.ocm-join-cmd")"
```
mkolesnik: Instead of doing this in a roundabout way, wouldn't it be possible to use `cvlusteradm` to get the details and call join directly?

Something like:
```bash
  local token_json
  token_json="$(${CLUSTERADM} get token --kubeconfig="${HUB_KUBECONFIG}" -o json)"
  local hub_token hub_apiserver
  hub_token="$(echo "${token_json}" | jq -r '.hub-token')"
  hub_apiserver="$(echo "${token_json}" | jq -r '.hub-apiserver')"
  ${CLUSTERADM} join \
      --hub-token "${hub_token}" \
      --hub-apiserver "${hub_apiserver}" \
      --cluster-name "${cluster_name}" \
      --force-internal-endpoint-lookup \ 
      --kubeconfig="${kubeconfig}"
```

(optionally could use --wait if we want to wait until it joined)

hack/dev-env.sh
```diff
@@ -0,0 +1,368 @@
...
+# Provisions and manages a local 3-cluster Kind/OCM development environment.
+# This file is invoked by Makefile targets with an action argument.
+#
+# Usage: hack/dev-env.sh <action> [options]
+# Actions: install-deps, create-clusters, install-olm, init-ocm, join-clusters, clean
+
+set -euo pipefail
+
+# Required environment variables that should be passed from Makefile
+required_vars=(DEV_BIN_DIR DEV_KUBE_DIR KIND_VERSION CLUSTERADM_VERSION K8S_VERSION OLM_VERSION)
```
mkolesnik: If you export them from Makefile they should be set, this becomes a bit redundant.

mkolesnik (commented):
Initial partial review

hack/dev-env.sh
sridhargaddam: Okay, let me move this to the Makefile.

hack/dev-env.sh
sridhargaddam: Agree, let me create a yaml file and use it in the script.

Makefile
sridhargaddam: The deploy-addon reuses some of the existing Makefile targets as prerequisites. Moving it to the script would mean duplicating the build logic in bash. 
The script currently handles infrastructure (clusters, OCM, OLM) and the Makefile handles build & deploy (compile, image, kustomize, apply), so I'd prefer to keep it this way.

sridhargaddam (commented):
Thanks for the review @mkolesnik.

Makefile
```diff
@@ -181,3 +179,89 @@ undeploy: gen ## Remove the controller from the cluster (PLATFORM=openshift|kind
...
 	@grep -E '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-20s\033[0m %s\n", $$1, $$2}'
+
+# Dev Environment (local Kind + OCM clusters)
+KIND_VERSION ?= v0.31.0
+CLUSTERADM_VERSION ?= v1.2.0
+K8S_VERSION ?= v1.30.0
+OLM_VERSION ?= v0.42.0
+
+OS := $(shell uname | tr '[:upper:]' '[:lower:]')
+ARCH := $(shell uname -m | sed 's/x86_64/amd64/;s/aarch64/arm64/')
```
mkolesnik: Q: Why use `uname` and not `go env` to determine these values?
They should handle the values as expected, no need to further process them

Makefile
```diff
@@ -181,3 +179,89 @@ undeploy: gen ## Remove the controller from the cluster (PLATFORM=openshift|kind
...
+
+.PHONY: join-clusters
+join-clusters: $(CLUSTERADM) ## Register managed clusters and create ManagedClusterSet
+	$(DEV_ENV_SCRIPT) join-clusters
+
+.PHONY: deploy-addon
+deploy-addon: $(KIND) images gen $(KUSTOMIZE) ## Build and deploy addon to the hub Kind cluster
+	# We use image-archive instead of docker-image because the latter is Docker-specific
+	# and fails when images are built with Podman (separate image stores).
+	$(CONTAINER_ENGINE) save $(IMG) -o $(DEV_KUBE_DIR)/.addon-image.tar
```
mkolesnik, line 248: I faced an edge case when running this on the original commit chain..

This fails when `$(IMG)` re-evaluates to a different tag than what images built.
if gen (or any other prerequisite) dirties the tree between images and here, `git describe --dirty` picks up the new state.

This can be fixed by pinning the tag: `$(eval DEPLOY_IMG := $(IMG))` at the top of this target and using $(DEPLOY_IMG) throughout.
Not sure how critical this is, as its very esoteric...

OR you can put the logic in the script, as I suggested, and it should be fine AFAICT :)

Makefile
```diff
@@ -181,3 +179,89 @@ undeploy: gen ## Remove the controller from the cluster (PLATFORM=openshift|kind
 .PHONY: help
 help: ## Display this help
 	@grep -E '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-20s\033[0m %s\n", $$1, $$2}'
+
+# Dev Environment (local Kind + OCM clusters)
+KIND_VERSION ?= v0.31.0
+CLUSTERADM_VERSION ?= v1.2.0
+K8S_VERSION ?= v1.30.0
+OLM_VERSION ?= v0.42.0
```
mkolesnik:
```suggestion
CLUSTERADM_VERSION ?= v1.3.1
K8S_VERSION ?= v1.35.0
OLM_VERSION ?= v0.43.0
```

Unless theres some reason you used these specific versions?

Makefile
```diff
@@ -181,3 +179,89 @@ undeploy: gen ## Remove the controller from the cluster (PLATFORM=openshift|kind
...
+		echo "kind $(KIND_VERSION) already installed"; \
+	else \
+		echo "Installing kind $(KIND_VERSION)..."; \
+		curl -sSL https://github.com/kubernetes-sigs/kind/releases/download/$(KIND_VERSION)/kind-$(OS)-$(ARCH) -o $(KIND); \
+		chmod +x $(KIND); \
+	fi
+
+$(CLUSTERADM): | $(BIN_DIR)
+	@ver="$$($(CLUSTERADM) version 2>/dev/null || true)"; \
+	if test -x $(CLUSTERADM) && echo "$$ver" | grep -q $(CLUSTERADM_VERSION); then \
```
mkolesnik: nit:
```suggestion
	@if test -x $(CLUSTERADM) && $(CLUSTERADM) version 2>/dev/null | grep -q $(CLUSTERADM_VERSION); then \
```

Makefile
```diff
@@ -181,3 +179,89 @@ undeploy: gen ## Remove the controller from the cluster (PLATFORM=openshift|kind
...
+	fi
+
+DEV_ENV_SCRIPT := $(CURDIR)/hack/dev-env.sh
+
+export DEV_KUBE_DIR K8S_VERSION OLM_VERSION
+export KIND CLUSTERADM
+export HUB_NAME CLUSTER1_NAME CLUSTER2_NAME
+
+.PHONY: dev-env
+dev-env: $(KIND) $(CLUSTERADM) create-clusters install-olm init-ocm join-clusters deploy-addon ## Provision full dev environment (Kind + OCM + addon)
```
mkolesnik: nit: no need to explicitly depend on the binaries here, they're already "pulled" transitively as needed

```suggestion
dev-env: create-clusters install-olm init-ocm join-clusters deploy-addon ## Provision full dev environment (Kind + OCM + addon)
```

Makefile
```diff
@@ -181,3 +179,89 @@ undeploy: gen ## Remove the controller from the cluster (PLATFORM=openshift|kind
...
+
+OS := $(shell uname | tr '[:upper:]' '[:lower:]')
+ARCH := $(shell uname -m | sed 's/x86_64/amd64/;s/aarch64/arm64/')
+
+DEV_KUBE_DIR := $(CURDIR)/.kube
+HUB_KUBECONFIG := $(DEV_KUBE_DIR)/hub.config
+
+HUB_NAME := hub
+CLUSTER1_NAME := cluster1
+CLUSTER2_NAME := cluster2
```
mkolesnik: Not urgent but I'd put these only in the script, as they're no used in the makefile itself

Also you can drop the `_NAME` suffix IMHO its stuttering and without it its easier to read

hack/dev-env.sh
```diff
@@ -0,0 +1,267 @@
...
+
+create_clusters() {
+    mkdir -p "${DEV_KUBE_DIR}"
+
+    local existing_clusters
+    existing_clusters="$(${KIND} get clusters 2>/dev/null || true)"
+    local found=()
+    for name in "${HUB_NAME}" "${CLUSTER1_NAME}" "${CLUSTER2_NAME}"; do
+        if echo "${existing_clusters}" | grep -qx "${name}"; then
+            found+=("${name}")
```
mkolesnik: nit:

```suggestion
    for cluster in "${HUB_NAME}" "${CLUSTER1_NAME}" "${CLUSTER2_NAME}"; do
        if echo "${existing_clusters}" | grep -qx "${cluster}"; then
            found+=("${cluster}")
```

hack/dev-env.sh
```diff
@@ -0,0 +1,267 @@
...
+
+        log "Creating Kind cluster: ${cluster_name}"
+        ${KIND} create cluster \
+            --name "${cluster_name}" \
+            --kubeconfig "${kubeconfig}" \
+            --image "${kind_node_image}" \
+            --wait 120s
+
+        log "Waiting for cluster ${cluster_name} API to be ready..."
+        kubectl --kubeconfig="${kubeconfig}" wait --for=condition=Ready nodes --all --timeout=120s
```
mkolesnik: nit: I think "cluster_name" is a bit stuttering, we can use just "cluster" and the code is clearer for this (also applies to other iterations which I won't bother to make suggestions for).

```suggestion
    for cluster in "${HUB_NAME}" "${CLUSTER1_NAME}" "${CLUSTER2_NAME}"; do
        local kubeconfig
        kubeconfig=$(kubeconfig_for "${cluster}")

        log "Creating Kind cluster: ${cluster}"
        ${KIND} create cluster \
            --name "${cluster}" \
            --kubeconfig "${kubeconfig}" \
            --image "${kind_node_image}" \
            --wait 120s

        log "Waiting for cluster ${cluster} API to be ready..."
        kubectl --kubeconfig="${kubeconfig}" wait --for=condition=Ready nodes --all --timeout=120s
```

hack/dev-env.sh
```diff
@@ -0,0 +1,267 @@
...
+        err "Hub kubeconfig not found at ${hub_kubeconfig}. Run 'make init-ocm' first."
+    fi
+
+    log "Retrieving hub token..."
+    local token_output
+    token_output="$(${CLUSTERADM} get token --kubeconfig="${hub_kubeconfig}" 2>/dev/null)"
+
+    local hub_token hub_apiserver
+    hub_token="$(echo "${token_output}" | grep -oP '(?<=--hub-token )\S+')"
+    hub_apiserver="$(echo "${token_output}" | grep -oP '(?<=--hub-apiserver )\S+')"
```
mkolesnik: This is fragile and could fail if `grep` doesn't support Perl syntax (e.g. on macOS).

I suggest we use `jq` which is more robust.

```suggestion
    local token_json hub_token hub_apiserver
    token_json="$(${CLUSTERADM} get token --kubeconfig="${hub_kubeconfig}" -o json)"
    hub_token="$(echo "${token_json}" | jq -r '.["hub-token"]')"
    hub_apiserver="$(echo "${token_json}" | jq -r '.["hub-apiserver"]')"
```

hack/dev-env.sh
```diff
@@ -0,0 +1,267 @@
...
+
+        # Check if OLM is already installed
+        if kubectl --kubeconfig="${kubeconfig}" get deployment olm-operator -n olm &>/dev/null; then
+            log "OLM already installed on ${cluster_name}, skipping"
+            continue
+        fi
+
+        log "Installing OLM ${OLM_VERSION} on ${cluster_name}..."
+
+        log "Applying OLM CRDs on ${cluster_name}..."
```
mkolesnik: nit: IMHO this log doesn't add any valuable info, since its the first one (arguably all the logs until "installed" seem redundant to me).

```suggestion
```

hack/dev-env.sh
```diff
@@ -0,0 +1,267 @@
...
+
+        log "OLM ${OLM_VERSION} installed on ${cluster_name}"
+    done
+
+    # Grant the OCM work agent (klusterlet-work-sa) permission to manage OLM resources.
+    local script_dir
+    script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+    for cluster_name in "${CLUSTER1_NAME}" "${CLUSTER2_NAME}"; do
+        local kubeconfig
+        kubeconfig="$(kubeconfig_for "${cluster_name}")"
```
mkolesnik: nit: This can be inlined

hack/dev-env.sh
```diff
@@ -0,0 +1,267 @@
...
+    hub_kubeconfig="$(kubeconfig_for "${HUB_NAME}")"
+
+    if [[ ! -f "${hub_kubeconfig}" ]]; then
+        err "Hub kubeconfig not found at ${hub_kubeconfig}. Run 'make create-clusters' first."
+    fi
+
+    log "Initializing OCM hub on cluster: ${HUB_NAME}"
+    ${CLUSTERADM} init --wait --kubeconfig="${hub_kubeconfig}"
+
+    log "Waiting for OCM hub components to be ready..."
```
mkolesnik, line 113: nit: I'd log success at the end instead of this

hack/dev-env.sh
```diff
@@ -0,0 +1,267 @@
...
+  clusterSelector:
+    selectorType: ExclusiveClusterSetLabel
+EOF
+
+    for cluster_name in "${CLUSTER1_NAME}" "${CLUSTER2_NAME}"; do
+        log "Labeling ${cluster_name} with clusterset=mesh-cluster-set"
+        kubectl --kubeconfig="${hub_kubeconfig}" label managedcluster "${cluster_name}" \
+            cluster.open-cluster-management.io/clusterset=mesh-cluster-set \
+            --overwrite
+    done
```
mkolesnik: `clusteradm` can handle this directly:

```suggestion
      ${CLUSTERADM} --kubeconfig="${hub_kubeconfig}" create clusterset mesh-cluster-set
      ${CLUSTERADM} --kubeconfig="${hub_kubeconfig}" clusterset set mesh-cluster-set --clusters "${CLUSTER1_NAME},${CLUSTER2_NAME}" 
```

hack/dev-env.sh
```diff
@@ -0,0 +1,267 @@
...
+
+        log "Creating product ClusterClaim on ${cluster_name}"
+        kubectl --kubeconfig="${kubeconfig}" apply -f - <<EOF
+apiVersion: cluster.open-cluster-management.io/v1alpha1
+kind: ClusterClaim
+metadata:
+  name: product.open-cluster-management.io
+spec:
+  value: Kind
+EOF
```
mkolesnik: This should be a static YAML

hack/dev-env.sh
```diff
@@ -0,0 +1,267 @@
...
+    done
+
+    # On OpenShift, the product ClusterClaim is created automatically by the
+    # klusterlet agent. On vanilla Kind clusters there is no such agent,
+    # so we create it manually. The OCM registration agent syncs it to
+    # ManagedCluster.status.clusterClaims on the hub, which the addon controller
+    # uses for platform detection.
+    for cluster_name in "${CLUSTER1_NAME}" "${CLUSTER2_NAME}"; do
+        local kubeconfig
+        kubeconfig="$(kubeconfig_for "${cluster_name}")"
```
mkolesnik: nit: this can be inlined

hack/dev-env.sh
```diff
@@ -0,0 +1,267 @@
...
+            if [[ -n "${claims}" ]]; then
+                log "Cluster ${cluster_name} product claim synced: ${claims}"
+                break
+            fi
+            retries=$((retries + 1))
+            sleep 2
+        done
+        if [[ ${retries} -ge 30 ]]; then
+            err "Timed out waiting for product claim on ${cluster_name} to propagate to hub"
+        fi
```
mkolesnik: This can be replaced with a simple `kubectl --wait`:

```suggestion
    log "Waiting for product claim on ${cluster_name} to propagate to the hub..."
    kubectl --kubeconfig="${hub_kubeconfig}" wait managedcluster/"${cluster_name}" \
    --for='jsonpath={.status.clusterClaims[?(@.name=="product.open-cluster-management.io")].value}=Kind' \
    --timeout=60s
```

Makefile
```diff
@@ -181,3 +179,89 @@ undeploy: gen ## Remove the controller from the cluster (PLATFORM=openshift|kind
...
+	$(CONTAINER_ENGINE) save $(IMG) -o $(DEV_KUBE_DIR)/.addon-image.tar
+	$(KIND) load image-archive $(DEV_KUBE_DIR)/.addon-image.tar --name hub
+	rm -f $(DEV_KUBE_DIR)/.addon-image.tar
+	kubectl --kubeconfig=$(HUB_KUBECONFIG) apply -f config/crd/
+	$(KUSTOMIZE) build config/deploy/overlays/kind | \
+		sed 's|image: quay.io/sail-dev/multicluster-mesh-addon:.*|image: $(IMG)|' | \
+		kubectl --kubeconfig=$(HUB_KUBECONFIG) apply -f -
+	kubectl --kubeconfig=$(HUB_KUBECONFIG) rollout status deployment/multicluster-mesh-controller \
+		-n multicluster-mesh-system --timeout=180s
+	kubectl --kubeconfig=$(HUB_KUBECONFIG) apply -f samples/kind-dev.yaml
```
mkolesnik: Are we sure we want to always deploy this sample mesh?

Specifically in E2E tests, but also local dev, one might not always want the sample deployed automatically

hack/dev-env.sh
```diff
@@ -0,0 +1,267 @@
...
+    existing_clusters="$(${KIND} get clusters 2>/dev/null || true)"
+    local found=()
+    for name in "${HUB_NAME}" "${CLUSTER1_NAME}" "${CLUSTER2_NAME}"; do
+        if echo "${existing_clusters}" | grep -qx "${name}"; then
+            found+=("${name}")
+        fi
+    done
+
+    if [[ ${#found[@]} -gt 0 ]]; then
+        err "Kind clusters already exist: ${found[*]}. Run 'make dev-clean' to tear them down first."
```
mkolesnik, line 36: Other actions are idempotent but this one isn't, are we sure we want to error out?

README.md
```diff
@@ -104,6 +104,50 @@ kubectl create namespace multicluster-mesh-system
...
+
+**Configuration (override via environment or command-line):**
+
+```bash
+make dev-env K8S_VERSION=v1.31.0 OLM_VERSION=v0.42.0
+```
+
+##### Known Issues
+
+**"Too many open files" on Fedora/Linux:**
```
mkolesnik: nit: 
```suggestion
**"Too many open files" on Linux:**
```

README.md
```diff
@@ -104,6 +104,50 @@ kubectl create namespace multicluster-mesh-system
...
+Kind clusters may fail to start or pods may crash with `too many open files` errors due to low inotify limits. Increase them on the host:
```
mkolesnik, line 135: nit: We can have this check in the makefile or the script, to warn the user this is a potential problem (maybe at the end of create clusters)

Makefile
mkolesnik: Yes and no, the prereqs in makefile should stay but almost the logic can all be in the script as it's pretty tightly coupled to it.
Perhaps the first line should stay in makefile, though you can simple export CONTAINER_ENGINE as well

hack/dev-env.sh
sridhargaddam: Yes, this was intentional. If there is already and existing deployment, the idea is to give an opportunity to the user to decide if he wants to tear it down first (using make dev-clean) instead of the scripts doing it internally.

hack/dev-env.sh
sridhargaddam: Okay, let me remove this log.

Makefile
sridhargaddam: Okay, let me remove this step. If required, we can consider adding a new makefile target to deploy a sample mesh.

README.md
sridhargaddam: I'd like to handle this in a follow up PR.

hack/dev-env.sh
sridhargaddam: It usually takes sometime for the Hub components to come up, so a log message would help. Let me add a success log.

Makefile
sridhargaddam: Hope you are okay with handling this in some future PR? Please feel free to push a fix, once this PR lands.

hack/dev-env.sh
mkolesnik: Fair, we can revisit this

mkolesnik (approved):

PR #83, opened 2026-05-25
Scope ManifestWork cleanup to a single ClusterSet

mkolesnik:
Cleanup is now scoped per ClusterSet, implementing the design's tear-down + bootstrap behavior for cluster movement. When a cluster leaves a ClusterSet, its ManifestWorks are cleaned up by the old mesh. If the cluster joins a new ClusterSet with a mesh, the new mesh bootstraps it with fresh configuration.

ManifestWorks are labeled with the ClusterSet they belong to, and cleanup filters by this label to avoid affecting other ClusterSets.

Depends-On: #87


pkg/hub/mesh/controller.go
```diff
@@ -299,8 +299,8 @@ func (r *Reconciler) ensureOperatorInstalled(ctx context.Context, mesh *meshv1al
 }
 
 // cleanupOperatorManifestWorks deletes operator ManifestWorks on clusters that no mesh needs anymore.
-func (r *Reconciler) cleanupOperatorManifestWorks(ctx context.Context) error {
-	neededClusters, err := r.getClustersNeededByAnyMesh(ctx)
+func (r *Reconciler) cleanupOperatorManifestWorks(ctx context.Context, clusterSet string) error {
+	neededClusters, err := r.getClustersNeededInClusterSet(ctx, clusterSet)
```
sridhargaddam, line 397: Here, `neededClusters` is created only for the current clusterSet, but L309 lists ManifestWorks across all clusterSets. This could result in deleting ManifestWorks from a different clusterSet in L319. Please take a look.

mkolesnik: Great catch :fishing_pole_and_fish: thanks!

mkolesnik: I fixed it, PTAL

mkolesnik:
The scope of the PR updated a bit, this is not only optimization this is the intended behavior based on the design.
Hence the test for moving a cluster has been updated to validate the behavior more thoroughly.

Since this now depends on #87, review only the last commit - d90f746a92fcba98b782a44e1fc0ee8d6a3ac588

sridhargaddam (approved):

PR #85, opened 2026-05-25
Extract helpers for iterating over clusters

mkolesnik:
Extracting the repetitive pattern of listing non-deleting meshes targeting a given ClusterSet and iterating over them.
This simplifies the code and makes sure places that use this processing pattern are consistent.

Depends on: #83


jewertow (commented):
LGTM - will approve once you remove `hold`.

pkg/hub/mesh/controller.go
```diff
@@ -521,17 +510,11 @@ func (r *Reconciler) setErrorStatus(mesh *meshv1alpha1.MultiClusterMesh, reconci
 func (r *Reconciler) getClustersNeededInClusterSet(ctx context.Context, clusterSet string) (map[string]bool, error) {
```
jewertow: I know that we aproved this function in a previous PR, so this comment may be little annoying, but I had a problem to understand why do we have this function, because this name is very confusing to me.
What do you think about:
```suggestion
func (r *Reconciler) getActiveMeshClusters(ctx context.Context, clusterSet string) (map[string]bool, error) {
```

mkolesnik: Not annoying at all, naming is important.

I think `getActiveMeshClusters` is a bit misleading though. The function doesn't return "clusters with active meshes", it returns all clusters in the ClusterSet when any active mesh targets the set (all-or-nothing for operator cleanup). The name reads like it filters clusters by mesh presence.

You're right that `InClusterSet` stutters with the parameter. How about just `getNeededClusters`? Shorter, and "needed" is accurate since it's used to decide which clusters keep their operator `ManifestWork`.

jewertow: > I think getActiveMeshClusters is a bit misleading though. The function doesn't return "clusters with active meshes"

Right, it was inspired by `hasActiveMesh` variable.

> How about just getNeededClusters?

It's better, but I still must jump to the implementation to get the sense of this function. Maybe:
```suggestion
func (r *Reconciler) getMeshEnabledClusters(ctx context.Context, clusterSet string) (map[string]bool, error) {
```

jewertow (approved):
Thank you for addressing my suggestion!

PR #87, opened 2026-05-27
Handle ManifestWork updates using OCM sdk-go WorkApplier

mkolesnik:
Previously the controller only created ManifestWorks and skipped them if they already existed. This meant config changes (e.g. operator channel) and external modifications were never corrected.

This replaces the hand-rolled create-only logic with the OCM sdk-go WorkApplier, which handles create, update, and delete via JSON merge patch. The controller now correctly reconciles ManifestWork spec and labels back to the desired state on every reconcile.

One known limitation: the WorkApplier cache doesn't detect external label-only changes (reported as https://github.com/open-cluster-management-io/sdk-go/issues/223). A test for this case is pending until the fix is released.


pkg/hub/mesh/controller.go
```diff
@@ -90,9 +93,25 @@ func RegisterController(mgr manager.Manager) error {
...
+	workClient, err := workclient.NewForConfig(mgr.GetConfig())
+	if err != nil {
+		return fmt.Errorf("failed to create work client: %w", err)
+	}
+
+	workInformerFactory := workinformers.NewSharedInformerFactory(workClient, 0)
+	workLister := workInformerFactory.Work().V1().ManifestWorks().Lister()
+
+	if err := mgr.Add(manager.RunnableFunc(func(ctx context.Context) error {
+		workInformerFactory.Start(ctx.Done())
```
jewertow, line 105: Don't you need to wait for the informer to sync and then block until the context is done?
```suggestion
		workInformerFactory.Start(ctx.Done())
        workInformerFactory.WaitForCacheSync(ctx.Done())
        <-ctx.Done()
````

mkolesnik: Thanks, good catch I fixed it!

jewertow (approved):

Issue #72, opened 2026-05-18
Ensure unique ManifestWork naming to prevent multi-tenant collisions


jewertow:
Currently, the controller distributes certificates using a hardcoded ManifestWork resource name: multicluster-mesh-cacerts. In multi-tenant environments where multiple MultiClusterMesh (MCM) instances exist across different namespaces, this static naming results in resource conflicts, as tenants try to overwrite the same ManifestWork.

To support multi-tenancy, we need a dynamic naming convention that incorporates the owner's identity (Name and Namespace). However, a naive string concatenation (e.g., namespace-name) is susceptible to collision edge cases due to how hyphens can blend boundaries.

If we use a simple hyphenated delimiter, two entirely different tenants will resolve to the exact same resource name:

- Tenant 1: Namespace: `mesh-prod`, Name: `cluster-a` -> `mesh-prod-cluster-a`
- Tenant 2: Namespace: `mesh`, Name: `prod-cluster-a` -> `mesh-prod-cluster-a`

mkolesnik:
This is not limited to certs, MSA also needs unique names.

Something like `<mesh-namespace>-<mesh-name>-istio-reader` in that case.

We should devise a consistent naming scheme.

mkolesnik:
I consulted with Gemini about this:

For safety and deterministic behavior, **keep `ManifestWork` and `ManagedServiceAccount` names to 63 characters or less** (standard DNS Label limits), even though the Hub API technically allows more. 

<details>
<summary><b>Click here for the technical breakdown of why</b></summary>

While the Hub cluster API allows 253 characters (DNS Subdomain Standard) for these CRDs, using long names will silently break downstream pipelines:

* **`ManifestWork` (188-Character Hard Trap):** The Klusterlet agent on the managed cluster creates an `AppliedManifestWork` to track status. This is hardcoded as `<64-character-hub-hash>-<manifestwork-name>`. If your original name is 189+ characters, the Hub accepts it, but the agent silently fails to create the applied tracker due to the 253-character CRD limit.
* **`ManagedServiceAccount` (Downstream Controller Crashes):** This CRD triggers a pipeline that generates downstream `ServiceAccounts`, `Secrets`, and `ManifestWorks`. Because OCM controllers concatenate your original name with suffixes and random hashes (e.g., `<msa-name>-token-xxxx`), long names will easily push those generated objects over the 253-character limit, crashing the controllers.

</details>

yxun:
[...]

PR #93, opened 2026-06-01
Clarify design doc on multi-tenancy, collisions, and lifecycle

mkolesnik:
- Clarify brownfield non-goal: operator adoption is supported,
  only mesh configuration adoption is excluded
- Add multi-mesh same ClusterSet example with different namespaces
- Add deterministic naming strategy paragraph, link to #72
- Detail operator Installation Workflow (MCV detection, adoption)
- Detail Collision Handling (hub-side and spoke-side)
- Clarify ClusterSet-scoped operator cleanup
- Clarify trust domain is a naming convention in Phase 1
- Switch tokenValidity to Go duration format with 360h default
- Add cleanup step for cluster removal in Endpoint Discovery
- Add ManagedClusterView and #72 reference links
- Document spec.clusterSet as immutable after creation

Signed-off-by: Mike Kolesnik <mkolesni@redhat.com>


docs/design.md
```diff
@@ -124,7 +125,7 @@ The MVP supports the [Multi-Primary Multi-Network] mesh topology. This aligns wi
 
 | Field | Required | Description |
 |-------|----------|-------------|
-| `spec.clusterSet` | Yes | Name of the [ManagedClusterSet] defining cluster membership |
+| `spec.clusterSet` | Yes | Name of the [ManagedClusterSet] defining cluster membership (immutable after creation) |
```
yxun, line 128: Does the `immutable` here mean that users cannot change the `spec.clusterSet` value from the MCM CR after creation ? 
What about the case when the user deletes a ClusterSet object manually from a hub ? Will that case cause errors in the mesh and following reconciliation ?

docs/design.md
```diff
@@ -133,7 +134,7 @@ The MVP supports the [Multi-Primary Multi-Network] mesh topology. This aligns wi
 | `spec.operator.startingCSV` | No | Pin to a specific operator version |
 | `spec.operator.installPlanApproval` | No | `Automatic` or `Manual` (default: `Automatic`) |
 | `spec.security.trust.certManager.issuerRef.name` | No | cert-manager Issuer name for Root CA |
-| `spec.security.discovery.tokenValidity` | No | ManagedServiceAccount token lifetime (default: `1m`) |
+| `spec.security.discovery.tokenValidity` | No | ManagedServiceAccount token lifetime, Go duration format (default: `360h`) |
```
yxun: The `ManagedServiceAccount` CRD requires a token validity duration no less than 10 minutes. We may also add a note about no less than 10 minutes about the `spec.security.discovery.tokenValidity`  here.

docs/design.md
```diff
@@ -157,23 +158,42 @@ spec:
...
+The `spec.clusterSet` field is immutable after creation. With exclusive ClusterSets, changing the reference means an entirely different set of clusters. All plumbing is cluster-specific, so nothing carries over, making migration equivalent to deleting and recreating the mesh. Users who need a different ClusterSet should delete the mesh CR and create a new one.
+
+`MultiClusterMesh` is namespace-scoped, enabling tenant isolation on the hub. Each mesh operates independently - its certificates, discovery tokens, and operator configuration are scoped to its namespace. Multiple meshes can target the same ClusterSet, provided they use different control plane namespaces. For example, Mesh A targets ClusterSet X with namespace `istio-system-a`, while Mesh B targets the same ClusterSet X with namespace `istio-system-b`. Each mesh gets its own trust domain, certificates, and discovery tokens. If two meshes target the same control plane namespace on the same ClusterSet, the older resource (by creation timestamp) wins and the newer one is rejected.
```
yxun, line 170: Should we also include the istio-cni namespace here as well ?
`If two meshes target the same istio CNI namespace on the same ClusterSet, the older resource (by creation timestamp) wins and the newer one is rejected.`

docs/design.md
```diff
@@ -187,18 +207,24 @@ The add-on implements Istio's [Plug-in CA] pattern:
...
 ## Endpoint Discovery
 
 For multi-primary mesh topologies, each control plane needs API access to its peers. The add-on automates this using [ManagedServiceAccount]:
 
-1. Creates a `ManagedServiceAccount` per cluster per mesh, yielding short-lived tokens
+1. Creates a `ManagedServiceAccount` per cluster per mesh, yielding short-lived tokens. See [#72] for the naming convention discussion.
 2. Constructs kubeconfig-style remote secrets from these tokens
 3. Distributes remote secrets to all peer clusters in the mesh
 4. Token rotation is handled automatically by the OCM platform
+5. When a cluster is removed from the mesh, its MSA is deleted and its remote secrets are removed from all peers
```
yxun, line 222: >  When a cluster is removed from the mesh

Does that mean a cluster is removed from a clusterSet ? Currently, I only implemented the case of cleanup MSA and remote secrets when the controller reconciles and finds a cluster is not in the clusterSet.

docs/design.md
mkolesnik: Yes, users can't change `spec.clusterSet` after creation. The rationale is in the paragraph below the table (line 168): changing the set means a completely different set of clusters, nothing carries over.

If the ClusterSet is deleted, the controller handles it gracefully: `getClustersFromSet` returns an empty list (not an error), so the controller cleans up all ManifestWorks for the set.
The mesh becomes non-functional until the ClusterSet is recreated, but no errors.

docs/design.md
mkolesnik: I will add a note about the 10-minute minimum.

docs/design.md
mkolesnik: IstioCNI is cluster-scoped (the CR name is always `default`), so it's a per-cluster singleton like the operator.
Two meshes on the same cluster would share the same CNI installation.
The collision detection here is about control plane namespaces that the addon creates and manages.

CNI isn't managed by the addon today, so adding collision detection for it would be new design rather than a clarification.
Worth tracking separately if we add CNI management later.

docs/design.md
mkolesnik: Yes, "removed from the mesh" means the cluster is no longer in the ClusterSet (label changed, cluster deleted, etc.).
The controller should reconcile on ClusterSet membership changes and clean up resources for clusters that are no longer in the set.
That should match what you implemented.

docs/design.md
```diff
@@ -157,23 +158,42 @@ spec:
...
+2. **Adoption (operator already present)**: If a compatible Subscription is found, the add-on skips ManifestWork creation. If the configuration is incompatible, the add-on reports a conflict.
+3. **Installation (operator missing)**: If no Subscription is found, the controller creates a [ManifestWork] containing the OLM objects (Namespace, OperatorGroup, Subscription) with platform-specific defaults.
+
+### Collision Handling
+
+The controller handles two types of collisions:
+
+1. **Hub-side (between meshes)**: If two `MultiClusterMesh` resources target the same cluster but request different operator configurations (e.g., different channels or catalog sources), the oldest mesh (by creation timestamp) takes precedence. Newer meshes with conflicting configs are halted with a `ConfigurationConflict` status.
+2. **Spoke-side (pre-existing operator)**: If the ManagedClusterView detects an existing Subscription not created by the add-on, the controller compares the installed configuration against the resolved defaults for the requesting mesh. If compatible, the operator is adopted. If incompatible, the controller halts and reports a `ConfigurationConflict`.
```
sridhargaddam, line 193: > If compatible, the operator is adopted.

Once the operator is adopted, will the lifecycle of the OSSM Operator on the spoke is subsequently handled by the MCM Addon?

docs/design.md
sridhargaddam, line 128: > immutable after creation

The current code does not handle this. So, the plan is to implement this in a future PR?

sridhargaddam:
@mkolesnik in case you missed the review comments....

mkolesnik:
Merged upstream/main, resolved conflict in docs/design.md (tokenValidity description wording, no logical changes).

docs/design.md
mkolesnik: No. Adoption here means the add-on *skips* creating a ManifestWork for the operator.
It uses the existing operator but does not take ownership of it.
The pre-existing operator remains managed by whoever installed it (e.g., the ingress operator, or a manual install).

docs/design.md
mkolesnik: Correct, the current code does not enforce immutability.
This PR documents the design intent.

sridhargaddam (approved):

PR #94, opened 2026-06-01
Refactor CreateMultiClusterMesh to accept optional spec

mkolesnik:
Replace the `OperatorConfig` parameter with a variadic `MeshSpec`, allowing callers to set any spec field (`controlPlane`, `security`, etc.) without needing separate helper functions.
Remove `CreateMultiClusterMeshWithCertManager` since the unified helper now covers that case.
Add `CertManagerSpec` helper to reduce boilerplate for cert-manager config.

Depends-On: #85

PR #95, opened 2026-06-02
Add namespace conflict detection for control plane namespace

mkolesnik:
Two MultiClusterMesh resources targeting the same controlPlane.namespace on the same ClusterSet would push conflicting certificates and discovery tokens into the same namespace.
The older mesh (by creationTimestamp) wins, the newer one gets a NamespaceConflict status condition.

Rename triggerReconcileForBlockedMeshes to triggerReconcileForNotReadyMeshes and broaden it to re-trigger any not-ready mesh in the ClusterSet, which is more robust and future-proof.

Depends-On: #94


mkolesnik:
/hold
while we wait for #94 to merge

yxun (approved):

PR #105, opened 2026-06-02
Add owner reference to Certificates

### LINKED ISSUE #67: [certificate distribution] implement ownership between MultiClusterMesh and Certificate resoruces
Currently, the `MultiClusterMesh` controller creates the `Certificate` resource as a one-time operation but no longer maintains it afterward. We need to implement a proper parent-child relationship within the reconciliation loop so that changes to the `MultiClusterMesh` or drift in the `Certificate` are continuously managed.

jewertow:
Fixes #67

### DIFF
--- a/pkg/hub/mesh/controller.go
+++ b/pkg/hub/mesh/controller.go
@@ -122,6 +122,7 @@ func RegisterController(mgr manager.Manager) error {
 
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&meshv1alpha1.MultiClusterMesh{}).
+		Owns(&certmanagerv1.Certificate{}).
 		Watches(
 			&clusterv1.ManagedCluster{},
 			handler.EnqueueRequestsFromMapFunc(reconciler.findMeshesForCluster),
--- a/pkg/hub/mesh/controller.go
+++ b/pkg/hub/mesh/controller.go
@@ -124,7 +124,7 @@ func RegisterController(mgr manager.Manager) error {
 
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&meshv1alpha1.MultiClusterMesh{}).
-		Owns(&certmanagerv1.Certificate{}).
+		Owns(&certmanagerv1.Certificate{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
 		Watches(
 			&clusterv1.ManagedCluster{},
 			handler.EnqueueRequestsFromMapFunc(reconciler.findMeshesForCluster),
--- a/pkg/hub/mesh/controller.go
+++ b/pkg/hub/mesh/controller.go
@@ -750,27 +750,15 @@ func (r *Reconciler) ensureCertificatesCreated(ctx context.Context, mesh *meshv1
 	return nil
 }
 
-// ensureCertificateForCluster creates a Certificate resource for a specific cluster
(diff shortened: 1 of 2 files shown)


pkg/hub/mesh/controller.go
```diff
@@ -122,6 +122,7 @@ func RegisterController(mgr manager.Manager) error {
 
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&meshv1alpha1.MultiClusterMesh{}).
+		Owns(&certmanagerv1.Certificate{}).
```
sridhargaddam: To ensure that only spec level changes trigger reconciliation, we can add `predicate.GenerationChangedPredicate{}`

pkg/hub/mesh/controller.go
sridhargaddam, line 804: Lets add a similar log for success scenario in the new implementation as well.

pkg/hub/mesh/controller.go
jewertow: Why do you assume that only spec level changes should trigger reconciliation?

pkg/hub/mesh/controller.go
jewertow: Do you mean an equivalent of this?
```
klog.Infof("Successfully created Certificate %s/%s", mesh.Namespace, certName)
```

pkg/hub/mesh/controller.go
sridhargaddam: I'm assuming that if cert-manager updates the status, reconciliation doesn't need to be triggered. Do we want reconciliation to occur even for status-only updates?

pkg/hub/mesh/controller.go
sridhargaddam: Yup

pkg/hub/mesh/controller.go
jewertow: Okay, now I understand your intention. I thought you want to let users modify metadata.

pkg/hub/mesh/controller.go
jewertow: Done

pkg/hub/mesh/controller.go
jewertow: Done

sridhargaddam (approved):

test/integration/controller_test.go
```diff
@@ -461,8 +461,48 @@ var _ = Describe("MultiClusterMesh Controller", func() {
...
+				Eventually(func() string {
+					c := &certmanagerv1.Certificate{}
+					if err := k8sClient.Get(ctx, types.NamespacedName{
+						Name:      fmt.Sprintf("cacerts-%s", clusterName),
+						Namespace: testNs,
+					}, c); err != nil {
+						return ""
+					}
+					return c.Spec.CommonName
+				}).Should(Equal("Intermediate Istio CA"))
```
mkolesnik, line 495: nit: Could use `cert.Name` and `cert.Namespace` from the already-fetched cert.

```suggestion
				Eventually(func() string {
					c := &certmanagerv1.Certificate{}
					if err := k8sClient.Get(ctx, types.NamespacedName{
						Name:      cert.Name,
						Namespace: cert.Namespace,
					}, c); err != nil {
						return ""
					}
					return c.Spec.CommonName
				}).Should(Equal("Intermediate Istio CA"))
```

PR #106, opened 2026-06-02
Skip cacerts distribution for clusters without product claims

sridhargaddam:
The operator installation code skips clusters that are
missing their product claim, but ensureCertificatesCreated
and ensureCacertsDistributed were still processing them.
This PR avoids creating certs and cacerts ManifestWorks
for clusters without the product claims.

### DIFF
--- a/pkg/hub/mesh/controller.go
+++ b/pkg/hub/mesh/controller.go
@@ -273,28 +273,26 @@ func (r *Reconciler) operatorConfigConflicts(a, b meshv1alpha1.OperatorConfig) b
 
 func (r *Reconciler) doReconcile(ctx context.Context, mesh *meshv1alpha1.MultiClusterMesh, clusters []clusterv1.ManagedCluster) (reconcile.Result, error) {
 	for _, cluster := range clusters {
-		klog.V(4).Infof("Reconciling cluster %s", cluster.Name)
-
 		if getProductClaim(&cluster) == "" {
 			klog.V(4).Infof("Cluster %s missing product claim (needed for platform detection), skipping", cluster.Name)
 			continue
 		}
+
+		klog.V(4).Infof("Reconciling cluster %s", cluster.Name)
 
 		work, err := r.workApplier.Apply(ctx, r.buildOperatorManifestWork(mesh, &cluster))
 		if err != nil {
 			return reconcile.Result{}, fmt.Errorf("failed to apply operator ManifestWork on cluster %s: %w", cluster.Name, err)
 		}
 		klog.V(4).Infof("Applied operator ManifestWork %s/%s", work.Namespace, work.Name)
-	}
 
-	// Create certificates for each cluster if cert-manager is configured
-	if mesh.Spec.Security.Trust.CertManager.IssuerRef.Name != "" {
-		if err := r.ensureCertificatesCreated(ctx, mesh, clusters); err != nil {
-			return reconcile.Result{}, err
-		}
-
-		if err := r.ensureCacertsDistributed(ctx, mesh, clusters); err != nil {
-			return reconcile.Result{}, err
+		if mesh.Spec.Security.Trust.CertManager.IssuerRef.Name != "" {
+			if err := r.ensureCertificateForCluster(ctx, mesh, &cluster); err != nil {
+				return reconcile.Result{}, fmt.Errorf("failed to ensure certificate for cluster %s: %w", cluster.Name, err)
+			}
+			if err := r.ensureCacertsManifestWork(ctx, mesh, &cluster); err != nil {
+				return reconcile.Result{}, fmt.Errorf("failed to ensure cacerts ManifestWork for cluster %s: %w", cluster.Name, err)
+			}
 		}
 	}
 
@@ -753,16 +751,6 @@ func getCacertsName(clusterName string) string {
 	return fmt.Sprintf("cacerts-%s", clusterName)
 }
 
-// ensureCertificatesCreated creates Certificate resources for each cluster in the mesh
-func (r *Reconciler) ensureCertificatesCreated(ctx context.Context, mesh *meshv1alpha1.MultiClusterMesh, clusters []clusterv1.ManagedCluster) error {
-	for _, cluster := range clusters {
-		if err := r.ensureCertificateForCluster(ctx, mesh, &cluster); err != nil {
-			return fmt.Errorf("failed to ensure certificate for cluster %s: %w", cluster.Name, err)
-		}
-	}
-	return nil
-}
-
 // ensureCertificateForCluster applies the desired Certificate state for a specific cluster using server-side apply.
 func (r *Reconciler) ensureCertificateForCluster(ctx context.Context, mesh *meshv1alpha1.MultiClusterMesh, cluster *clusterv1.ManagedCluster) error {
 	certName := getCacertsName(cluster.Name)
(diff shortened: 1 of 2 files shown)


pkg/hub/mesh/controller.go
```diff
@@ -255,29 +255,33 @@ func (r *Reconciler) operatorConfigConflicts(a, b meshv1alpha1.OperatorConfig) b
...
-		if err := r.ensureCertificatesCreated(ctx, mesh, clusters); err != nil {
+		if err := r.ensureCertificatesCreated(ctx, mesh, readyClusters); err != nil {
 			return reconcile.Result{}, err
 		}
 
-		if err := r.ensureCacertsDistributed(ctx, mesh, clusters); err != nil {
+		if err := r.ensureCacertsDistributed(ctx, mesh, readyClusters); err != nil {
 			return reconcile.Result{}, err
 		}
 	}
```
mkolesnik: Instead of building a separate `readyClusters` slice and passing it to `ensureCertificatesCreated` / `ensureCacertsDistributed` (which are just thin `for range` wrappers), consider folding everything into the main per-cluster loop:

```go
for _, cluster := range clusters {
    if getProductClaim(&cluster) == "" {
        klog.V(4).Infof("Cluster %s missing product claim, skipping", cluster.Name)
        continue
    }

    result, err := r.ensureOperatorInstalled(ctx, mesh, &cluster)
    if err != nil {
        return reconcile.Result{}, fmt.Errorf("failed to ensure mesh operator on cluster %s: %w", cluster.Name, err)
    } else if result.RequeueAfter > 0 {
        return result, nil
    }

    if mesh.Spec.Security.Trust.CertManager.IssuerRef.Name != "" {
        if err := r.ensureCertificateForCluster(ctx, mesh, &cluster); err != nil {
            return reconcile.Result{}, fmt.Errorf("failed to ensure certificate for cluster %s: %w", cluster.Name, err)
        }
        if err := r.ensureCacertsManifestWork(ctx, mesh, &cluster); err != nil {
            return reconcile.Result{}, fmt.Errorf("failed to ensure cacerts ManifestWork for cluster %s: %w", cluster.Name, err)
        }
    }
}
```

Then delete `ensureCertificatesCreated` and `ensureCacertsDistributed` — they're just wrappers around the per-cluster functions.

mkolesnik:
Just a note: this won't be needed if we proceed with #115

sridhargaddam:
> Just a note: this won't be needed if we proceed with #115

Since its still a draft and we have not yet finalized on the approach, let me update the current PR with your suggestion.

pkg/hub/mesh/controller.go
```diff
@@ -259,29 +259,27 @@ func (r *Reconciler) operatorConfigConflicts(a, b meshv1alpha1.OperatorConfig) b
...
 
-	// Create certificates for each cluster if cert-manager is configured
-	if mesh.Spec.Security.Trust.CertManager.IssuerRef.Name != "" {
-		if err := r.ensureCertificatesCreated(ctx, mesh, clusters); err != nil {
-			return reconcile.Result{}, err
-		}
-
-		if err := r.ensureCacertsDistributed(ctx, mesh, clusters); err != nil {
-			return reconcile.Result{}, err
+		if mesh.Spec.Security.Trust.CertManager.IssuerRef.Name != "" {
```
mkolesnik, line 289: Now that cert/cacerts creation is guarded by the product claim check, its worth adding an integration test for this.

sridhargaddam: Done

test/integration/controller_test.go
```diff
@@ -601,6 +601,27 @@ var _ = Describe("MultiClusterMesh Controller", func() {
...
+		When("cluster has no product claim and issuer is configured", func() {
+			BeforeEach(func() {
+				util.CreateManagedCluster(ctx, k8sClient, clusterName, testClusterSet)
+				util.CreateMultiClusterMesh(ctx, k8sClient, meshName, testNs, testClusterSet, util.CertManagerSpec("mesh-issuer"))
+			})
+
+			It("should not create Certificate or cacerts ManifestWork", func() {
+				expectMeshNotReady(meshName, testNs)
+				expectNoCertificate(testNs, clusterName)
+				expectNoCacertsManifestWork(clusterName)
```
mkolesnik: I don't thikn we need to check this, the MW is expected to be created as a response to a secret anyway, it's not expected at this point (and even if it was its not something for this test IMHO)
```suggestion
```

Or. if you want to check it then I'd check the MW isnt created when the claim is missing and a secret is created, but this workflow is highly unlikely (it means that user created a mesh with the issuer first, and then removed it once certificate was created but right before the cert-manager had time to create the secret)

mkolesnik (commented):
Small comment otherwise LGTM

test/integration/controller_test.go
sridhargaddam: Agree let me update it.

mkolesnik (approved):

sridhargaddam:
@jewertow I had to modify an existing integration test as we cannot have a namespace greater than 63 chars. PTAL, thanks.

sridhargaddam:
@mkolesnik I had to rebase this PR, can you reapprove? TIA

test/integration/controller_test.go
```diff
@@ -813,6 +833,17 @@ func expectCacertsManifestWork(clusterNamespace string) *workv1.ManifestWork {
...
 
+func expectNoCertificate(namespace, clusterName string) {
+	Consistently(func() bool {
+		cert := &certmanagerv1.Certificate{}
+		err := k8sClient.Get(ctx, types.NamespacedName{
+			Name:      fmt.Sprintf("cacerts-%s", clusterName),
+			Namespace: namespace,
+		}, cert)
+		return errors.IsNotFound(err)
+	}).Should(BeTrue())
```
mkolesnik: List by mesh labels and assert empty instead of getting by constructed name:

```suggestion
	Consistently(func() []certmanagerv1.Certificate {
		certList := &certmanagerv1.CertificateList{}
		Expect(k8sClient.List(ctx, certList,
			client.InNamespace(namespace),
			client.MatchingLabels{meshcontroller.MeshNameLabel: meshName},
		)).To(Succeed())
		return certList.Items
	}).Should(BeEmpty())
```

mkolesnik (commented):
Small suggestion to make test more robust

test/integration/controller_test.go
sridhargaddam: Thanks @mkolesnik. The test intent here is to assert "no certificate was created for this specific cluster" (because it lacks a product claim), not "no certificates exist for the mesh." Filtering by MeshNameLabel alone would pass today since each test has one cluster, but it would give a false positive if we ever added a test with two clusters in the same mesh where one has a product claim and the other doesn't. Checking by the exact certificate name matches how the controller creates them and keeps the assertion precise to what we're actually testing.

mkolesnik: You're wrong.
For integration tests this should cover as much as possible, and the code as-is leave a huge blind spot.
If you add some such test as you're worried about in the future you can test it differently, this one needs to be correct as it.

sridhargaddam: > You're wrong.

Which specific statement in my previous comment is incorrect? Can you elaborate?

> For integration tests this should cover as much as possible, and the code as-is leave a huge blind spot.

I'm not sure where the blind spot is. The test case I added explicitly sets the product claim (using util.SetProductClaim) for a specific cluster and then validates that certificates are created for that managedCluster only after the product claim is added.

mkolesnik: You're wrong since the test is not robust.
The "Expect no certificate" is not robust it'll pass if the certificate gets created by a different name for example.
It's better in this case to just check no certificates at all are created for that mesh since that's the real expectation, and not based on the name which is fragile.

test/integration/controller_test.go
```diff
@@ -821,12 +854,20 @@ func expectNoCacertsManifestWork(clusterNamespace string) {
...
+	Eventually(func() int {
+		Expect(k8sClient.List(ctx, certList,
+			client.InNamespace(namespace),
+			client.MatchingLabels{
+				meshcontroller.MeshNameLabel:    meshName,
+				meshcontroller.ClusterNameLabel: clusterName,
+			},
+		)).To(Succeed())
+		return len(certList.Items)
+	}).Should(Equal(1), "expected exactly one Certificate for cluster %s", clusterName)
```
mkolesnik: nit: Gomega convention prefers asserting on the collection directly rather than on its length.
This gives better failure output (the actual items are printed).

```suggestion
	Eventually(func() []certmanagerv1.Certificate {
		Expect(k8sClient.List(ctx, certList,
			client.InNamespace(namespace),
			client.MatchingLabels{
				meshcontroller.MeshNameLabel:    meshName,
				meshcontroller.ClusterNameLabel: clusterName,
			},
		)).To(Succeed())
		return certList.Items
	}).Should(HaveLen(1), "expected exactly one Certificate for cluster %s", clusterName)
```

mkolesnik (approved):

Issue #101, opened 2026-06-02
Decide mutability of spec.security.trust.certManager.issuerRef


mkolesnik:
If the user changes `issuerRef.name` after creation, existing Certificate resources still point to the old issuer.
The controller doesn't update them because `ensureCertificateForCluster` returns early if the Certificate exists.

Need to decide:
- Should `issuerRef` be immutable (like `spec.clusterSet`, see #97)?
- Should the controller detect the change and update existing Certificates?
- How does this interact with the designed Root CA rotation path (admin updates the Issuer's secret, not the mesh's issuerRef)?

PR #108, opened 2026-06-02
Cleanup certificates on cluster set change

### LINKED ISSUE #96: Clean up hub-side Certificate when cluster is removed from ClusterSet
When a cluster is removed from a ClusterSet (label changed or cluster deleted), the controller cleans up the operator and cacerts ManifestWorks in the cluster namespace.
However, the hub-side `Certificate` resource (e.g., `cacerts-cluster-east` in the mesh namespace) and its corresponding Secret persist.

cert-manager keeps renewing these certificates for clusters that are no longer part of the mesh.
The controller should delete the Certificate (and let cert-manager clean up the Secret) when the cluster is no longer in the ClusterSet.

The cleanup should happen in the same reconciliation path that handles ManifestWork cleanup (`cleanupManifestWorks` or a similar function scoped to the mesh namespace).

Related: #67 (Certificate ownership on mesh deletion)

jewertow:
Fixes #96

### DIFF
--- a/pkg/hub/mesh/controller.go
+++ b/pkg/hub/mesh/controller.go
@@ -280,6 +280,10 @@ func (r *Reconciler) doReconcile(ctx context.Context, mesh *meshv1alpha1.MultiCl
 		if err := r.ensureCacertsDistributed(ctx, mesh, clusters); err != nil {
 			return reconcile.Result{}, err
 		}
+
+		if err := r.cleanupCertificates(ctx, mesh, clusters); err != nil {
+			return reconcile.Result{}, fmt.Errorf("failed to cleanup Certificates: %w", err)
+		}
 	}
 
 	if err := r.cleanupManifestWorks(ctx, mesh.Spec.ClusterSet); err != nil {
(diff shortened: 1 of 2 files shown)


pkg/hub/mesh/controller.go
```diff
@@ -280,6 +280,10 @@ func (r *Reconciler) doReconcile(ctx context.Context, mesh *meshv1alpha1.MultiCl
 		if err := r.ensureCacertsDistributed(ctx, mesh, clusters); err != nil {
 			return reconcile.Result{}, err
 		}
+
+		if err := r.cleanupCertificates(ctx, mesh, clusters); err != nil {
```
mkolesnik: If the user removes `issuerRef.name`, this entire block is skipped and existing Certificates are never cleaned up (orphaned until the mesh itself is deleted). 
Please consider moving `cleanupCertificates` outside the `if` guard so it runs unconditionally, or note it as a known limitation for #101.

jewertow: I expected this question, but I forgot to leave a comment explaining this decision. In general, it depends on whether we allow BYO CA. Since we do not allow to skip installation of the operator, I assumed we won't support BYO CA for now. Then handling the above case is not needed.

However, I am totally for supporting skipping installation of the operator and CA distribution.

mkolesnik: I don't think this is a BYO CA question. The `security.trust.certManager` fields are all `+optional`, so a user clearing them is a valid operation by API contract. If they enable cert-manager and later remove the config, the Certificates get orphaned.

We can discuss if BYO CA is valid but that's besides the point of this PR I feel

mkolesnik: Reported #112 for further discussion

jewertow: > The security.trust.certManager fields are all +optional

Fair point. Done

mkolesnik (approved):

Issue #89, opened 2026-05-28
Report actual operator installation status from spokes via ManifestWork feedbackRules


mkolesnik:
Currently, the controller creates the operator `ManifestWork` and immediately reports `OperatorInstalled: False / Reason: ManifestWorkCreated`.
It never transitions to `True` because there's no mechanism to observe the actual installation state on the spoke.

We should add `feedbackRules` to the operator `ManifestWork` to report the `ClusterServiceVersion` phase back to the hub.
The controller can then use this feedback to set `OperatorInstalled: True` when the CSV reaches `Succeeded`, or report a degraded state if it stalls.

This also unblocks the mesh-level `Ready` condition, which currently can never be `True`.

PR #114, opened 2026-06-04
Add e2e test suite for operator lifecycle on Kind dev-env

mkolesnik:
Tests the full hub-to-spoke round-trip via real OCM work agents:
- Mesh creation deploys operator on both spokes
- Config changes propagate to spoke Subscriptions
- Mesh deletion cleans up ManifestWorks and spoke resources

Adds `make test-e2e` and `make dev-clean-meshes` targets.

Depends-On: #113

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -267,7 +267,7 @@ deploy-addon: $(KIND) gen images $(KUSTOMIZE) ## Build and deploy addon to the h
 	@echo "==> Addon controller deployed successfully. Use KUBECONFIG=$(HUB_KUBECONFIG) to interact with the hub."
 
 .PHONY: dev-clean-meshes
-dev-clean-meshes: ## Delete all MultiClusterMesh CRs and addon ManifestWorks from dev-env
+dev-clean-meshes: ## Delete all addon resources from the dev-env whether user created, or "leaked" resources left by an aborted/failed test
 	kubectl --kubeconfig=$(HUB_KUBECONFIG) delete multiclustermeshes -A --all --ignore-not-found=true
 	kubectl --kubeconfig=$(HUB_KUBECONFIG) delete manifestwork -A -l app.kubernetes.io/managed-by=multicluster-mesh-addon --ignore-not-found=true
 
--- a/Makefile
+++ b/Makefile
@@ -185,6 +185,7 @@ KIND_VERSION ?= v0.31.0
 CLUSTERADM_VERSION ?= v1.3.1
 K8S_VERSION ?= v1.35.0
 OLM_VERSION ?= v0.43.0
+CERT_MANAGER_VERSION ?= v1.20.2
 
 OS := $(shell go env GOOS)
 ARCH := $(shell go env GOARCH)
@@ -219,11 +220,11 @@ $(CLUSTERADM): | $(BIN_DIR)
 
 DEV_ENV_SCRIPT := $(CURDIR)/hack/dev-env.sh
 
-export DEV_KUBE_DIR K8S_VERSION OLM_VERSION
+export DEV_KUBE_DIR K8S_VERSION OLM_VERSION CERT_MANAGER_VERSION
 export KIND CLUSTERADM
 
 .PHONY: dev-env
-dev-env: create-clusters install-olm init-ocm join-clusters deploy-addon ## Provision full dev environment (Kind + OCM + addon)
+dev-env: create-clusters install-olm install-cert-manager init-ocm join-clusters deploy-addon ## Provision full dev environment (Kind + OCM + addon)
 
 .PHONY: create-clusters
 create-clusters: $(KIND) ## Create 3 Kind clusters (hub, cluster1, cluster2)
@@ -232,6 +233,10 @@ create-clusters: $(KIND) ## Create 3 Kind clusters (hub, cluster1, cluster2)
 .PHONY: install-olm
 install-olm: ## Install OLM on managed clusters (cluster1, cluster2)
 	$(DEV_ENV_SCRIPT) install-olm
+
+.PHONY: install-cert-manager
+install-cert-manager: ## Install cert-manager on the hub cluster
+	$(DEV_ENV_SCRIPT) install-cert-manager
 
 .PHONY: init-ocm
 init-ocm: $(CLUSTERADM) ## Initialize hub as OCM control plane
--- a/Makefile
+++ b/Makefile
@@ -140,7 +140,12 @@ test-integration: $(ENVTEST) gen-crds deps update-test-crds ## Run integration t
 	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) -p path)" \
 	go run github.com/onsi/ginkgo/v2/ginkgo -v --tags=integration ./test/integration/...
 
-# TODO: Add test-e2e target for end-to-end tests against real cluster
+.PHONY: test-e2e
+test-e2e: ## Run e2e tests against dev-env clusters (requires make dev-env)
+	HUB_KUBECONFIG=$(HUB_KUBECONFIG) \
+	CLUSTER1_KUBECONFIG=$(DEV_KUBE_DIR)/cluster1.config \
+	CLUSTER2_KUBECONFIG=$(DEV_KUBE_DIR)/cluster2.config \
+	go run github.com/onsi/ginkgo/v2/ginkgo -v --fail-fast --tags=e2e ./test/e2e/...
 
 .PHONY: build
 build: $(BIN_DIR) ## Build addon binary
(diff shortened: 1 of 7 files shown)


mkolesnik:
/hold
Depends on #113

mkolesnik:
Merged upstream/main (includes #113), no conflicts.

/unhold

Makefile
```diff
@@ -261,6 +266,11 @@ deploy-addon: $(KIND) gen images $(KUSTOMIZE) ## Build and deploy addon to the h
 		-n multicluster-mesh-system --timeout=180s
 	@echo "==> Addon controller deployed successfully. Use KUBECONFIG=$(HUB_KUBECONFIG) to interact with the hub."
 
+.PHONY: dev-clean-meshes
+dev-clean-meshes: ## Delete all MultiClusterMesh CRs and addon ManifestWorks from dev-env
+	kubectl --kubeconfig=$(HUB_KUBECONFIG) delete multiclustermeshes -A --all --ignore-not-found=true
+	kubectl --kubeconfig=$(HUB_KUBECONFIG) delete manifestwork -A -l app.kubernetes.io/managed-by=multicluster-mesh-addon --ignore-not-found=true
```
FilipB, line 272: Shouldn't be all the manifestwork resources removed automatically when multiclustermeshes is removed?

test/e2e/mesh_lifecycle_test.go
```diff
@@ -0,0 +1,155 @@
...
+)
+
+var clusters = []string{"cluster1", "cluster2"}
+
+var _ = Describe("Controller health", func() {
+	It("should have the controller deployment available", func(ctx SpecContext) {
+		deploy := &appsv1.Deployment{}
+		err := hubClient.Get(ctx, types.NamespacedName{Name: controllerName, Namespace: controllerNamespace}, deploy)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(deploy.Status.AvailableReplicas).To(Not(BeZero()))
```
FilipB: Would it be safer to check for Ready status of the deployment or this is equivalent?

test/e2e/mesh_lifecycle_test.go
```diff
@@ -0,0 +1,155 @@
...
+		}
+
+		Step("Verifying Subscriptions are removed from spoke clusters")
+		for _, spokeClient := range spokeClients {
+			util.ExpectResourceDeleted(ctx, spokeClient, &operatorsv1alpha1.Subscription{}, meshcontroller.OperatorNameSail, meshcontroller.DefaultOperatorNs)
+		}
+	})
+})
+
+func refreshMesh(ctx context.Context, mesh *meshv1alpha1.MultiClusterMesh) error {
```
FilipB: A bit confusing name as it's not refreshing the mesh? Maybe `fetchLatestMesh`?

test/e2e/mesh_lifecycle_test.go
```diff
@@ -0,0 +1,155 @@
...
+	BeforeEach(func(ctx SpecContext) {
+		Step("Creating test resources")
+		ns = util.UniqueName("test-ns")
+		util.CreateNamespace(ctx, hubClient, ns)
+		mesh = util.CreateMultiClusterMesh(ctx, hubClient, util.UniqueName("test-mesh"), ns, "mesh-cluster-set")
+
+		Step("Waiting for controller to reconcile")
+		Eventually(func(g Gomega) {
+			g.Expect(refreshMesh(ctx, mesh)).To(Succeed())
+			g.Expect(meta.FindStatusCondition(mesh.Status.Conditions, meshv1alpha1.ConditionReady)).NotTo(BeNil())
```
FilipB, line 71: `NotTo(BeNil()` Should it expect `True`?

test/e2e/mesh_lifecycle_test.go
```diff
@@ -0,0 +1,155 @@
...
+			Expect(err).NotTo(HaveOccurred())
+
+			Step("Verifying OperatorGroup exists on %s", cluster)
+			ogList := &operatorsv1.OperatorGroupList{}
+			err = spokeClient.List(ctx, ogList, client.InNamespace(meshcontroller.DefaultOperatorNs))
+			Expect(err).NotTo(HaveOccurred())
+			Expect(ogList.Items).To(HaveLen(1))
+
+			Step("Verifying Subscription exists on %s", cluster)
+			Expect(getSubscription(ctx, spokeClient)).NotTo(BeNil())
```
FilipB: Do we want to also verify content? Or even if the operator is installed correctly?

test/e2e/mesh_lifecycle_test.go
```diff
@@ -0,0 +1,155 @@
...
+	})
+})
+
+var _ = Describe("MultiClusterMesh lifecycle", func() {
+	var (
+		mesh *meshv1alpha1.MultiClusterMesh
+		ns   string
+	)
+
+	BeforeEach(func(ctx SpecContext) {
```
FilipB: We can use the same namespace for all 3 tests can't we? No reason to re-create it 3 times.

test/e2e/mesh_lifecycle_test.go
FilipB: This will pass if error is returned, we should check both error and nil.

FilipB (changes requested):

test/e2e/mesh_lifecycle_test.go
mkolesnik: No because the mesh never becomes true (this is open as issue #89 and the PR #119 is still open).
We can update it later once the issue is fixed.
I'll add a TODO

test/e2e/mesh_lifecycle_test.go
mkolesnik: Gomega's multi-return semantics handle the error: `Expect(val, err)` fails if err is non-nil before evaluating the matcher.

mkolesnik: I'll add some basic assertions on the content just as a sanity, since this comes from the MW.
Full propagation mesh -> MW is covered by the integ tests, covering it here would just be testing that OCM works as expected (which isn't the point of the test - basically you would be testing that MW on hub created the spec you asked for on the spoke).

test/e2e/mesh_lifecycle_test.go
mkolesnik: Will rename to `getMesh`.

Makefile
mkolesnik: They are, and we even have tests for that.

This is just in case the tests didn't clean up properly (either they didn't do it for whatever reason, or the process was killed).
This is actually useful in general, we could expand it in the future just as a "clean the stuff" action when developing on the dev-env, but right now it was useful when writing the tests.

test/e2e/mesh_lifecycle_test.go
mkolesnik: `AvailableReplicas > 0` and the `Available` condition being True mean the same thing for a single-replica deployment. But I don't mind changing it..

test/e2e/mesh_lifecycle_test.go
mkolesnik: Will move namespace creation to `BeforeAll`/`AfterAll`, keep mesh creation per-test in `BeforeEach`.

mkolesnik (commented):
Thanks @FilipB I addressed your review comments

Makefile
FilipB: Please add a comment here so nobody else is wondering why it's needed.

test/e2e/mesh_lifecycle_test.go
FilipB: Checking if the operator is installed correctly would also catch things like typo in the e.g. channel name. Is there any test which would catch typo in channel name?

test/e2e/mesh_lifecycle_test.go
FilipB: I'm not sure why you need to create/delete mesh before/after each test. All 3 tests can use the same mesh resource if run in order.

FilipB: Sorry, bad idea to try to run tests in order. Please ignore.

mkolesnik: I'm not sure what you mean?

For BeforeAll/AfterAll Ginkgo requires we use Ordered specs, so did you mean this?

Makefile
mkolesnik: I'll update the help text instead I think it'll be clearer

test/e2e/mesh_lifecycle_test.go
FilipB: I was thinking about running the tests in order so we could create the mesh just once, but it's bad idea. Please ignore.

test/e2e/mesh_lifecycle_test.go
FilipB: @mkolesnik WDYT? Is this valid question?

mkolesnik: I think we discussed it on the call?
The tests right now validate the addon works as it should - in general.
Adding validation for further parts would add considerable slowness, and as such should be put in separate "slow tests".

So this is more like "Smoke E2E tests" that check the subscription actually got created as requested.

I think such slow tests can even go further and install Istio and see that it works, but we'd need the other parts for that.
Either way I think such tests should be separate from these more "basic" ones, and we can (and should) do them in a separate PR.

WDYT?

FilipB: Sounds good.

FilipB (approved):
Thank you

PR #115, opened 2026-06-04
Simplify operator config by defaulting to OSSM

mkolesnik:
Removes platform branching from the controller.
Kubebuilder defaults now set OSSM values at admission time, so the controller reads `spec.operator` directly.
Users can override all fields to use a different operator (e.g. Sail on Kind for dev/testing).

Depends-On: #114

### DIFF
--- a/pkg/apis/mesh/v1alpha1/types.go
+++ b/pkg/apis/mesh/v1alpha1/types.go
@@ -82,7 +82,7 @@ type MultiClusterMeshSpec struct {
 	// +optional
 	ControlPlane ControlPlaneConfig `json:"controlPlane,omitempty"`
 
-	// Operator defines the Sail Operator installation configuration
+	// Operator defines the service mesh operator installation configuration
 	// +optional
 	Operator OperatorConfig `json:"operator,omitempty"`
 
@@ -99,11 +99,17 @@ type ControlPlaneConfig struct {
 	Namespace string `json:"namespace,omitempty"`
 }
 
-// OperatorConfig defines the Sail Operator installation settings
+// OperatorConfig defines the service mesh operator installation settings.
+// Defaults target OSSM on OpenShift. Override fields to use a different operator variant (e.g. Sail).
 type OperatorConfig struct {
-	// Namespace is the namespace where the Sail Operator will be installed
-	// Defaults to "openshift-operators" on OpenShift, "sail-operator" on vanilla Kubernetes
+	// Name is the OLM package name of the operator
 	// +optional
+	// +kubebuilder:default="servicemeshoperator3"
+	Name string `json:"name,omitempty"`
+
+	// Namespace is the namespace where the operator will be installed
+	// +optional
+	// +kubebuilder:default="openshift-operators"
 	Namespace string `json:"namespace,omitempty"`
 
 	// Channel is the OLM subscription channel (e.g., "stable", "1.23")
@@ -112,13 +118,13 @@ type OperatorConfig struct {
 	Channel string `json:"channel,omitempty"`
 
 	// Source is the CatalogSource name
-	// Defaults to "redhat-operators" on OpenShift, "operatorhubio-catalog" on vanilla Kubernetes
 	// +optional
+	// +kubebuilder:default="redhat-operators"
 	Source string `json:"source,omitempty"`
 
 	// SourceNamespace is the namespace of the CatalogSource
-	// Defaults to "openshift-marketplace" on OpenShift, "olm" on vanilla Kubernetes
 	// +optional
+	// +kubebuilder:default="openshift-marketplace"
 	SourceNamespace string `json:"sourceNamespace,omitempty"`
 
 	// StartingCSV is the specific operator version to install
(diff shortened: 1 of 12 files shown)


mkolesnik:
/hold
Depends on #114

mkolesnik:
@coderabbit review

sridhargaddam:
Thanks @mkolesnik, for putting this together. I ran this PR through an AI code review tool (Claude) and it flagged a few observations, mostly around how this would play out in mixed-platform environments and how other addons in the org handle similar scenarios. Sharing them below for discussion.

### Mixed ClusterSets

The current code auto-detects platform per-cluster, so a ClusterSet with both OCP and vanilla K8s clusters gets OSSM on one and Sail on the other - automatically. With the CRD-level defaults, all clusters in a set would get `servicemeshoperator3` from `redhat-operators`, which wouldn't exist on a Kind/EKS/GKE cluster.

Since `spec.operator` is mesh-scoped (not cluster-scoped), there's no way for a user to override per-cluster within a single mesh CR. How are you thinking about handling that case? Is the assumption that mixed ClusterSets are out of scope for the controller?

### How other stolostron addons handle this

The tool scanned other addons in the org for product claim usage and found it's a **pretty established pattern**:

- **submariner-addon** : reads the claim to decide whether to include SCC manifests (closest architectural sibling to us, same ManifestWork pattern)
- **search-v2-operator** : gates Prometheus scraping to OCP clusters
- **multicloud-integrations** : gates GitOps addon installation to OCP
- **console** : drives distribution display, managed-OpenShift badges, upgrade paths
- **clusterlifecycle-state-metrics**, **multicluster-global-hub**, **multicloud-operators-foundation** : metrics and inventory labeling

Since the claim is already populated by the klusterlet for free, it seems like we'd be leaving useful data on the table by not reading it. Curious if you considered keeping the detection internal to the controller while still simplifying the CRD surface?

### Diagnostic signal

The tool also flagged the loss of the `ReasonMissingProductClaim` per-cluster condition. Right now it tells operators that a cluster hasn't finished registration. Without it, a cluster that hasn't registered its claims would just get OSSM defaults pushed and fail with a potentially confusing OLM error. Any thoughts on an alternative signal there?

### Future platform-specific needs

Even if today we only use the claim for operator selection, there's a decent chance we'll need it later for things like SCC handling or PodSecurity admission differences. Removing the infrastructure now and re-adding it later would be a larger change than keeping it.

### The current code already supports overrides

The existing method `applyOperatorDefaults` already gives us the best of both worlds - every field is guarded by `if field == ""`, so user-specified values are used as-is and auto-detection only fills in the gaps. The only piece that doesn't have a user override path is the package name (`servicemeshoperator3` vs `sailoperator`), which is hardcoded off the `isOCP` flag. That could be fixed by adding a `Name` field to the CRD (which this PR actually does) without removing the product claim detection.

Curious to hear your thoughts on these observations. Happy to discuss further.

mkolesnik:
Thanks for running this through a review tool and sharing.

On mixed ClusterSets: it's not just that nobody asked for it, I'd argue mixed operator flavors is actually a flawed design.
Sail stable and OSSM stable can track different Istio versions.
We have no good way to guarantee a viable working mesh when control planes run different operator builds.
So this isn't a limitation we're accepting, it's the right behavior.
Picking one operator flavor and sticking to it is what we want.

On the established pattern: those addons use claims for different things (SCC manifests, Prometheus gating, GitOps).
If we need platform-specific behavior later (SCC, PodSecurity), we can add that independently.
It's not related to which operator flavor gets installed.

On diagnostic signal: the `MissingProductClaim` condition was just an unnecessary gate.
It didn't add real signal.
The operator can be created (or not) based on cluster readiness.
If OSSM defaults are pushed to a cluster without the right catalog, the ManifestWork feedback reports the actual OLM error, which is more actionable.

On future needs: same point as above.
Mixed operator flavors is not something we can reliably support.

Either way we need to pick a default, and since this targets OCP, OSSM is the right choice.
If we ever need platform detection for other concerns, that's a separate change.

sridhargaddam:
@mkolesnik, as discussed on the call, lets go ahead with the PR. Please update the design doc as well and address the merge conflicts.

pkg/apis/mesh/v1alpha1/types.go
```diff
@@ -112,13 +118,13 @@ type OperatorConfig struct {
 	Channel string `json:"channel,omitempty"`
 
 	// Source is the CatalogSource name
-	// Defaults to "redhat-operators" on OpenShift, "operatorhubio-catalog" on vanilla Kubernetes
 	// +optional
+	// +kubebuilder:default="redhat-operators"
```
sridhargaddam, line 122: Please update the `samples/basic.yaml` file as its used for KIND deployments as part of `make setup-mesh` script.

mkolesnik: Good catch, will update. The e2e test already overrides these for KIND (`mesh_lifecycle_test.go:72-80`), so the sample should match.

While at it I will also update the other samples since they still reference per-platform defaults that this PR removed.

sridhargaddam (approved):
LGTM, thank you.

PR #116, opened 2026-06-05
Set ObservedGeneration on status conditions

### LINKED ISSUE #99: Set ObservedGeneration on status conditions
Status conditions don't set `ObservedGeneration`.
Consumers (other controllers, UI, CLI) can't tell whether a condition reflects the current spec or a previous one.

Standard k8s convention is to set `ObservedGeneration` to `.metadata.generation` when writing conditions, so consumers can compare it against the resource's current generation.

sridhargaddam:
Standard k8s convention is to set ObservedGeneration to .metadata.generation when writing conditions, so consumers can compare it against the resource's current generation. This PR implements the necessary changes.

Fixes: https://github.com/stolostron/multicluster-mesh-addon/issues/99

### DIFF
--- a/test/integration/controller_test.go
+++ b/test/integration/controller_test.go
@@ -102,6 +102,7 @@ var _ = Describe("MultiClusterMesh Controller", func() {
 			expectMeshNotReady(meshName, testNs)
 			expectClusterOperatorConditionReason(meshName, testNs, cluster1, meshv1alpha1.ReasonManifestWorkCreated)
 			expectClusterOperatorConditionReason(meshName, testNs, cluster2, meshv1alpha1.ReasonManifestWorkCreated)
+			expectConditionsObservedGeneration(meshName, testNs)
 		})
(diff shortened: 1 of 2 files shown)


test/integration/controller_test.go
```diff
@@ -603,6 +603,62 @@ var _ = Describe("MultiClusterMesh Controller", func() {
 		})
 	})
 
+	Context("ObservedGeneration", func() {
```
mkolesnik: ObservedGeneration is a standard property of every condition, not a separate feature.
Instead of adding a dedicated Context, you should add `expectConditionsObservedGeneration` calls to the existing tests where it makes sense to do so (e.g. basic reconcile, conflict, missing product claim, spec update).

sridhargaddam: Good point, Mike.

mkolesnik (approved):

PR #117, opened 2026-06-07
Extract key package for object key construction

mkolesnik:
Inline types.NamespacedName struct literals add verbosity,
especially in multi-line form. A small key package with By()
and For() makes client call sites more concise.

### DIFF
--- a/pkg/hub/mesh/controller.go
+++ b/pkg/hub/mesh/controller.go
@@ -19,7 +19,6 @@ import (
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/types"
 	applyconfigv1 "k8s.io/client-go/applyconfigurations/meta/v1"
 	"k8s.io/klog/v2"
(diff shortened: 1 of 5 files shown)


pkg/hub/mesh/controller.go
```diff
@@ -830,10 +820,7 @@ func (r *Reconciler) ensureCacertsDistributed(ctx context.Context, mesh *meshv1a
 func (r *Reconciler) ensureCacertsManifestWork(ctx context.Context, mesh *meshv1alpha1.MultiClusterMesh, cluster *clusterv1.ManagedCluster) error {
 	secretName := getCacertsName(cluster.Name)
 	secret := &corev1.Secret{}
-	err := r.Get(ctx, types.NamespacedName{
-		Name:      secretName,
-		Namespace: mesh.Namespace,
-	}, secret)
+	err := r.Get(ctx, key.By(secretName, mesh.Namespace), secret)
```
sridhargaddam: When we read key.By(secretName, ns), it sounds like we're querying by a secretName, not building a key from one. A better alternative could be key.Of(secretName, ns), instead of key.By(...). WDYS?

mkolesnik: Sure sounds good, I'll make the change!

sridhargaddam (approved):

PR #119, opened 2026-06-08
Report operator installation status via ManifestWork feedback

### LINKED ISSUE #89: Report actual operator installation status from spokes via ManifestWork feedbackRules

mkolesnik:
The mesh Ready condition could never become True because the controller never confirmed operator installation on spoke clusters.

Add a feedback rule to the operator ManifestWork that pulls back the Subscription's installedCSV field.
When the work agent reports a non-empty value, the controller sets OperatorInstalled=True for that cluster.
Once all clusters confirm, the mesh becomes Ready.

Closes #89

### DIFF
--- a/pkg/hub/mesh/controller.go
+++ b/pkg/hub/mesh/controller.go
@@ -498,7 +498,7 @@ func (r *Reconciler) determineStatus(ctx context.Context, mesh *meshv1alpha1.Mul
 			return fmt.Errorf("failed to get operator ManifestWork for cluster %s: %w", cluster.Name, err)
 		}
 
-		if installedCSV := getManifestWorkFeedback(operatorWork, FeedbackInstalledCSV); installedCSV != nil {
+		if installedCSV := getManifestWorkFeedback(operatorWork); installedCSV != nil {
 			meta.SetStatusCondition(&status.Conditions, metav1.Condition{
 				Type:               meshv1alpha1.ConditionOperatorInstalled,
 				Status:             metav1.ConditionTrue,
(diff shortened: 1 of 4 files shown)


pkg/hub/mesh/controller.go
```diff
@@ -528,6 +537,17 @@ func (r *Reconciler) determineStatus(ctx context.Context, mesh *meshv1alpha1.Mul
...
+func getManifestWorkFeedback(work *workv1.ManifestWork, name string) *string {
+	for _, manifest := range work.Status.ResourceStatus.Manifests {
+		for _, value := range manifest.StatusFeedbacks.Values {
+			if value.Name == name {
+				return value.Value.String
+			}
+		}
+	}
+	return nil
+}
```
jewertow: The `name` parameter is an unnecessary generality:
1. Single caller, single value - the function is only ever called with `FeedbackInstalledCSV`.
2. The ManifestWork only configures one feedback rule.
3. If a second feedback value is needed later, the function can be generalized at that point.

The function could be simplified to this:
```suggestion
func getManifestWorkFeedback(work *workv1.ManifestWork) *string {
	for _, manifest := range work.Status.ResourceStatus.Manifests {
		for _, value := range manifest.StatusFeedbacks.Values {
			if value.Name == "installedCSV" {
				return value.Value.String
			}
		}
	}
	return nil
}
```

mkolesnik: There was a plan to reuse this later for other feedback, but I'll simplify for now and change later when needed

jewertow (approved):

PR #122, opened 2026-06-10
Adding initial CodeRabbit configuration

FilipB:
Trying to reduce verbosity

### DIFF
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -1,0 +1,18 @@
+# yaml-language-server: $schema=https://coderabbit.ai/integrations/schema.v2.json
+
+# org level configuration: https://github.com/stolostron/coderabbit/blob/main/.coderabbit.yaml
+inheritance: true
+tone_instructions: Prefer concise responses (high information density, low fluff)
+reviews:
+  high_level_summary: false
+  review_status: false
+  collapse_walkthrough: true
+  path_instructions:
+    - path: '**'
+      instructions: >
+        - Focus on major issues impacting performance, readability, maintainability and security. Avoid nitpicks and avoid verbosity.
+        - Group similar issues into a single comment instead of posting multiple notes.
+        - Skip repetition — if a pattern repeats, mention it once at a summary level only.
+        - Do not post comments on code style, formatting, or non-critical improvements.
+chat:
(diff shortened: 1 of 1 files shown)


.coderabbit.yaml
```diff
@@ -0,0 +1,18 @@
...
+  collapse_walkthrough: true
+  path_instructions:
+    - path: '**'
+      instructions: >
+        - Focus on major issues impacting performance, readability, maintainability and security. Avoid nitpicks and avoid verbosity.
+        - Group similar issues into a single comment instead of posting multiple notes.
+        - Skip repetition — if a pattern repeats, mention it once at a summary level only.
+        - Do not post comments on code style, formatting, or non-critical improvements.
+chat:
+  auto_reply: false
```
sridhargaddam, line 19: The org default for `auto_reply` is true. Setting this flag to false disables follow-up questions to CodeRabbit in PR comment threads, such as "explain this further" or "suggest a fix". This might be useful in some scenarios. WDYS?

.coderabbit.yaml
```diff
@@ -0,0 +1,18 @@
+# yaml-language-server: $schema=https://coderabbit.ai/integrations/schema.v2.json
+
+# org level configuration: https://github.com/stolostron/coderabbit/blob/main/.coderabbit.yaml
+inheritance: true
+tone_instructions: Prefer concise responses (high information density, low fluff)
+reviews:
+  high_level_summary: false
+  review_status: false
+  collapse_walkthrough: true
```
sridhargaddam, line 10: We can also consider the following setting to reduce verbosity.

changed_files_summary: false

sridhargaddam (commented):
Thanks @FilipB for looking into it.

.coderabbit.yaml
FilipB: I'm not sure, we need to try. From doc: `Let CodeRabbit reply automatically without requiring a mention/tag.` So my understanding is `false` means it will still work, you just need to use `@coderabbitai`

.coderabbit.yaml
FilipB: Added, thanks.

sridhargaddam (approved):

PR #123, opened 2026-06-10
Fix update-test-crds failure when vendor directory is present

sridhargaddam:
When a `vendor` directory exists, Go defaults to `-mod=vendor` which causes `go list -m -f '{{.Dir}}'` to return empty output since vendored modules don't have a cache directory path. This made the Makefile's update-test-crds target fail with "not found in go.mod".

This PR modifies the command to use `-mod=mod` to always resolve module paths from the module cache. This works correctly with or without a vendor directory.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,5 +20,8 @@ bin/
 config/crd
 test/integration/crds/
 
+# Go vendor directory
+vendor/
+
 # Dev environment (make dev-env)
 .kube/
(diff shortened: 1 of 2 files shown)


mkolesnik (commented):
looks good, should we also add `vendor` to gitignore so that noone checks it in by mistake?

mkolesnik (approved):

PR #125, opened 2026-06-10
Add frontend skeleton

nrfox:
Add the initial frontend implementation.

### DIFF
--- a/frontend/README.md
+++ b/frontend/README.md
@@ -1,0 +1,156 @@
+# ossm-acm-ui
+
+An OpenShift `ConsolePlugin` that registers a standalone **Fleet Service Mesh** perspective in the OpenShift Console perspective switcher, providing fleet-wide service mesh visibility across ACM-managed clusters.
+
+## What it does
+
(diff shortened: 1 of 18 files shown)


frontend/README.md
```diff
@@ -0,0 +1,156 @@
...
+oc apply -f deploy/consoleplugin.yaml
+```
+
+### 6. Enable the plugin in the Console operator
+
+```bash
+oc patch console.operator.openshift.io cluster \
+  --type=merge \
+  --patch='{"spec":{"plugins":["ossm-acm"]}}'
+```
```
jmazzitelli, line 104: This replaces the entire `plugins` list, wiping out any previously enabled plugins (e.g. `acm`, `mce`, `monitoring-plugin`). Use a JSON patch to append instead:

```suggestion
oc patch console.operator.openshift.io cluster \
  --type=json \
  --patch='[{"op":"add","path":"/spec/plugins/-","value":"ossm-acm"}]'
```

jmazzitelli (commented):
fix readme

jmazzitelli:
nick is on PTO for a while - I created a PR with this content and will build it out some more before merging. See https://github.com/stolostron/multicluster-mesh-addon/pull/132

I think we can close this PR as obsolete since that new PR has this PR's content with more

jmazzitelli (changes requested):
this PR should be closed in lieu of https://github.com/stolostron/multicluster-mesh-addon/pull/132

PR #127, opened 2026-06-15
Update DiscoveryConfig TokenValidity API

yxun:
This PR updates the DiscoveryConfig TokenValidity API. We will use this `TokenValidity`  value when creating a ManagedServiceAccount resource for its `spec.rotation.validity` field.

Because the type differences between those two fields and the allowed value range is limited, we need an update of this DiscoveryConfig TokenValidity API field.

This PR implements those changes by using kubebuilder validation. 
The first changes default value from 1 month to 360 hours (15 days). 
The second validates input value's format which should be allowed as the `metav1.Duration` format. 
The third checks the value is no less than 10 minutes. When I manually create a `ManagedServiceAccount` CR, its webhook shows the validity value should be no less than 10 minutes. Otherwise, the creation of a `ManagedServiceAccount` will fail.

This PR is the second part of the original last PR: 


pkg/apis/mesh/v1alpha1/types.go
```diff
@@ -123,10 +123,12 @@ type IssuerReference struct {
 // DiscoveryConfig defines endpoint discovery token configuration
 type DiscoveryConfig struct {
 	// TokenValidity defines how long discovery tokens are valid
-	// Supports hours (h), days (d), weeks (w), or months (m)
+	// Supports hours (h), minutes (m), seconds (s) or milliseconds (ms). If unset, defaults to 360h
 	// +optional
-	// +kubebuilder:default="1m"
-	TokenValidity string `json:"tokenValidity,omitempty"`
+	// +kubebuilder:default="360h"
+	// +kubebuilder:validation:XValidation:rule="matches(self, '^([0-9]{1,5}(h|m|s|ms)){1,4}$')",message="invalid duration value"
```
mkolesnik: The regex has a minor alternation bug (`m` matches before `ms`) and is extra maintenance. You can validate the format with `duration()` instead, it throws on invalid input so this rejects garbage with a clear message:

```suggestion
	// +kubebuilder:validation:XValidation:rule="duration(self) >= duration('0s')",message="invalid duration value"
```

Or just drop this and rely on the "minimum 10m" validation, which would also enforce the formatting..

pkg/apis/mesh/v1alpha1/types.go
```diff
@@ -123,10 +123,11 @@ type IssuerReference struct {
 // DiscoveryConfig defines endpoint discovery token configuration
 type DiscoveryConfig struct {
 	// TokenValidity defines how long discovery tokens are valid
-	// Supports hours (h), days (d), weeks (w), or months (m)
+	// Supports hours (h), minutes (m), seconds (s) or milliseconds (ms). If unset, defaults to 360h
```
sridhargaddam: Do we need a milliseconds granularity here? If not, probably the lowest can be seconds.

yxun: The type of this field is `metav1.Duration` . It supports those time units. I dropped a regular expression from Mike's comment above. 
I can add it back and update it as 

`// +kubebuilder:validation:XValidation:rule="matches(self, '^([0-9]{1,5}(h|m|s)){1,3}$')",message="invalid  value: TokenValidity supports hours (h), minutes (m), seconds (s)"`

This will check and tell user using hours, minutes or seconds as time unit.

pkg/apis/mesh/v1alpha1/types.go
```diff
@@ -123,10 +123,12 @@ type IssuerReference struct {
...
 	// TokenValidity defines how long discovery tokens are valid
-	// Supports hours (h), days (d), weeks (w), or months (m)
+	// Supports hours (h), minutes (m), seconds (s). If unset, defaults to 360h
 	// +optional
-	// +kubebuilder:default="1m"
-	TokenValidity string `json:"tokenValidity,omitempty"`
+	// +kubebuilder:default="360h"
+	// +kubebuilder:validation:XValidation:rule="matches(self, '^([0-9]{1,5}(h|m|s)){1,3}$')",message="invalid value: TokenValidity supports hours (h), minutes (m), seconds (s)"
+	// +kubebuilder:validation:XValidation:rule="duration(self) >= duration('10m')", message="TokenValidity must be at least 10 minutes"
+	TokenValidity *metav1.Duration `json:"tokenValidity,omitempty"`
```
sridhargaddam: Can you please update the design.md file as well https://github.com/stolostron/multicluster-mesh-addon/blob/bf31f52db1f6b4b5ec494ca1c2c0cb9d8acbd76e/docs/design.md?plain=1#L136

pkg/apis/mesh/v1alpha1/types.go
```diff
@@ -123,10 +123,12 @@ type IssuerReference struct {
 // DiscoveryConfig defines endpoint discovery token configuration
 type DiscoveryConfig struct {
 	// TokenValidity defines how long discovery tokens are valid
-	// Supports hours (h), days (d), weeks (w), or months (m)
+	// Supports hours (h), minutes (m), seconds (s). If unset, defaults to 360h
 	// +optional
-	// +kubebuilder:default="1m"
-	TokenValidity string `json:"tokenValidity,omitempty"`
+	// +kubebuilder:default="360h"
+	// +kubebuilder:validation:XValidation:rule="matches(self, '^([0-9]{1,5}(h|m|s)){1,3}$')",message="invalid value: TokenValidity supports hours (h), minutes (m), seconds (s)"
```
sridhargaddam: I think you can remove this regex validation as the validation in the next line is sufficient (i.e., duration() throws on invalid input, which Kubernetes treats as a validation failure).

pkg/apis/mesh/v1alpha1/types.go
sridhargaddam: I think its better to use it as a non-pointer here. 

```suggestion
	TokenValidity metav1.Duration `json:"tokenValidity,omitempty"`
```

sridhargaddam (commented):
LGTM, thanks @yxun

jewertow (approved):

docs/design.md
```diff
@@ -133,7 +133,7 @@ The MVP supports the [Multi-Primary Multi-Network] mesh topology. This aligns wi
 | `spec.operator.startingCSV` | No | Pin to a specific operator version |
 | `spec.operator.installPlanApproval` | No | `Automatic` or `Manual` (default: `Automatic`) |
 | `spec.security.trust.certManager.issuerRef.name` | No | cert-manager Issuer name for Root CA |
-| `spec.security.discovery.tokenValidity` | No | ManagedServiceAccount token lifetime (default: `1m`) |
+| `spec.security.discovery.tokenValidity` | No | ManagedServiceAccount token lifetime (default: `360h`, minimum value: `10m`) |
 
 ### Example
```
mkolesnik, line 139: The example YAML at line 160 still uses `tokenValidity: "1w"`, but `metav1.Duration` doesn't support weeks. `time.ParseDuration` only accepts `h`, `m`, `s` (and sub-second units).

mkolesnik (commented):
Small fix to design doc needed

docs/design.md
yxun: I just did a search of all files which mention the `tokenValidity`. I pushed changes for those samples and docs in the last commit. Thanks for catching that.

pkg/apis/mesh/v1alpha1/types.go
yxun: I reverted this pointer type change. 
I found it's bug there[1] from our test run logs. The `// +kubebuilder:default="360h"` never sets the default value in that Go Zero-Value Pitfall situation. 

I will explain this in next comment when my local e2e test run complete. 

[1] https://pkg.go.dev/open-cluster-management.io/managed-serviceaccount/apis/authentication/v1beta1#ManagedServiceAccountRotation

yxun:
`make dev-env && make test-e2e` passed in my local run for this PR latest change.

The previous test job failure[1] was caused by changing the pointer type to non-pointer type 

```
TokenValidity metav1.Duration `json:"tokenValidity,omitempty"`
```

Because Go implicitly assigns "zero values" (e.g., false for booleans, 0s for metav1.Duration) to unset fields. 
We have defined a `//+kubebuilder:default` marker, Kubernetes cannot tell the difference between a user explicitly providing a zero value and the Go zero value initializing an empty field. 
When generating the CRD schemas, Kubernetes sees a zero value for the `TokenValidity` rather than a missing field (nil), meaning it checks the validation rule against 0s instead of treating it as an empty field that needs defaulting.

* The better solution is using pointer type for optional fields according to the api-conventions doc[2].

I see the ManageServiceAccount pkg and their CRD doc is in [3] and now I think there is a bug there as well, That default may never set its value there.

cc @sridhargaddam @jewertow , wdyt

[1] https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/stolostron_multicluster-mesh-addon/127/pull-ci-stolostron-multicluster-mesh-addon-main-integration/2067207785396834304

[2] https://github.com/kubernetes/community/blob/main/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required

[3] https://pkg.go.dev/open-cluster-management.io/managed-serviceaccount/apis/authentication/v1beta1#ManagedServiceAccountRotation

pkg/apis/mesh/v1alpha1/types.go
yxun: My defense about this pointer type field is the comment below:

https://github.com/stolostron/multicluster-mesh-addon/pull/127#issuecomment-4730547170

mkolesnik: I understood that to be the case as well, otherwise you can't send empty field and expect k8s to fill it correctly, but maybe I'm missing something

sridhargaddam: You are right @yxun, thanks.

sridhargaddam (approved):

Issue #120, opened 2026-06-08
Update existing ManagedServiceAccount resources when the user changes the `tokenValidity` on the mesh


yxun:
Currently, if the `ManagedServiceAccount` exists in a managed cluster namespace, we skip following updates from the mesh controller reconciliation. 
If the user changes the `tokenValidity` field value on the mesh CR spec, we should update the existing `ManagedServiceAccount` resource(s)' spec value instead of using the old value.

yxun:
~This has been implemented in #130~

yxun:
This has been implemented as #189

PR #130, opened 2026-06-15
Implement ManagedServiceAccount resources reconciliation

yxun:
This PR implements the creation and cleanup of the `ManagedServiceAccount` resources for each managed cluster in the `ClusterSet`.

The goal of this PR is getting API access secret by creating a `ManagedServiceAccount` custom resource for each managed cluster. The distribution of those API access secrets or tokens will be implemented in the next PR.

References:
- https://docs.redhat.com/en/documentation/red_hat_advanced_cluster_management_for_kubernetes/2.16/html/clusters/cluster_mce_overview#managed-serviceaccount-addon

This PR is the third part of the original large PR:

The controller reconciliation need an update method for existing MSA. I created an issue in https://github.com/stolostron/multicluster-mesh-addon/issues/120. That TODO item will be added in follow up PRs.
e2e test suite need new tests. I created an issue in https://github.com/stolostron/multicluster-mesh-addon/issues/133 . That TODO item will be added in follow up PRs.


pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -0,0 +1,140 @@
...
+func (r *Reconciler) createManagedServiceAccounts(ctx context.Context, mesh *meshv1alpha1.MultiClusterMesh, clusters []clusterv1.ManagedCluster) error {
+	if len(clusters) == 0 {
+		klog.V(4).Info("The ClusterSet has no managed cluster")
+		return nil
+	}
+
+	klog.Info("Creating ManagedServiceAccount resources for each managed cluster in the ClusterSet")
+	for _, cluster := range clusters {
+		existing := &msav1beta1.ManagedServiceAccount{}
+		msaName := fmt.Sprintf("%s-istio-reader", mesh.Name)
```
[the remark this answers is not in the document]
yxun: > Use a namespace-qualified identity (or hashed (namespace,name) key) and include MeshNamespaceLabel: mesh.Namespace in MSA list selectors.

There is a concern about using (namespace, name) in https://github.com/stolostron/multicluster-mesh-addon/issues/72 

The Label has been included in the ` meshOwnedLabels`  function.

test/integration/controller_test.go
```diff
@@ -629,6 +643,88 @@ var _ = Describe("MultiClusterMesh Controller", func() {
...
+			})
+
+			It("should process a ManagedServiceAccount", func() {
+				msa := &msav1beta1.ManagedServiceAccount{}
+				Eventually(func() error {
+					return k8sClient.Get(ctx, types.NamespacedName{
+						Name: fmt.Sprintf("%s-istio-reader", meshName), Namespace: cluster3,
+					}, msa)
+				}).Should(Succeed())
+			})
```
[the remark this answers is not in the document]
yxun: This has been set in `test/integration/suite_test.go`

pkg/hub/mesh/controller.go
```diff
@@ -298,6 +298,12 @@ func (r *Reconciler) doReconcile(ctx context.Context, mesh *meshv1alpha1.MultiCl
 		}
 	}
 
+	// Create ManagedServiceAccount resources for each cluster
+	if err := r.createManagedServiceAccounts(ctx, mesh, clusters); err != nil {
```
mkolesnik: Missing RBAC marker for ManagedServiceAccount. The controller creates, gets, lists, and deletes MSAs but `make gen-rbac` won't generate the permission without a marker. Add:

```
//+kubebuilder:rbac:groups=authentication.open-cluster-management.io,resources=managedserviceaccounts,verbs=get;list;watch;create;delete
```

next to the other RBAC markers in controller.go.

pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -0,0 +1,150 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	multiClusterSecretLabel  = "istio/multiCluster"
+	clusterNameAnnotationKey = "networking.istio.io/cluster"
+	msaRootWord              = "istio-reader"
+)
+
+var validity metav1.Duration
```
mkolesnik: Why is this a package level variable?

pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -0,0 +1,150 @@
...
+	return nil
+}
+
+// cleanupManagedServiceAccounts deletes ManagedServiceAccount on clusters that are removed from the given ClusterSet.
+func (r *Reconciler) cleanupManagedServiceAccounts(ctx context.Context, mesh *meshv1alpha1.MultiClusterMesh, clusters []clusterv1.ManagedCluster) error {
+	clusterNames := clusterNameSet(clusters)
+
+	msaList := &msav1beta1.ManagedServiceAccountList{}
+	if err := r.List(ctx, msaList,
+		client.MatchingLabels{ManagedByLabel: ManagedByValue, MeshNameLabel: mesh.Name}); err != nil {
```
mkolesnik: The list selector uses `MeshNameLabel` but not `MeshNamespaceLabel`.
Two meshes with the same name in different namespaces would cross-delete each other's MSAs during cleanup.

MSAs live in cluster namespaces (cross-namespace), so the label selector is the only scoping mechanism.
Compare with `cleanupCertificates` which already includes `MeshNamespaceLabel` in its selector.

Fix here and in `deleteAllManagedServiceAccounts` (L122):

```suggestion
		client.MatchingLabels{MeshNameLabel: mesh.Name, MeshNamespaceLabel: mesh.Namespace}); err != nil {
```

pkg/apis/mesh/v1alpha1/types.go
mkolesnik, line 1: Isn't this PR #127?
You should mark dependent PRs with "Depends on: #127" and comment /hold

pkg/hub/mesh/controller.go
```diff
@@ -298,6 +298,12 @@ func (r *Reconciler) doReconcile(ctx context.Context, mesh *meshv1alpha1.MultiCl
...
+		klog.Errorf("Failed to create ManagedServiceAccount resources: %v", err)
+		return reconcile.Result{}, err
+	}
```
mkolesnik: This logs the error and returns it (handle or propagate, not both).
Also returns unwrapped, unlike other errors in `doReconcile`.

This should match the existing pattern:

```suggestion
	if err := r.createManagedServiceAccounts(ctx, mesh, clusters); err != nil {
		return reconcile.Result{}, fmt.Errorf("failed to create ManagedServiceAccounts: %w", err)
	}
```

pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+
+	for _, msa := range msaList.Items {
+		clusterName := msa.Labels[ClusterNameLabel]
+		if clusterNames[clusterName] {
+			continue
+		}
+
+		klog.Infof("Deleting ManagedServiceAccount %s/%s (cluster %s no longer in ClusterSet %s)", msa.Namespace, msa.Name, clusterName, mesh.Spec.ClusterSet)
+		if err := r.Delete(ctx, &msa); err != nil {
```
mkolesnik: Use `client.IgnoreNotFound` for Delete calls (the controller-runtime convention).
The MSA could be deleted between the List and Delete, and NotFound is a no-op here.

Same applies to L111, L134, L142.

```suggestion
		if err := client.IgnoreNotFound(r.Delete(ctx, &msa)); err != nil {
```

pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -0,0 +1,150 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/klog/v2"
+	clusterv1 "open-cluster-management.io/api/cluster/v1"
+	msav1beta1 "open-cluster-management.io/managed-serviceaccount/apis/authentication/v1beta1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	multiClusterSecretLabel  = "istio/multiCluster"
+	clusterNameAnnotationKey = "networking.istio.io/cluster"
```
mkolesnik: This constant isn't referenced anywhere in this file.
If it's for the secret distribution PR, add it there.

pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -0,0 +1,150 @@
...
+func (r *Reconciler) createManagedServiceAccounts(ctx context.Context, mesh *meshv1alpha1.MultiClusterMesh, clusters []clusterv1.ManagedCluster) error {
+	if len(clusters) == 0 {
+		klog.V(4).Info("The ClusterSet has no managed cluster")
+		return nil
+	}
+
+	klog.Info("Creating ManagedServiceAccount resources for each managed cluster in the ClusterSet")
+	for _, cluster := range clusters {
+		existing := &msav1beta1.ManagedServiceAccount{}
+		msaName := fmt.Sprintf("%s-%s-%s", mesh.Namespace, msaRootWord, mesh.Name)
```
mkolesnik: nit: This can be set before the loop

pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -0,0 +1,150 @@
...
+			continue
+		} else if !errors.IsNotFound(err) {
+			return fmt.Errorf("failed to get ManagedServiceAccount %s/%s: %w", cluster.Name, msaName, err)
+		}
+
+		if mesh.Spec.Security.Discovery.TokenValidity == nil {
+			validity = metav1.Duration{Duration: 360 * time.Hour}
+		} else {
+			validity = *mesh.Spec.Security.Discovery.TokenValidity
+		}
```
mkolesnik: nit: `TokenValidity` nil-check doesn't depend on `cluster`.
You can compute it before the loop.

test/integration/controller_test.go
```diff
@@ -629,6 +640,97 @@ var _ = Describe("MultiClusterMesh Controller", func() {
...
+
+		BeforeEach(func() {
+			cluster1 = util.UniqueName("cluster1")
+			cluster2 = util.UniqueName("cluster2")
+			util.CreateK8sManagedCluster(ctx, k8sClient, cluster1, testClusterSet)
+			util.CreateK8sManagedCluster(ctx, k8sClient, cluster2, testClusterSet)
+			util.CreateMultiClusterMesh(ctx, k8sClient, meshName, testNs, testClusterSet, meshv1alpha1.MultiClusterMeshSpec{})
+		})
+
+		It("should create ManagedServiceAccount resources with default validity for each ManagedCluster in the ClusterSet", func() {
```
mkolesnik: Please add a test that sets custom validity and verifies the MSA gets that value

test/integration/controller_test.go
```diff
@@ -629,6 +640,97 @@ var _ = Describe("MultiClusterMesh Controller", func() {
...
+			msa1 := expectManagedServiceAccount(fmt.Sprintf("%s-%s-%s", testNs, msaRootWord, meshName), cluster1)
+			msa2 := expectManagedServiceAccount(fmt.Sprintf("%s-%s-%s", testNs, msaRootWord, meshName), cluster2)
+
+			Expect(msa1.Labels[meshcontroller.ManagedByLabel]).To(Equal(meshcontroller.ManagedByValue))
```
mkolesnik: This only checks `ManagedByLabel`.
The cleanup functions filter by `MeshNameLabel` (and should filter by `MeshNamespaceLabel`), so assert those too.

Also please verify `ClusterNameLabel` matches the expected cluster.

pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -0,0 +1,150 @@
...
+var validity metav1.Duration
+
+// createManagedServiceAccounts creates ManagedServiceAccount resources for each cluster in the ClusterSet
+func (r *Reconciler) createManagedServiceAccounts(ctx context.Context, mesh *meshv1alpha1.MultiClusterMesh, clusters []clusterv1.ManagedCluster) error {
+	if len(clusters) == 0 {
+		klog.V(4).Info("The ClusterSet has no managed cluster")
+		return nil
+	}
+
+	klog.Info("Creating ManagedServiceAccount resources for each managed cluster in the ClusterSet")
```
mkolesnik: nit: this fires on every reconcile even when all MSAs already exist.
Should be `klog.V(4).Info` or removed.

pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -0,0 +1,150 @@
...
+
+	secretList := &corev1.SecretList{}
+	if err := r.List(ctx, secretList, client.InNamespace(mesh.Namespace), client.MatchingLabels{
+		multiClusterSecretLabel: "true", ManagedByLabel: ManagedByValue, MeshNameLabel: mesh.Name,
+	}); err != nil && !errors.IsNotFound(err) {
+		return fmt.Errorf("failed to list istio-remote secrets managed by mesh %s: %w", mesh.Name, err)
+	}
+
+	for _, msa := range msaList.Items {
+		if err := r.Delete(ctx, &msa); err != nil && !errors.IsNotFound(err) {
```
mkolesnik: nit: the `else` logs "Deleting" after the delete already happened, and also fires for the `IsNotFound` case.
`} else {` after a `return` is also redundant.

The pattern in `cleanupManagedServiceAccounts` (log before delete) is cleaner.

```suggestion
		klog.Infof("Deleting ManagedServiceAccount %s/%s", msa.Namespace, msa.Name)
		if err := client.IgnoreNotFound(r.Delete(ctx, &msa)); err != nil {
			return fmt.Errorf("failed to delete ManagedServiceAccount %s/%s: %w", msa.Namespace, msa.Name, err)
		}
```

pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -0,0 +1,150 @@
...
+	msaList := &msav1beta1.ManagedServiceAccountList{}
+	if err := r.List(ctx, msaList,
+		client.MatchingLabels{ManagedByLabel: ManagedByValue, MeshNameLabel: mesh.Name}); err != nil {
+		return fmt.Errorf("failed to list ManagedServiceAccounts: %w", err)
+	}
+
+	secretList := &corev1.SecretList{}
+	if err := r.List(ctx, secretList, client.InNamespace(mesh.Namespace), client.MatchingLabels{
+		multiClusterSecretLabel: "true", ManagedByLabel: ManagedByValue, MeshNameLabel: mesh.Name,
+	}); err != nil && !errors.IsNotFound(err) {
```
mkolesnik: nit: `client.List` returns an empty list when no resources match, never `NotFound`.
The `!errors.IsNotFound(err)` check is dead code.
Same on L122 and L129.

pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -0,0 +1,148 @@
...
+		if err := client.IgnoreNotFound(r.Delete(ctx, &msa)); err != nil {
+			return fmt.Errorf("failed to delete ManagedServiceAccount %s/%s: %w", msa.Namespace, msa.Name, err)
+		}
+	}
+
+	for _, sec := range secretList.Items {
+		if err := r.Delete(ctx, &sec); err != nil && !errors.IsNotFound(err) {
+			return fmt.Errorf("failed to delete an istio remote secret %s: %w", sec.Name, err)
+		} else {
+			klog.Infof("Deleting an istio remote secret %s", sec.Name)
```
sridhargaddam: By the time we are here, its already deleted. Please move it to Line140.

Also, since there is a possiblity for IsNotFound, can you wrap it as above (i.e., IgnoreNotFound in Line 134)

chart/templates/clusterrole.yaml
```diff
@@ -14,6 +14,16 @@ rules:
   - get
   - list
   - watch
+- apiGroups:
+  - authentication.open-cluster-management.io
+  resources:
+  - managedserviceaccounts
+  verbs:
+  - create
```
sridhargaddam, line 22: Do we need `update` as well?

pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -0,0 +1,148 @@
...
+		msa := &msav1beta1.ManagedServiceAccount{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      msaName,
+				Namespace: cluster.Name,
+				Labels:    meshOwnedLabels(mesh, cluster.Name),
+			},
+			Spec: msav1beta1.ManagedServiceAccountSpec{
+				Rotation: msav1beta1.ManagedServiceAccountRotation{
+					Enabled:  true,     // the ServiceAccount token will be rotated before it expires
+					Validity: validity, // the duration of validity for requesting the signed ServiceAccount token
```
jewertow: I don't think these comments are needed :) 
```suggestion
					Enabled:  true,
					Validity: validity,
```

jewertow (commented):
LGTM, but I assume you will address multi cluster secret creation in a follow-up PR, right?

yxun:
> LGTM, but I assume you will address multi cluster secret creation in a follow-up PR, right?

@jewertow , yes, that was in the original PR and I will have some updates and address those in the next smaller PR.
The following steps will be as follows:

MSA CR creation, update
--> The MSA controller (which is not a concern from multicluster-mesh-addon side) creates a SA and an API server access Secret with access token 
--> multicluster-mesh-addon controller builds an Istio remote secret using the API server access Secret token and adds required labels.
--> The controller creates ManifestWork for distributing Istio remote secrets.

sridhargaddam (commented):
@yxun, please take a look at the e2e test failures. Otherwise, LGTM.

pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+
+	for _, sec := range secretList.Items {
+		clusterName := sec.Labels[ClusterNameLabel]
+		if clusterNames[clusterName] {
+			continue
+		}
+
+		klog.Infof("Deleting istio remote secret %s/%s (cluster %s no longer in ClusterSet %s)", sec.Namespace, sec.Name, clusterName, mesh.Spec.ClusterSet)
+		if err := client.IgnoreNotFound(r.Delete(ctx, &sec)); err != nil {
```
mkolesnik: The cleanup functions delete Secrets (lines 113, 144), but the RBAC marker at controller.go:162 only grants `get;list;watch` on Secrets. Add `delete` (and `create` if the follow-up PR will create secrets here) to the Secrets RBAC marker:

```suggestion
//+kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch;create;delete
```

Then run `make gen-rbac` to regenerate the clusterrole.

pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -0,0 +1,150 @@
...
+	if mesh.Spec.Security.Discovery.TokenValidity == nil {
+		validity = metav1.Duration{Duration: 360 * time.Hour}
+	} else {
+		validity = *mesh.Spec.Security.Discovery.TokenValidity
+	}
+
+	for _, cluster := range clusters {
+		existing := &msav1beta1.ManagedServiceAccount{}
+		if err := r.Get(ctx, types.NamespacedName{Name: msaName, Namespace: cluster.Name}, existing); err == nil {
+			klog.Infof("Cluster %s has an existing ManagedServiceAccount resource %s, skipping createManagedServiceAccount", cluster.Name, msaName)
```
mkolesnik: nit: This fires on every reconcile when MSAs exist. If you happen to respin this, otherwise you can leave as is:

```suggestion
			klog.V(4).Infof("Cluster %s has an existing ManagedServiceAccount resource %s, skipping createManagedServiceAccount", cluster.Name, msaName)
```

mkolesnik (changes requested):

yxun:
The fix of e2e test error is in https://github.com/stolostron/multicluster-mesh-addon/pull/157

It follows the following doc:
https://github.com/open-cluster-management-io/managed-serviceaccount#installation-steps

hack/dev-env.sh
```diff
@@ -253,54 +261,83 @@ join_clusters() {
...
     log "OCM topology ready"
-    on "${HUB}" kubectl get managedclusters
-    on "${HUB}" kubectl get managedclustersets
+    kubectl --kubeconfig="${hub_kubeconfig}" get managedclusters
+    kubectl --kubeconfig="${hub_kubeconfig}" get managedclustersets
+}
+
+install_managed_serviceaccount() {
+    local hub_kubeconfig
+    hub_kubeconfig="$(echo "${DEV_KUBE_DIR}/${HUB}.config")"
```
yxun: Reason of this change:
There was a failure in the job of the previous commit:
https://github.com/stolostron/multicluster-mesh-addon/actions/runs/28129345197/job/83303594782

```
/home/runner/work/multicluster-mesh-addon/multicluster-mesh-addon/hack/dev-env.sh: line 273: kubeconfig_for: command not found
``` 

I could never reproduce this error from my local run. Still not clear why the github workflow job was not able to find that function name.

mkolesnik: Please see #152

pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -0,0 +1,150 @@
...
+// If there is an existing ManagedServiceAccount, it skips creation.
+func (r *Reconciler) createManagedServiceAccounts(ctx context.Context, mesh *meshv1alpha1.MultiClusterMesh, clusters []clusterv1.ManagedCluster) error {
+	if len(clusters) == 0 {
+		klog.V(4).Info("The ClusterSet has no managed cluster")
+		return nil
+	}
+
+	msaName := fmt.Sprintf("%s-%s-%s", mesh.Namespace, msaRootWord, mesh.Name)
+
+	var validity metav1.Duration
```
mkolesnik: The kubebuilder default (`+kubebuilder:default="360h"` on `TokenValidity`) guarantees this field is always populated for objects created through the API.

This nil check and the hardcoded `360 * time.Hour` are dead code with a hidden duplicate default.
Just use it directly.

pkg/hub/mesh/controller.go
```diff
@@ -388,6 +399,11 @@ func (r *Reconciler) handleDeletion(ctx context.Context, mesh *meshv1alpha1.Mult
 		return reconcile.Result{}, fmt.Errorf("failed to cleanup ManifestWorks: %w", err)
 	}
 
+	klog.Infof("Handling deletion for ManagedServiceAccount resources managed by mesh %s/%s", mesh.Namespace, mesh.Name)
```
mkolesnik: No need to log this, you already log each deleted MSA

```suggestion
```

test/integration/controller_test.go
```diff
@@ -647,6 +658,124 @@ var _ = Describe("MultiClusterMesh Controller", func() {
...
+				otherMesh = util.UniqueName("other-mesh")
+				util.CreateNamespace(ctx, k8sClient, otherNs)
+				util.CreateMultiClusterMesh(ctx, k8sClient, otherMesh, otherNs, testClusterSet)
+			})
+
+			It("should keep the ManagedServiceAccount when one mesh is deleted", func() {
+				expectMeshNotReady(otherMesh, otherNs)
+				util.DeleteResource(ctx, k8sClient, &meshv1alpha1.MultiClusterMesh{}, otherMesh, otherNs)
+				msa := &msav1beta1.ManagedServiceAccount{}
+				Expect(k8sClient.Get(ctx, types.NamespacedName{Name: fmt.Sprintf("%s-%s-%s", testNs, msaRootWord, meshName), Namespace: cluster1}, msa)).Should(Succeed())
```
mkolesnik: This is a point-in-time check that could pass before cleanup runs.
Use `Consistently` to verify MSAs remain present over a polling window, similar to other tests.

pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -0,0 +1,150 @@
...
+)
+
+// createManagedServiceAccounts creates ManagedServiceAccount resources for each cluster.
+// It checks and uses the mesh.Spec.Security.Discovery.TokenValidity value.
+// If there is an existing ManagedServiceAccount, it skips creation.
+func (r *Reconciler) createManagedServiceAccounts(ctx context.Context, mesh *meshv1alpha1.MultiClusterMesh, clusters []clusterv1.ManagedCluster) error {
+	if len(clusters) == 0 {
+		klog.V(4).Info("The ClusterSet has no managed cluster")
+		return nil
+	}
```
mkolesnik: This check is inconsistent with the rest of the controller

pkg/hub/mesh/controller.go
```diff
@@ -305,6 +306,11 @@ func (r *Reconciler) doReconcile(ctx context.Context, mesh *meshv1alpha1.MultiCl
 		}
 	}
 
+	// Create ManagedServiceAccount resources for each cluster.
+	if err := r.createManagedServiceAccounts(ctx, mesh, clusters); err != nil {
```
mkolesnik: Since this needs to run for each cluster, it makes more sense to put this inside the main `for _, cluster := range clusters` loop on L283 and handle each cluster with `ensureEndpointDiscovery` or some similar terminology.

test/integration/controller_test.go
```diff
@@ -647,6 +658,124 @@ var _ = Describe("MultiClusterMesh Controller", func() {
...
+			var otherNs, otherMesh string
+
+			BeforeEach(func() {
+				otherNs = util.UniqueName("other-ns")
+				otherMesh = util.UniqueName("other-mesh")
+				util.CreateNamespace(ctx, k8sClient, otherNs)
+				util.CreateMultiClusterMesh(ctx, k8sClient, otherMesh, otherNs, testClusterSet)
+			})
+
+			It("should keep the ManagedServiceAccount when one mesh is deleted", func() {
```
mkolesnik: These two tests use addon-owned (shared resource) semantics, but MSAs are mesh-owned.
Each mesh creates its own MSAs, so there's no shared resource to "keep."

"should keep the ManagedServiceAccount when one mesh is deleted" (L755): deletes `otherMesh`, checks `meshName`'s MSAs survive, but never checks `otherMesh`'s MSAs are deleted.
This is a tautology for mesh-owned resources.

"should delete the ManagedServiceAccount when both meshes are deleted" (L763): passes even with broken scoping since everything is deleted.

Replace both with one test that verifies correct scoping: 
  delete one mesh, assert its MSAs are gone AND the other mesh's MSAs survive.


For example:

```go
It("should delete only the deleted mesh's MSAs", func() {
    expectMeshNotReady(otherMesh, otherNs)
    // Verify both meshes have MSAs
    expectManagedServiceAccount(fmt.Sprintf("%s-%s-%s", testNs, msaRootWord, meshName), cluster1)
    expectManagedServiceAccount(fmt.Sprintf("%s-%s-%s", otherNs, msaRootWord, otherMesh), cluster1)
    // Delete one mesh
    util.DeleteResource(ctx, k8sClient, &meshv1alpha1.MultiClusterMesh{}, otherMesh, otherNs)
    // Deleted mesh's MSAs are gone
    util.ExpectResourceDeleted(ctx, k8sClient, &msav1beta1.ManagedServiceAccount{},
        fmt.Sprintf("%s-%s-%s", otherNs, msaRootWord, otherMesh), cluster1)
    util.ExpectResourceDeleted(ctx, k8sClient, &msav1beta1.ManagedServiceAccount{},
        fmt.Sprintf("%s-%s-%s", otherNs, msaRootWord, otherMesh), cluster2)
    // Surviving mesh's MSAs remain
    Expect(k8sClient.Get(ctx, types.NamespacedName{Name: fmt.Sprintf("%s-%s-%s", testNs, msaRootWord, meshName), Namespace: cluster1}, &msav1beta1.ManagedServiceAccount{})).To(Succeed())
    Expect(k8sClient.Get(ctx, types.NamespacedName{Name: fmt.Sprintf("%s-%s-%s", testNs, msaRootWord, meshName), Namespace: cluster2}, &msav1beta1.ManagedServiceAccount{})).To(Succeed())
})
```

test/integration/controller_test.go
```diff
@@ -647,6 +658,124 @@ var _ = Describe("MultiClusterMesh Controller", func() {
...
+			cluster1 = util.UniqueName("cluster1")
+			cluster2 = util.UniqueName("cluster2")
+			util.CreateK8sManagedCluster(ctx, k8sClient, cluster1, testClusterSet)
+			util.CreateK8sManagedCluster(ctx, k8sClient, cluster2, testClusterSet)
+			util.CreateMultiClusterMesh(ctx, k8sClient, meshName, testNs, testClusterSet, meshv1alpha1.MultiClusterMeshSpec{})
+		})
+
+		It("should create ManagedServiceAccount resources with default validity for each ManagedCluster in the ClusterSet", func() {
+			msa1 := expectManagedServiceAccount(fmt.Sprintf("%s-%s-%s", testNs, msaRootWord, meshName), cluster1)
+			msa2 := expectManagedServiceAccount(fmt.Sprintf("%s-%s-%s", testNs, msaRootWord, meshName), cluster2)
```
mkolesnik: This is very repetitive, you can extract a simple func to format the MSA name:

```golang
	func expectedMSAName(meshNamespace, meshName string) string {
		return fmt.Sprintf("%s-istio-reader-%s", meshNamespace, meshName)
	}
```
Then you can use it inside `expectManagedServiceAccount` and in other places where you look for MSA

mkolesnik (commented):
Partial review, also please remove parts that were moved to #157

test/integration/controller_test.go
```diff
@@ -860,6 +998,33 @@ func expectCacertsSecret(work *workv1.ManifestWork) {
...
+	}).Should(Succeed())
+	return msa
+}
+
+// expectNoManagedServiceAccount makes sure that no ManagedServiceAccount is created for a cluster, checking consistently
+func expectNoManagedServiceAccount(meshNamespace, meshName, clusterName string) {
+	Consistently(func() bool {
+		msa := &msav1beta1.ManagedServiceAccount{}
+		err := k8sClient.Get(ctx, types.NamespacedName{
+			Name:      fmt.Sprintf("%s-%s-%s", meshNamespace, msaRootWord, meshName),
```
mkolesnik: This still manually formats the MSA name instead of using `expectedManagedServiceAccountName`. 

```suggestion
			Name:      expectedManagedServiceAccountName(meshNamespace, meshName),
```

pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -0,0 +1,134 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/klog/v2"
+	clusterv1 "open-cluster-management.io/api/cluster/v1"
+	msav1beta1 "open-cluster-management.io/managed-serviceaccount/apis/authentication/v1beta1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	multiClusterSecretLabel = "istio/multiCluster"
```
mkolesnik: The `multiClusterSecretLabel` constant and all the secret listing/deletion code in `cleanupManagedServiceAccounts` and `deleteAllManagedServiceAccounts` is dead code. Nothing in this PR or the codebase sets the `istio/multiCluster` label on any secret, so the query always returns an empty list.

Also note: the MSA token secret (created by the MSA controller) already has an ownerReference to the MSA and gets garbage collected when the MSA is deleted. Verified on dev-env.

Remove the secret cleanup code, the `multiClusterSecretLabel` constant, and the `corev1` import. If the follow-up PR creates separate istio-remote secrets, it should add both creation and cleanup together.

test/e2e/suite_test.go
```diff
@@ -47,6 +48,7 @@ var _ = BeforeSuite(func(ctx context.Context) {
 		workv1.Install,
 		operatorsv1.AddToScheme,
 		operatorsv1alpha1.AddToScheme,
+		msav1beta1.AddToScheme,
```
mkolesnik: Is there a need to register with the scheme if you're not adding any E2E tests here?

pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -0,0 +1,134 @@
...
+	msaRootWord             = "istio-reader"
```
mkolesnik: nit: Only used in one place please inline

pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -0,0 +1,134 @@
...
+)
+
+// ensureManagedServiceAccountCreated creates ManagedServiceAccount resources for a cluster.
+// It checks and uses the mesh.Spec.Security.Discovery.TokenValidity value.
+// If there is an existing ManagedServiceAccount, it skips creation.
+func (r *Reconciler) ensureManagedServiceAccountCreated(ctx context.Context, mesh *meshv1alpha1.MultiClusterMesh, cluster *clusterv1.ManagedCluster) error {
+	msaName := fmt.Sprintf("%s-%s-%s", mesh.Namespace, msaRootWord, mesh.Name)
+	existing := &msav1beta1.ManagedServiceAccount{}
+	if err := r.Get(ctx, types.NamespacedName{Name: msaName, Namespace: cluster.Name}, existing); err == nil {
```
mkolesnik: nit: The rest of the codebase uses `key.Of` for `NamespacedName` construction.
This file and the new test code use raw `types.NamespacedName` instead.

Production code:
- `endpoint_discovery.go:29`: `types.NamespacedName{Name: msaName, Namespace: cluster.Name}` -> `key.Of(msaName, cluster.Name)` (also drops the `types` import)

Test code:
- `controller_test.go:781`, `837`, `844` (inline Gets)
- `controller_test.go:1008` (`expectManagedServiceAccount`)
- `controller_test.go:1020` (`expectNoManagedServiceAccount`)

pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -0,0 +1,134 @@
...
+
+	msa := &msav1beta1.ManagedServiceAccount{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      msaName,
+			Namespace: cluster.Name,
+			Labels:    meshOwnedLabels(mesh, cluster.Name),
+		},
+		Spec: msav1beta1.ManagedServiceAccountSpec{
+			Rotation: msav1beta1.ManagedServiceAccountRotation{
+				Enabled:  true,
```
mkolesnik: Since this is deprecated we shouldn't use it
```suggestion
```

pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -0,0 +1,134 @@
...
+)
+
+// ensureManagedServiceAccountCreated creates ManagedServiceAccount resources for a cluster.
+// It checks and uses the mesh.Spec.Security.Discovery.TokenValidity value.
+// If there is an existing ManagedServiceAccount, it skips creation.
+func (r *Reconciler) ensureManagedServiceAccountCreated(ctx context.Context, mesh *meshv1alpha1.MultiClusterMesh, cluster *clusterv1.ManagedCluster) error {
+	msaName := fmt.Sprintf("%s-%s-%s", mesh.Namespace, msaRootWord, mesh.Name)
+	existing := &msav1beta1.ManagedServiceAccount{}
+	if err := r.Get(ctx, types.NamespacedName{Name: msaName, Namespace: cluster.Name}, existing); err == nil {
+		klog.V(4).Infof("Cluster %s has an existing ManagedServiceAccount resource %s, skipping ensureManagedServiceAccountCreated", cluster.Name, msaName)
```
mkolesnik: nit: The function name in the log is noise. Compare with controller.go:280 which just says "skipping".

```suggestion
		klog.V(4).Infof("Cluster %s already has ManagedServiceAccount %s, skipping", cluster.Name, msaName)
```

pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -0,0 +1,134 @@
...
+	msav1beta1 "open-cluster-management.io/managed-serviceaccount/apis/authentication/v1beta1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	multiClusterSecretLabel = "istio/multiCluster"
+	msaRootWord             = "istio-reader"
+)
+
+// ensureManagedServiceAccountCreated creates ManagedServiceAccount resources for a cluster.
```
mkolesnik: nit: The three-line comment restates what the function name and code already say. L60-61 also mentions "istio-remote secret" which this PR does not implement. Trim to one-line godoc at most, or remove entirely.

test/integration/controller_test.go
```diff
@@ -712,6 +723,133 @@ var _ = Describe("MultiClusterMesh Controller", func() {
...
+
+			Expect(msa1.Labels[meshcontroller.ManagedByLabel]).To(Equal(meshcontroller.ManagedByValue))
+			Expect(msa1.Labels[meshcontroller.MeshNameLabel]).To(Equal(secondMesh))
+			Expect(msa1.Labels[meshcontroller.MeshNamespaceLabel]).To(Equal(secondNs))
+			Expect(msa1.Labels[meshcontroller.ClusterNameLabel]).To(Equal(cluster1))
+			Expect(msa2.Spec.Rotation.Validity).To(Equal(metav1.Duration{Duration: 15 * time.Minute}))
+			Expect(msa2.Labels[meshcontroller.ClusterNameLabel]).To(Equal(cluster2))
+		})
+
+		When("add a cluster to the ClusterSet", func() {
```
mkolesnik: nit:

```suggestion
		When("a cluster is added to the ClusterSet", func() {
```

test/integration/controller_test.go
```diff
@@ -712,6 +723,133 @@ var _ = Describe("MultiClusterMesh Controller", func() {
...
+			BeforeEach(func() {
+				cluster3 = util.UniqueName("cluster3")
+				util.CreateK8sManagedCluster(ctx, k8sClient, cluster3, testClusterSet)
+			})
+
+			It("should process a ManagedServiceAccount", func() {
```
mkolesnik: nit:

```suggestion
			It("should create a ManagedServiceAccount for the new cluster", func() {
```

test/integration/controller_test.go
```diff
@@ -712,6 +723,133 @@ var _ = Describe("MultiClusterMesh Controller", func() {
...
+				msa := &msav1beta1.ManagedServiceAccount{}
+				Eventually(func() error {
+					return k8sClient.Get(ctx, types.NamespacedName{
+						Name: expectedManagedServiceAccountName(testNs, meshName), Namespace: cluster3,
+					}, msa)
+				}).Should(Succeed())
+			})
+		})
+
+		When("remove a cluster from the ClusterSet", func() {
```
mkolesnik: nit:

```suggestion
		When("a cluster is removed from the ClusterSet", func() {
```

test/integration/controller_test.go
```diff
@@ -712,6 +723,133 @@ var _ = Describe("MultiClusterMesh Controller", func() {
...
+				updateClusterSetLabel(cluster2, "")
+			})
+
+			It("should cleanup the ManagedServiceAccount", func() {
+				util.ExpectResourceDeleted(ctx, k8sClient, &msav1beta1.ManagedServiceAccount{},
+					expectedManagedServiceAccountName(testNs, meshName), cluster2)
+			})
+		})
+
+		When("create a cluster without clusterset label", func() {
```
mkolesnik: nit:

```suggestion
		When("a cluster is created without a ClusterSet label", func() {
```

test/integration/controller_test.go
```diff
@@ -712,6 +723,133 @@ var _ = Describe("MultiClusterMesh Controller", func() {
...
+			BeforeEach(func() {
+				cluster4 = util.UniqueName("cluster4")
+				util.CreateK8sManagedCluster(ctx, k8sClient, cluster4, "")
+			})
+
+			It("shouldn't process ManagedServiceAccount", func() {
```
mkolesnik: nit:

```suggestion
			It("shouldn't create a ManagedServiceAccount", func() {
```

pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -0,0 +1,134 @@
...
+	if err := r.Get(ctx, types.NamespacedName{Name: msaName, Namespace: cluster.Name}, existing); err == nil {
+		klog.V(4).Infof("Cluster %s has an existing ManagedServiceAccount resource %s, skipping ensureManagedServiceAccountCreated", cluster.Name, msaName)
+		return nil
+	} else if !errors.IsNotFound(err) {
+		return fmt.Errorf("failed to get ManagedServiceAccount %s/%s: %w", cluster.Name, msaName, err)
+	}
+
+	msa := &msav1beta1.ManagedServiceAccount{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      msaName,
```
mkolesnik, line 30: Technically this shouldn't cause a problem since name is up to 255 characters long, but it might cause issues in the MSA addon itself if too long.

I suggest you check it manually with a long name (i.e. long mesh ns and name) and if it causes problems open an issue for it to handle in the future.

pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -0,0 +1,134 @@
...
+
+	secretList := &corev1.SecretList{}
+	if err := r.List(ctx, secretList, client.InNamespace(mesh.Namespace), client.MatchingLabels{
+		multiClusterSecretLabel: "true", MeshNameLabel: mesh.Name, MeshNamespaceLabel: mesh.Namespace,
+	}); err != nil {
+		return fmt.Errorf("failed to list istio-remote secrets managed by mesh %s: %w", mesh.Name, err)
+	}
+
+	for _, msa := range msaList.Items {
+		clusterName := msa.Labels[ClusterNameLabel]
```
mkolesnik: Since you create it in the cluster ns you should use it here.

Also you can just drop the `clusterName` var and use the ns directly

```suggestion
		clusterName := msa.Namespace
```

pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -0,0 +1,134 @@
...
+		clusterName := sec.Labels[ClusterNameLabel]
+		if clusterNames[clusterName] {
+			continue
+		}
+
+		klog.Infof("Deleting istio remote secret %s/%s (cluster %s no longer in ClusterSet %s)", sec.Namespace, sec.Name, clusterName, mesh.Spec.ClusterSet)
+		if err := client.IgnoreNotFound(r.Delete(ctx, &sec)); err != nil {
+			return fmt.Errorf("failed to delete istio remote secret %s/%s: %w", sec.Namespace, sec.Name, err)
+		}
+	}
```
mkolesnik: This is dead code

```suggestion
```

pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -0,0 +1,134 @@
...
+		client.MatchingLabels{MeshNameLabel: mesh.Name, MeshNamespaceLabel: mesh.Namespace}); err != nil {
+		return fmt.Errorf("failed to list ManagedServiceAccounts: %w", err)
+	}
+
+	secretList := &corev1.SecretList{}
+	if err := r.List(ctx, secretList, client.InNamespace(mesh.Namespace), client.MatchingLabels{
+		multiClusterSecretLabel: "true", MeshNameLabel: mesh.Name, MeshNamespaceLabel: mesh.Namespace,
+	}); err != nil {
+		return fmt.Errorf("failed to list istio-remote secrets managed by mesh %s: %w", mesh.Name, err)
+	}
```
mkolesnik: This is dead code
```suggestion
```

pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -0,0 +1,134 @@
...
+	if err := r.List(ctx, msaList, client.MatchingLabels{MeshNameLabel: mesh.Name, MeshNamespaceLabel: mesh.Namespace}); err != nil {
+		return fmt.Errorf("failed to list ManagedServiceAccount resources managed by mesh %s: %w", mesh.Name, err)
+	}
+
+	secretList := &corev1.SecretList{}
+	if err := r.List(ctx, secretList, client.InNamespace(mesh.Namespace), client.MatchingLabels{
+		multiClusterSecretLabel: "true", MeshNameLabel: mesh.Name, MeshNamespaceLabel: mesh.Namespace,
+	}); err != nil {
+		return fmt.Errorf("failed to list istio-remote secrets managed by mesh %s: %w", mesh.Name, err)
+	}
```
mkolesnik: This is dead code

```suggestion
```

pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -0,0 +1,134 @@
...
+			return fmt.Errorf("failed to delete ManagedServiceAccount %s/%s: %w", msa.Namespace, msa.Name, err)
+		}
+	}
+
+	for _, sec := range secretList.Items {
+		klog.Infof("Deleting an istio remote secret %s", sec.Name)
+		if err := client.IgnoreNotFound(r.Delete(ctx, &sec)); err != nil {
+			return fmt.Errorf("failed to delete an istio remote secret %s: %w", sec.Name, err)
+		}
+	}
```
mkolesnik: This is dead code

```suggestion
```

test/integration/controller_test.go
```diff
@@ -712,6 +723,133 @@ var _ = Describe("MultiClusterMesh Controller", func() {
 		})
 	})
 
+	Context("Endpoint discovery", func() {
+		var cluster1, cluster2, cluster3, cluster4 string
```
mkolesnik: You don't really need 4 clusters, it doesnt add value beyond using 2 clusters.

Please rearrange tests so that by default 1 cluster is created.
Then tests when 2 clusters exist in the set when the mesh is created should create a different cluster, and test basic scenario and removal of cluster.

Another context will test when cluster is added.

You can look at how test are structured for the operator install to get an idea of how to structure them better

test/integration/controller_test.go
```diff
@@ -712,6 +723,133 @@ var _ = Describe("MultiClusterMesh Controller", func() {
...
+				},
+			})
+
+			msa1 := expectManagedServiceAccount(secondNs, secondMesh, cluster1)
+			msa2 := expectManagedServiceAccount(secondNs, secondMesh, cluster2)
+
+			Expect(msa1.Labels[meshcontroller.ManagedByLabel]).To(Equal(meshcontroller.ManagedByValue))
+			Expect(msa1.Labels[meshcontroller.MeshNameLabel]).To(Equal(secondMesh))
+			Expect(msa1.Labels[meshcontroller.MeshNamespaceLabel]).To(Equal(secondNs))
+			Expect(msa1.Labels[meshcontroller.ClusterNameLabel]).To(Equal(cluster1))
```
mkolesnik: Does this add any value?
You already tested it in the default validity case

```suggestion
```

test/integration/controller_test.go
```diff
@@ -712,6 +723,133 @@ var _ = Describe("MultiClusterMesh Controller", func() {
...
+			secondNs := util.UniqueName("second-ns")
+			util.CreateNamespace(ctx, k8sClient, secondNs)
+			secondMesh := util.UniqueName("second-mesh")
+			util.CreateMultiClusterMesh(ctx, k8sClient, secondMesh, secondNs, testClusterSet, meshv1alpha1.MultiClusterMeshSpec{
+				Security: meshv1alpha1.SecurityConfig{
+					Discovery: meshv1alpha1.DiscoveryConfig{
+						TokenValidity: &metav1.Duration{Duration: 15 * time.Minute},
+					},
+				},
+			})
```
mkolesnik: This illustrates quite well why the tests should be restructured, you already have a mesh sol logically theres no need to create another one

test/integration/controller_test.go
```diff
@@ -712,6 +723,133 @@ var _ = Describe("MultiClusterMesh Controller", func() {
...
+
+			Expect(msa1.Labels[meshcontroller.ManagedByLabel]).To(Equal(meshcontroller.ManagedByValue))
+			Expect(msa1.Labels[meshcontroller.MeshNameLabel]).To(Equal(meshName))
+			Expect(msa1.Labels[meshcontroller.MeshNamespaceLabel]).To(Equal(testNs))
+			Expect(msa1.Labels[meshcontroller.ClusterNameLabel]).To(Equal(cluster1))
+			Expect(msa2.Spec.Rotation.Validity).To(Equal(metav1.Duration{Duration: 360 * time.Hour}))
+			Expect(msa2.Labels[meshcontroller.ClusterNameLabel]).To(Equal(cluster2))
+		})
+
+		It("should create ManagedServiceAccount resources with custom TokenValidity value", func() {
```
mkolesnik: I think this has almost no added value when tested on two clusters.
It's very unlikely the code would create non homogeneous MSAs across different clusters.

test/integration/controller_test.go
```diff
@@ -712,6 +723,133 @@ var _ = Describe("MultiClusterMesh Controller", func() {
...
+		When("add a cluster to the ClusterSet", func() {
+			BeforeEach(func() {
+				cluster3 = util.UniqueName("cluster3")
+				util.CreateK8sManagedCluster(ctx, k8sClient, cluster3, testClusterSet)
+			})
+
+			It("should process a ManagedServiceAccount", func() {
+				msa := &msav1beta1.ManagedServiceAccount{}
+				Eventually(func() error {
+					return k8sClient.Get(ctx, types.NamespacedName{
```
mkolesnik: Why not use `expectManagedServiceAccount`?

test/integration/controller_test.go
```diff
@@ -712,6 +723,133 @@ var _ = Describe("MultiClusterMesh Controller", func() {
...
+		})
+
+		When("create a cluster without clusterset label", func() {
+			BeforeEach(func() {
+				cluster4 = util.UniqueName("cluster4")
+				util.CreateK8sManagedCluster(ctx, k8sClient, cluster4, "")
+			})
+
+			It("shouldn't process ManagedServiceAccount", func() {
+				expectNoManagedServiceAccount(testNs, meshName, cluster4)
```
mkolesnik: This mimics "shouldn't process a cluster without clusterset label" from L250 but just that one test. 

If you want to mimic the "referencing an empty ClusterSet" tests from the basic reconciliation, then the tests here should look and do the same but for MSA.

test/integration/controller_test.go
```diff
@@ -712,6 +723,133 @@ var _ = Describe("MultiClusterMesh Controller", func() {
...
+						Namespace: cluster1,
+					}, msa)
+				}).Should(Succeed())
+				Consistently(func() error {
+					msa := &msav1beta1.ManagedServiceAccount{}
+					return k8sClient.Get(ctx, types.NamespacedName{
+						Name:      expectedManagedServiceAccountName(otherNs, otherMesh),
+						Namespace: cluster2,
+					}, msa)
+				}).Should(Succeed())
```
mkolesnik, line 851: Here as well having multiple clusters adds no test value (this is already covered by the basic test)..
I'd test this with a single cluster.

pkg/hub/mesh/endpoint_discovery.go
yxun: This is needed for my next part of the original PR. I can remove this here now.

> Also note: the MSA token secret (created by the MSA controller) already has an ownerReference to the MSA and gets garbage collected when the MSA is deleted. Verified on dev-env.

yes, the cleanup part is not cleaning the MSA token secret. I assume it will be GC by the MSA controller.

Because we need additional label such as "istio/multiCluster" and the secret need to be in control plane ns, I build another secret using the MSA token secret data in my next part of the PR.

I will remove that and we can address those changes in my next part PR. I will create a PR draft today for those discussions.

test/e2e/suite_test.go
yxun: no, I will remove this line.

test/integration/controller_test.go
yxun: Because there was a race condition when adding a cluster to a ClusterSet, the clent Get call failed to find the MSA in my earlier test runs. 
When I wait a lit bit, it gets the expected MSA. So I added `Eventually` wrap here.

mkolesnik: `expectManagedServiceAccount` already uses `Eventually` (L1007-1012).
The race condition you describe is exactly what the helper handles.
You can use it here.

pkg/hub/mesh/endpoint_discovery.go
yxun: https://github.com/stolostron/multicluster-mesh-addon/issues/184

pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -0,0 +1,102 @@
...
+		return fmt.Errorf("failed to create a ManagedServiceAccount %s/%s: %w", cluster.Name, msaName, err)
+	}
+
+	klog.Infof("Successfully created a ManagedServiceAccount %s/%s", cluster.Name, msaName)
+	return nil
+}
+
+// ensureManagedServiceAccountUpdated updates an existing ManagedServiceAccount with the mesh's spec.Security.Discovery.TokenValidity value
+func (r *Reconciler) ensureManagedServiceAccountUpdated(ctx context.Context, mesh *meshv1alpha1.MultiClusterMesh, existing *msav1beta1.ManagedServiceAccount) error {
+	existing.Spec.Rotation.Validity = *mesh.Spec.Security.Discovery.TokenValidity
```
mkolesnik: This always calls `Update` even when the validity has not changed, triggering unnecessary watch events. Compare before updating:

```suggestion
	if existing.Spec.Rotation.Validity == *mesh.Spec.Security.Discovery.TokenValidity {
		return nil
	}
	existing.Spec.Rotation.Validity = *mesh.Spec.Security.Discovery.TokenValidity
```

pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -0,0 +1,102 @@
...
+	if err := r.Update(ctx, existing); err != nil {
+		return fmt.Errorf("failed to update a ManagedServiceAccount %s/%s: %w", existing.Namespace, existing.Name, err)
+	} else {
+		klog.V(4).Infof("Successfully updated a ManagedServiceAccount %s/%s", existing.Namespace, existing.Name)
+	}
```
mkolesnik: nit: `else` after `return` is redundant.

```suggestion
	}
	klog.V(4).Infof("Successfully updated ManagedServiceAccount %s/%s", existing.Namespace, existing.Name)
```

test/integration/controller_test.go
```diff
@@ -734,6 +745,109 @@ var _ = Describe("MultiClusterMesh Controller", func() {
 		})
 	})
 
+	Context("Endpoint discovery", func() {
+		var cluster1, cluster2, cluster3 string
+
+		When("A MultiClusterMesh is created", func() {
```
mkolesnik: nit: 
```suggestion
		When("a MultiClusterMesh is created", func() {
```

pkg/hub/mesh/controller.go
```diff
@@ -160,7 +160,8 @@ func RegisterController(mgr manager.Manager) error {
 //+kubebuilder:rbac:groups=cluster.open-cluster-management.io,resources=managedclustersets,verbs=get;list;watch
 //+kubebuilder:rbac:groups=work.open-cluster-management.io,resources=manifestworks,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=cert-manager.io,resources=certificates,verbs=get;list;watch;create;update;patch;delete
-//+kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch
+//+kubebuilder:rbac:groups=authentication.open-cluster-management.io,resources=managedserviceaccounts,verbs=get;list;watch;create;update;delete
+//+kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch;create;delete
```
mkolesnik: No code in this PR creates or deletes secrets. The `create;delete` verbs were added proactively for the follow-up PR. Add them in the PR that actually needs them.

```suggestion
//+kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch
```

pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -0,0 +1,102 @@
...
+
+// ensureManagedServiceAccountCreated creates ManagedServiceAccount resources using the mesh.Spec.Security.Discovery.TokenValidity value.
+func (r *Reconciler) ensureManagedServiceAccountCreated(ctx context.Context, mesh *meshv1alpha1.MultiClusterMesh, cluster *clusterv1.ManagedCluster) error {
+	msaName := fmt.Sprintf("%s-%s-%s", mesh.Namespace, "istio-reader", mesh.Name)
+	existing := &msav1beta1.ManagedServiceAccount{}
+	if err := r.Get(ctx, key.Of(msaName, cluster.Name), existing); err == nil {
+		if existing.Labels[MeshNameLabel] == mesh.Name || existing.Labels[MeshNamespaceLabel] == mesh.Namespace {
+			return r.ensureManagedServiceAccountUpdated(ctx, mesh, existing)
+		}
+		// when labels are not matching, the existing msa is not owned by the mesh, skipping update and continue creation
```
mkolesnik: Two issues:

1. `||` should be `&&`. An MSA is owned by this mesh only if *both* labels match.

2. When labels do not match (L25), the code falls through to Create which always fails with AlreadyExists since the name is deterministic. Return an error instead:

```suggestion
		if existing.Labels[MeshNameLabel] == mesh.Name && existing.Labels[MeshNamespaceLabel] == mesh.Namespace {
			return r.ensureManagedServiceAccountUpdated(ctx, mesh, existing)
		}
		return fmt.Errorf("ManagedServiceAccount %s/%s exists but is not owned by mesh %s/%s", cluster.Name, msaName, mesh.Namespace, mesh.Name)
```

test/integration/controller_test.go
```diff
@@ -734,6 +745,109 @@ var _ = Describe("MultiClusterMesh Controller", func() {
...
+
+				Expect(msa1.Labels[meshcontroller.ManagedByLabel]).To(Equal(meshcontroller.ManagedByValue))
+				Expect(msa1.Labels[meshcontroller.MeshNameLabel]).To(Equal(meshName))
+				Expect(msa1.Labels[meshcontroller.MeshNamespaceLabel]).To(Equal(testNs))
+				Expect(msa1.Labels[meshcontroller.ClusterNameLabel]).To(Equal(cluster1))
+				Expect(msa2.Spec.Rotation.Validity).To(Equal(metav1.Duration{Duration: 360 * time.Hour}))
+				Expect(msa2.Labels[meshcontroller.ClusterNameLabel]).To(Equal(cluster2))
+			})
+
+			It("should create ManagedServiceAccount resources with custom TokenValidity value", func() {
```
mkolesnik, line 773: This test name says "create" but it actually tests the update path.
The MSA already exists from BeforeEach.

```suggestion
			It("should update ManagedServiceAccount TokenValidity when the mesh spec changes", func() {
```

pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -0,0 +1,102 @@
...
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/klog/v2"
+	clusterv1 "open-cluster-management.io/api/cluster/v1"
+	msav1beta1 "open-cluster-management.io/managed-serviceaccount/apis/authentication/v1beta1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// ensureManagedServiceAccountCreated creates ManagedServiceAccount resources using the mesh.Spec.Security.Discovery.TokenValidity value.
+func (r *Reconciler) ensureManagedServiceAccountCreated(ctx context.Context, mesh *meshv1alpha1.MultiClusterMesh, cluster *clusterv1.ManagedCluster) error {
```
mkolesnik, line 18: This function name says "Created" but it now does create, update, and ownership checking. The name lies about what it does.

The ownership check at L22 belongs inside the update function, not here. A create function should only create. If you keep both create and update in this PR, rename this to `ensureManagedServiceAccount` and move the label check into `ensureManagedServiceAccountUpdated`.

pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -0,0 +1,102 @@
...
+	if err := r.Create(ctx, msa); err != nil {
+		return fmt.Errorf("failed to create a ManagedServiceAccount %s/%s: %w", cluster.Name, msaName, err)
+	}
+
+	klog.Infof("Successfully created a ManagedServiceAccount %s/%s", cluster.Name, msaName)
+	return nil
+}
+
+// ensureManagedServiceAccountUpdated updates an existing ManagedServiceAccount with the mesh's spec.Security.Discovery.TokenValidity value
+func (r *Reconciler) ensureManagedServiceAccountUpdated(ctx context.Context, mesh *meshv1alpha1.MultiClusterMesh, existing *msav1beta1.ManagedServiceAccount) error {
```
mkolesnik: This update handling addresses #120 but this PR is already large as is.
Please remove the update function and ownership check from this PR.
Have the create function just skip if the MSA already exists like it used to.
Handle updates in a follow-up PR on #120.

This simplifies the create function to:

```go
func (r *Reconciler) ensureManagedServiceAccountCreated(...) error {
	if err := r.Get(ctx, key.Of(msaName, cluster.Name), existing); err == nil {
		return nil // already exists
	} else if !errors.IsNotFound(err) {
		return fmt.Errorf("failed to get ManagedServiceAccount: %w", err)
	}
	// ... create ...
}
```

pkg/hub/mesh/controller.go
```diff
@@ -803,7 +818,6 @@ func (r *Reconciler) ensureCacertsManifestWork(ctx context.Context, mesh *meshv1
 	secretName := getCacertsName(cluster.Name)
 	secret := &corev1.Secret{}
 	err := r.Get(ctx, key.Of(secretName, mesh.Namespace), secret)
-
```
mkolesnik, line 806: nit: no need to delete this

mkolesnik (commented):
LGTM overall, we can improve some parts (like integ tests) further once this is merged.

Will wait to see if @jewertow or @sridhargaddam have any further input

mkolesnik (approved):

sridhargaddam (approved):

jewertow (approved):

PR #131, opened 2026-06-15
Implement ManagedServiceAccount Secret distribution using ManifestWork

yxun:
This PR depends on #130 . The first two commits are cherry picked from #130 

This PR builds Istio remote Secret resources from `ManagedServiceAccount` custom resource Secret token.
Then it creates `ManifestWork` resources for distributing those Secrets so that each Istiod control plane can discover remote endpoints from remote cluster API server(s).

Reference:
- https://github.com/istio/istio/blob/master/istioctl/pkg/multicluster/remote_secret.go

 This PR is the last part of the original PR:


yxun:
hold until resolving changes from https://github.com/stolostron/multicluster-mesh-addon/pull/130
I will sync changes between those PRs.

Issue #84, opened 2026-05-25
Configure `make gen` as a pre-commit check for the PRs


sridhargaddam:
Currently, updates to the controller code involving Kubebuilder annotations or instructions can sometimes introduce RBAC changes that are easy to miss during PR reviews. As discussed in the [following PR](https://github.com/stolostron/multicluster-mesh-addon/pull/82#pullrequestreview-4355406599), adding `make gen` as a CI prerequisite/check could help catch these changes automatically.

PR #138, opened 2026-06-17
Remove config/crd from gitignore

sridhargaddam:
When we run make gen-crds, the manifests are generated correctly. However, the config/crd directory was being ignored by .gitignore, which prevented any changes to these files from being included in pull requests.

### DIFF
--- a/config/crd/mesh.open-cluster-management.io_multiclustermeshes.yaml
+++ b/config/crd/mesh.open-cluster-management.io_multiclustermeshes.yaml
@@ -1,0 +1,279 @@
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.20.1
+  name: multiclustermeshes.mesh.open-cluster-management.io
+spec:
+  group: mesh.open-cluster-management.io
+  names:
+    kind: MultiClusterMesh
+    listKind: MultiClusterMeshList
+    plural: multiclustermeshes
+    singular: multiclustermesh
+  scope: Namespaced
+  versions:
+  - name: v1alpha1
+    schema:
+      openAPIV3Schema:
+        description: MultiClusterMesh represents a multi-cluster service mesh configuration
+        properties:
+          apiVersion:
+            description: |-
+              APIVersion defines the versioned schema of this representation of an object.
+              Servers should convert recognized schemas to the latest internal value, and
+              may reject unrecognized values.
+              More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources
+            type: string
+          kind:
+            description: |-
+              Kind is a string value representing the REST resource this object represents.
+              Servers may infer this from the endpoint the client submits requests to.
+              Cannot be updated.
+              In CamelCase.
+              More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds
+            type: string
+          metadata:
+            type: object
+          spec:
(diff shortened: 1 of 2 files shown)


config/crd/mesh.open-cluster-management.io_multiclustermeshes.yaml
```diff
@@ -0,0 +1,279 @@
+---
```
FilipB, line 1: Do we want to store generated files in git hub? This is not source of truth.

sridhargaddam: We do the same in Sail Operator isnt it? 

For example see https://github.com/istio-ecosystem/sail-operator/blob/f527b2d1e78f888e7e37e21b3c32ad2007db4fc9/Makefile.core.mk#L472

Generated CRDs: https://github.com/istio-ecosystem/sail-operator/tree/main/chart/crds

Also, in submariner-addon the manifests are checked in github.
https://github.com/stolostron/submariner-addon/tree/main/deploy/config/crds

sridhargaddam: We have an open [issue](https://github.com/stolostron/multicluster-mesh-addon/issues/84) for running "make gen" as a pre-commit check so that the generated manifests are in sync with the actual APIs.

FilipB: Yes, with a gen-check similar to what we have in sail operator, it's fine.

FilipB (approved):

mkolesnik:
This was intentional, I'm not sure what's the reason to store these CRDs which we use just for testing in the repo?!

This just introduces sync issues and more friction for maintenance IMO.

> We do the same in Sail Operator isnt it?

This is not a valid argument...
Some addons do store it and some don't so if you do want to store it there should be a reason that justifies the extra maintenance burden?

sridhargaddam:
> This is not a valid argument... Some addons do store it and some don't so if you do want to store it there should be a reason that justifies the extra maintenance burden?

thanks for the feedback. I think those are valid concerns, so I wanted to explain what I was optimizing for here.

Just one small clarification: `config/crd/` isn't where the test CRDs live. Those are under `test/integration/crds/` and remain gitignored. `config/crd/` contains our own CRD manifest (MultiClusterMesh) and many projects include it in the repos.

The main reasons I was leaning toward checking it in are:

1. Review visibility. When someone updates types.go, the generated CRD changes show up directly in the PR. That gives reviewers a chance to catch schema-related issues, validation changes, or unintended API modifications before they make it into a deployment.

2. Detecting drift automatically. If we track the generated CRD, we can add a simple verification step in CI to ensure generated artifacts stay in sync with the source. We already have a TODO in the Makefile for something along those lines and an [open issue](https://github.com/stolostron/multicluster-mesh-addon/issues/84) for that.

I agree that generated files can become a maintenance headache if they're not kept in sync. My thinking was that the CI check would largely remove that burden by catching drift automatically rather than relying on people to remember.

That said, I don't feel strongly enough about it to block on the approach. If the preference is to keep generated artifacts out of the repository, I'm happy to revert the change. Just wanted to explain the thinking behind it. What do you think?

mkolesnik:
I was wrong here, sorry.
I confused `config/crd/` (our own CRD) with `test/integration/crds/` (test CRDs) and didn't check the convention before pushing back.
Checked across ACM addons and major k8s projects and they all commit generated CRDs.
Your reasoning is sound.

I'll pick up #84 to add verify-gen so we catch drift like the one from #127.

PR #139, opened 2026-06-17
certificate: Set X.509 subject fields and SAN on Istio CA certificates

### LINKED ISSUE #65: Investigate and set subject fields in Istio CA
**Background context:**

We need to configure custom X.509 Subject fields in our Istio CA certificates. Our goal is to include identity data like the cluster name and trust domain within the subject.

However, using the Common Name (CN) field isn't working well for this use case, primarily due to the strict 64-character limit imposed by the X.509 standard (RFC 5280) and industry best practices.

**TODO:**

- Evaluate other Subject fields - such as Organizational Unit (OU), Organization (O), or Custom OIDs - to determine which is best suited for storing cluster and trust domain metadata.
- Ensure the chosen fields comply with industry standards and safely avoid max-length constraints.

jewertow:
Configure the cert-manager Certificate with Organization (trust domain),
Organizational Unit (cluster name with deterministic truncation for names exceeding 63 chars),
and a URI SAN for full cluster traceability.

Fixes #65

### DIFF
--- a/pkg/hub/mesh/certificate.go
+++ b/pkg/hub/mesh/certificate.go
@@ -5,18 +5,28 @@ import (
 	"fmt"
 )
 
-const maxOULength = 64
+const (
+	maxOULength    = 64
+	maxLabelLength = 63
+)
 
-// formatOU formats a cluster name for use as the X.509 Organizational Unit (OU) field.
-// If the name fits within the 64-character limit, it is returned as-is.
-// Otherwise, it is truncated to 55 characters and suffixed with a dash and the
-// first 8 hex characters of the SHA-256 hash of the full name (55 + 1 + 8 = 64).
 func formatOU(clusterName string) string {
-	if len(clusterName) <= maxOULength {
-		return clusterName
-	}
+	return truncateWithHash(clusterName, maxOULength)
+}
 
-	return fmt.Sprintf("%.55s-%.4x", clusterName, sha256.Sum256([]byte(clusterName)))
+func formatLabelValue(clusterName string) string {
+	return truncateWithHash(clusterName, maxLabelLength)
+}
+
+// truncateWithHash returns s as-is if it fits within maxLen.
+// Otherwise, it truncates to (maxLen - 9) characters and appends a dash
+// followed by the first 8 hex characters of the SHA-256 hash of the full string.
+func truncateWithHash(s string, maxLen int) string {
+	if len(s) <= maxLen {
+		return s
+	}
+	hash := fmt.Sprintf("%.4x", sha256.Sum256([]byte(s)))
+	return s[:maxLen-9] + "-" + hash
 }
 
 func certURI(clusterName, trustDomain string) string {
--- a/pkg/hub/mesh/certificate.go
+++ b/pkg/hub/mesh/certificate.go
@@ -3,7 +3,6 @@ package mesh
 import (
 	"crypto/sha256"
 	"fmt"
-	"strings"
 )
 
 const maxOULength = 63
@@ -21,8 +20,6 @@ func formatOU(clusterName string) string {
 	return clusterName[:54] + "-" + hash[:8]
 }
 
-// certDNSName generates the SAN DNS name for an Istio CA certificate.
-// Trailing dashes are stripped from the cluster name to produce a valid DNS name.
-func certDNSName(clusterName, trustDomain string) string {
-	return strings.TrimRight(clusterName, "-") + ".istio-ca." + trustDomain
+func certURI(clusterName, trustDomain string) string {
+	return "spiffe://" + trustDomain + "/cluster/" + clusterName + "/ca/istio-ca"
 }
--- a/pkg/hub/mesh/certificate.go
+++ b/pkg/hub/mesh/certificate.go
@@ -1,0 +1,28 @@
+package mesh
+
+import (
+	"crypto/sha256"
+	"fmt"
+	"strings"
+)
+
+const maxOULength = 63
+
+// formatOU formats a cluster name for use as the X.509 Organizational Unit (OU) field.
+// If the name fits within the 63-character limit, it is returned as-is.
+// Otherwise, it is truncated to 54 characters and suffixed with a dash and the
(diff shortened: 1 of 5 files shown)


pkg/hub/mesh/controller.go
```diff
@@ -788,7 +788,11 @@ func (r *Reconciler) ensureCertificateForCluster(ctx context.Context, mesh *mesh
 				WithLabels(meshOwnedLabels(mesh, cluster.Name))).
 			WithDuration(metav1.Duration{Duration: 60 * Day}).
 			WithRenewBefore(metav1.Duration{Duration: 15 * Day}).
-			WithCommonName("Intermediate Istio CA").
+			WithCommonName("Istio CA").
+			WithSubject(certmanagerapply.X509Subject().
+				WithOrganizations(mesh.Name).
+				WithOrganizationalUnits(formatOU(cluster.Name))).
+			WithDNSNames(certDNSName(cluster.Name, mesh.Name)).
```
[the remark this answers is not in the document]
jewertow: Replaced with URI

pkg/hub/mesh/certificate.go
```diff
@@ -0,0 +1,28 @@
+package mesh
```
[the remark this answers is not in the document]
jewertow: The project has no established convention of adding copyright headers to Go files — none of the existing files (controller.go, types.go, main.go, etc.) have them.

pkg/hub/mesh/controller.go
```diff
@@ -788,7 +788,11 @@ func (r *Reconciler) ensureCertificateForCluster(ctx context.Context, mesh *mesh
 				WithLabels(meshOwnedLabels(mesh, cluster.Name))).
 			WithDuration(metav1.Duration{Duration: 60 * Day}).
 			WithRenewBefore(metav1.Duration{Duration: 15 * Day}).
-			WithCommonName("Intermediate Istio CA").
+			WithCommonName("Istio CA").
+			WithSubject(certmanagerapply.X509Subject().
+				WithOrganizations(mesh.Name).
+				WithOrganizationalUnits(formatOU(cluster.Name))).
+			WithURIs(certURI(cluster.Name, mesh.Name)).
```
yxun: Should we use `mesh.Name + ".local"`  as the certURI trustDomain name here ?
I see the test case is using `mesh.local`

jewertow: No, we derive the trust domain from MultiClusterMesh resource name. Resource names can contain the dots, so we don't need to append anything.

I also encapsulated `mesh.Name` to `GetTrustDomain()` to make it clear where does it come from.

Btw. I don't like this approach, but we already agreed on that.

test/integration/controller_test.go
```diff
@@ -513,6 +513,21 @@ var _ = Describe("MultiClusterMesh Controller", func() {
 				Expect(*ownerRef.BlockOwnerDeletion).To(BeTrue())
 			})
 
+			It("should set Subject with Organization and OrganizationalUnit", func() {
```
mkolesnik: Would be good to add a test with a 64+ character cluster name to verify the OU truncation wires through end-to-end.
The unit test covers `formatOU` logic, but nothing verifies the controller actually calls it.

pkg/hub/mesh/certificate.go
```diff
@@ -0,0 +1,25 @@
...
+
+import (
+	"crypto/sha256"
+	"fmt"
+)
+
+const maxOULength = 63
```
mkolesnik: nit: `ub-organizational-unit-name` in X.520 is 64, not 63.
Not a practical issue, but the constant is off-by-one vs the spec.

```suggestion
const maxOULength = 64
```

pkg/hub/mesh/certificate.go
```diff
@@ -0,0 +1,25 @@
...
+// If the name fits within the 63-character limit, it is returned as-is.
+// Otherwise, it is truncated to 54 characters and suffixed with a dash and the
+// first 8 hex characters of the SHA-256 hash of the full name (54 + 1 + 8 = 63).
+func formatOU(clusterName string) string {
+	if len(clusterName) <= maxOULength {
+		return clusterName
+	}
+
+	hash := fmt.Sprintf("%x", sha256.Sum256([]byte(clusterName)))
+	return clusterName[:54] + "-" + hash[:8]
```
mkolesnik: nit: Formatting can truncate, the hex will print first 8 hex chars (for 4 bytes).
This simplifies the call and leaves the formatting to `fmt`

```suggestion
	return fmt.Sprintf("%.54s-%.4x", clusterName, sha256.Sum256([]byte(clusterName)))
```

mkolesnik (commented):
Few small suggestions, otherwise LGTM

pkg/hub/mesh/certificate.go
jewertow: Good catch - done.

pkg/hub/mesh/certificate.go
jewertow: Thanks! Done.

test/integration/controller_test.go
jewertow: I changed this test case to make sure that `formatOU` is called.

Btw. it turned out that we these long cluster names were also used in resource labels, and k8s rejected them.

mkolesnik (commented):
LGTM, if there are more reviewers PTAL

sridhargaddam (approved):
LGTM

mkolesnik:
Hey @jewertow, the truncation code from this PR can't trigger in practice. Cluster names double as hub namespace names in OCM, so they're capped at 63 chars. OU limit is 64, label limit is 63, both always return the input unchanged.

The 72-char integration test only worked because it bypassed the test helper and skipped the namespace. After #106 merged it started failing.

Opened #154 to drop the dead code and simplify the test.

mkolesnik:
Also, the mesh name goes into the X.509 Organization field (64-char limit) and into label values (63-char limit) but had no length validation on the CRD. Mesh names can be up to 253 chars (DNS subdomain), so a long name would fail at label creation with a confusing error. Filed #155 to add a CEL validation capping it at 63.

PR #142, opened 2026-06-18
Simplify dev-clean-meshes and replace ClusterIssuer with Issuer

mkolesnik:
Replace the cluster-scoped ClusterIssuer with a namespace-scoped self-signed Issuer.
The ClusterIssuer was unnecessary since the bootstrap issuer only signs a single Certificate in mesh-system.
This simplifies dev-clean-meshes to just deleting meshes, ManifestWorks, and the namespace.

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,7 +160,7 @@ See the [Kind known issues](https://kind.sigs.k8s.io/docs/user/known-issues/#pod
 kubectl create namespace mesh-system
 ```
 
-2. Create the cert-manager trust chain (ClusterIssuer, root CA Certificate, and CA-backed Issuer):
+2. Create the cert-manager trust chain (self-signed Issuer, root CA Certificate, and CA-backed Issuer):
 ```bash
 kubectl apply -f samples/cert-manager-issuer.yaml
 ```
@@ -178,7 +178,7 @@ For more configuration options, see the [samples](./samples/) directory:
 - **[complete.yaml](./samples/complete.yaml)** - All available fields with documentation
 - **[openshift.yaml](./samples/openshift.yaml)** - OpenShift-specific configuration
 - **[pinned-version.yaml](./samples/pinned-version.yaml)** - Version pinning with manual approval
-- **[cert-manager-issuer.yaml](./samples/cert-manager-issuer.yaml)** - cert-manager trust chain (ClusterIssuer + root CA + Issuer)
+- **[cert-manager-issuer.yaml](./samples/cert-manager-issuer.yaml)** - cert-manager trust chain (self-signed Issuer + root CA + CA-backed Issuer)
 
 #### What the Addon Does
 
--- a/Makefile
+++ b/Makefile
@@ -284,9 +284,6 @@ setup-mesh: ## Create cert-manager trust chain, mesh-system namespace, and Multi
 dev-clean-meshes: ## Delete all mesh resources, cert-manager trust chain, and mesh-system namespace
 	kubectl --kubeconfig=$(HUB_KUBECONFIG) delete multiclustermeshes -A --all --ignore-not-found=true
 	kubectl --kubeconfig=$(HUB_KUBECONFIG) delete manifestwork -A -l app.kubernetes.io/managed-by=multicluster-mesh-addon --ignore-not-found=true
-	kubectl --kubeconfig=$(HUB_KUBECONFIG) delete issuer mesh-root-ca -n mesh-system --ignore-not-found=true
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -160,7 +160,7 @@ See the [Kind known issues](https://kind.sigs.k8s.io/docs/user/known-issues/#pod
 kubectl create namespace mesh-system
 ```
 
-2. Create the cert-manager trust chain (ClusterIssuer, root CA Certificate, and CA-backed Issuer):
+2. Create the cert-manager trust chain (self-signed Issuer, root CA Certificate, and CA-backed Issuer):
```
jewertow, line 163: `root CA Certificate` is not the root - it's an intermediate CA, and `self-signed Issuer` is the actual root.

jewertow (changes requested):
I do not understand why we have so complex certificate chain in samples and tests.
I don't think it follows any production-ready pattern as most users point to an external or cloud-provided issuer.

Our tests and samples should only have a self-signed issuer to keep this part as simple as possible:
```
apiVersion: cert-manager.io/v1
kind: Issuer
metadata:
  name: root-ca
spec:
  selfSigned: {}
```

sridhargaddam:
> Our tests and samples should only have a self-signed issuer to keep this part as simple as possible:
> 
> ```
> apiVersion: cert-manager.io/v1
> kind: Issuer
> metadata:
>   name: root-ca
> spec:
>   selfSigned: {}
> ```

Looking at the controller code, it seems to create a per-cluster intermediate CA Certificates referencing the issuer from `spec.security.trust.certManager.issuerRef.name with kind: Issuer`. If that issuer is a selfSigned issuer, each intermediate CA certificate would be self-signed (cert-manager's self-signed issuer means each certificate signs itself with its own key). The clusters would NOT share a common trust root, they'd each have an independent self-signed intermediate CA, breaking cross-cluster mTLS.

**3-step chain**
1. Self-signed Issuer bootstraps a root CA key pair
2. Root CA Certificate holds that key pair in a Secret
3. CA-backed Issuer uses that Secret to sign intermediate CAs // this is what gives all clusters a shared trust root

So, a 3-step chain seems to be necessary for cross-cluster trust. A single self-signed Issuer produces independent, unrelated CAs on each cluster.

I cross-checked this understanding with Claude as well. Am I missing something?

README.md
mkolesnik: The Certificate IS the root.
A cert-manager self-signed Issuer doesn't have its own cert or key pair, it just tells cert-manager to create self-signed certificates.
The resulting `mesh-root-ca` cert has `subject == issuer` (signs itself), which is the definition of a root CA.

Verified on dev-env:
```
root CA:    subject=CN=Mesh Root CA, issuer=CN=Mesh Root CA  (self-signed)
cluster1:   subject=CN=Intermediate Istio CA, issuer=CN=Mesh Root CA
cluster2:   subject=CN=Intermediate Istio CA, issuer=CN=Mesh Root CA
```

The per-cluster certs signed by the CA-backed Issuer are the intermediates.

mkolesnik:
Sridhar is right, I tested both approaches on a dev-env with MetalLB + east-west gateways.

**3-step chain** (this PR): cross-cluster mTLS works. Both clusters share `CN=Mesh Root CA` as root, traffic load-balances across clusters (v1 + v2 responses).

**Single self-signed Issuer**: each cluster gets its own independent self-signed CA:
```
cluster1:   subject=CN=Intermediate Istio CA, issuer=CN=Intermediate Istio CA, serial=589BD7...
cluster2:   subject=CN=Intermediate Istio CA, issuer=CN=Intermediate Istio CA, serial=2DE3E2...
```
Same CN but different keys (different serials). Each signs itself, no shared root.
Cross-cluster mTLS fails:
```
TLS_error: CERTIFICATE_VERIFY_FAILED: self signed certificate in certificate chain
```
0/30 requests reached the remote cluster.

A self-signed Issuer generates a new key pair per Certificate request, there's no persistent root CA to share.
The 3-step chain produces exactly one root key pair that backs all per-cluster intermediate CAs.

jewertow:
> cert-manager's self-signed issuer means each certificate signs itself with its own key

Sorry, I forgot about that. Thanks for claryfing that.

README.md
jewertow: Yeah, sorry. I forhat that `selfSigned: {}` does not work the same way as an Issuer created from a self-signed key pair.

jewertow (approved):

PR #143, opened 2026-06-18
Fix make help to show targets with digits

sridhargaddam:
The help target's pattern used [a-zA-Z_-]+ which excluded digits, so targets like test-e2e were silently missing from the output. This PR fixes it.


mkolesnik (commented):
/lgtm

PR #145, opened 2026-06-18
Enforce immutability and non-empty validation for spec.clusterSet

### LINKED ISSUE #97: Add validating webhook for spec.clusterSet
`spec.clusterSet` needs two validations that can't be enforced by kubebuilder markers or the reconciler:

1. **Immutability after creation.** Changing the ClusterSet is equivalent to tearing down the entire mesh (different clusters, certificates, discovery tokens).
The reconciler can't detect the change since it only sees the current value, and old resources get orphaned.
Users should delete and recreate the mesh CR instead.

2. **Non-empty value.** An empty `clusterSet` passes CRD validation today (`+required` only prevents the field from being absent, not empty string).
The reconciler handles it gracefully but the error is delayed and unclear.

This requires a validating webhook since neither constraint can be enforced declaratively.

sridhargaddam:
A Mesh CR should not be allowed to switch ClusterSets after creation, since doing so can orphan ManifestWorks, certificates, and discovery tokens associated with the original ClusterSet.

This PR adds CEL validation rules that:
* Reject empty values using `MinLength=1`
* Prevent updates using `self == oldSelf`

Fixes: https://github.com/stolostron/multicluster-mesh-addon/issues/97

### DIFF
--- a/chart/crds/mesh.open-cluster-management.io_multiclustermeshes.yaml
+++ b/chart/crds/mesh.open-cluster-management.io_multiclustermeshes.yaml
@@ -43,7 +43,11 @@ spec:
               clusterSet:
                 description: ClusterSet references the ACM ManagedClusterSet that
                   defines cluster membership
+                minLength: 1
                 type: string
+                x-kubernetes-validations:
+                - message: spec.clusterSet is immutable
+                  rule: self == oldSelf
               controlPlane:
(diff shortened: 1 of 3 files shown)


mkolesnik (approved):

sridhargaddam:
@mkolesnik I had to rebase it to fix some merge conflicts. Can you PTAL once again, thanks.

PR #148, opened 2026-06-19
Migrate from Kustomize to Helm charts

FilipB:
- Create Helm chart structure in /chart/ (project root)
- Generate CRDs and ClusterRole directly to chart/ (single source of truth)
- Use Helm's .Release.Namespace instead of values-based namespace
- Remove all Kustomize files and dependencies
- Add dynamic Helm installation (v4.2.2) if not in PATH
- Update documentation (README.md, AGENTS.md)

Chart structure:
- chart/crds/ - CRDs generated by make gen-crds
- chart/templates/clusterrole.yaml - generated by make gen-rbac
- chart/templates/ - all other Kubernetes manifests
- chart/values.yaml - minimal configuration (image, platform)

Configuration:
- ADDON_NAMESPACE controls deployment namespace (default: multicluster-mesh-system)
- PLATFORM=kind sets platform-specific values via --set
- No symlinks, no duplication - single source of truth for all files


chart/values.yaml
```diff
@@ -0,0 +1,11 @@
...
+  repository: quay.io/sail-dev/multicluster-mesh-addon
+  tag: latest
+  # pullPolicy: Always is used for OpenShift (default).
+  # For Kind, this is overridden to IfNotPresent via --set because images are
+  # side-loaded into Kind via "kind load image-archive" and never pushed to a
+  # registry. With "Always", the kubelet would try to pull from the registry
+  # and fail with ErrImagePull.
+  pullPolicy: Always
+
+platform: openshift
```
jewertow, line 11: nit: we don't have any condition, like `eq .Values.platform "openshift"` so we do not need to set it at all.

jewertow (approved):

PR #149, opened 2026-06-19
certificate: add support for cluster issuer

### LINKED ISSUE #71: Add support for ClusterIssuer

jewertow:
Fixes #71

### DIFF
--- a/samples/basic.yaml
+++ b/samples/basic.yaml
@@ -2,14 +2,22 @@ apiVersion: mesh.open-cluster-management.io/v1alpha1
 kind: MultiClusterMesh
 metadata:
   name: basic-mesh
-  namespace: mesh-system
+  namespace: open-cluster-management
 spec:
   # Reference to the ManagedClusterSet that defines which clusters participate in the mesh
   # NOTE: Update this to match your ManagedClusterSet name
-  clusterSet: mesh-cluster-set
+  clusterSet: default
   # Minimal security configuration using cert-manager for trust
   security:
     trust:
       certManager:
         issuerRef:
-          name: mesh-root-ca
+          name: root-ca
+---
+apiVersion: cert-manager.io/v1
+kind: Issuer
+metadata:
+  name: root-ca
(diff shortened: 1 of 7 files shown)


samples/basic.yaml
```diff
@@ -2,14 +2,22 @@ apiVersion: mesh.open-cluster-management.io/v1alpha1
...
+  clusterSet: default
   # Minimal security configuration using cert-manager for trust
   security:
     trust:
       certManager:
         issuerRef:
-          name: mesh-root-ca
+          name: root-ca
+---
+apiVersion: cert-manager.io/v1
```
sridhargaddam: Adding a hold on this change until we hear back from @jewertow about the following clarification - https://github.com/stolostron/multicluster-mesh-addon/pull/142#issuecomment-4757375283

sridhargaddam (changes requested):

samples/basic.yaml
jewertow: Sorry, I reverted this unintentional change.

sridhargaddam (approved):
LGTM, thanks @jewertow

mkolesnik:
Hey @jewertow, now that this merged I wanted to flag something for awareness.

When two meshes reference the same ClusterIssuer (or even the same namespaced Issuer), they share a trust root, which means cross-mesh mTLS succeeds between them. That's fine for single-tenant setups but worth knowing about in multi-tenant ones.

This isn't a problem with the PR itself, just a documentation gap. Filed #153 to clarify the trust isolation implications in the design doc.

PR #151, opened 2026-06-22
Add retry wrapper for kubectl wait and rollout status in dev-env

mkolesnik:
kubectl wait crashes instead of retrying when `.status.conditions` is nil, which happens when the API server has not finished processing a resource right after creation. This caused flaky e2e failures during OLM setup.

#150 fixed this for the one OLM CRD wait, but the same race can hit any kubectl wait call. It also exposed retry count and delay as parameters, which no caller has reason to tune.

Simplify retry to use fixed defaults (3 attempts, 2s delay) and wrap all kubectl wait calls that might hit it.

### DIFF
--- a/hack/dev-env.sh
+++ b/hack/dev-env.sh
@@ -18,19 +18,24 @@ warn() { echo "WARNING: $*" >&2; }
 err() { echo "ERROR: $*" >&2; exit 1; }
 
 retry() {
-    local retries=$1 delay=$2
-    shift 2
-    local attempt
-    for attempt in $(seq 1 "$retries"); do
-        if "$@"; then
+    local attempts=3 delay=2 attempt=1
+    while (( attempt <= attempts )); do
+        local output rc=0
+        output=$("$@" 2>&1) || rc=$?
+        echo "${output}"
+        if (( rc == 0 )); then
             return 0
         fi
-        if (( attempt < retries )); then
-            log "Attempt ${attempt}/${retries} failed, retrying in ${delay}s..."
-            sleep "$delay"
+        if [[ "${output}" == *"timed out waiting for"* ]]; then
+            err "Command timed out, not retrying: $*"
         fi
+        if (( attempt == attempts )); then
+            err "Command failed after ${attempts} attempts: $*"
+        fi
+        log "Attempt ${attempt}/${attempts} failed, retrying in ${delay}s..."
+        sleep "${delay}"
+        (( attempt++ ))
     done
-    err "Command failed after ${retries} attempts: $*"
 }
 
 MIN_INOTIFY_WATCHES=524288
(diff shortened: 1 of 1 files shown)


hack/dev-env.sh
```diff
@@ -18,19 +18,18 @@ warn() { echo "WARNING: $*" >&2; }
...
-            log "Attempt ${attempt}/${retries} failed, retrying in ${delay}s..."
-            sleep "$delay"
+        if (( attempt == attempts )); then
+            return 1
         fi
+        log "Attempt ${attempt}/${attempts} failed, retrying in ${delay}s..."
+        sleep "${delay}"
+        (( attempt++ ))
     done
-    err "Command failed after ${retries} attempts: $*"
```
sridhargaddam, line 33: Can we include this error log as it helps users understand the exact reason why the command failed?

hack/dev-env.sh
```diff
@@ -153,8 +150,8 @@ install_olm() {
         kubectl --kubeconfig="${kubeconfig}" apply -f "${olm_base_url}/olm.yaml"
 
         log "Waiting for OLM components to be ready on ${cluster}..."
-        kubectl --kubeconfig="${kubeconfig}" rollout status deployment/olm-operator -n olm --timeout=180s
-        kubectl --kubeconfig="${kubeconfig}" rollout status deployment/catalog-operator -n olm --timeout=180s
+        retry kubectl --kubeconfig="${kubeconfig}" rollout status deployment/olm-operator -n olm --timeout=180s
+        retry kubectl --kubeconfig="${kubeconfig}" rollout status deployment/catalog-operator -n olm --timeout=180s
```
sridhargaddam: Using `retry` for `kubectl --wait..` is fine, but using it on `kubectl rollout status...` is redundant and IMO is not required as it has its own built-in mechanism to poll/watch.  If the deployment is genuinely broken, it simply increases the execution time before returning an error.

hack/dev-env.sh
```diff
@@ -184,8 +181,8 @@ install_cert_manager() {
         "https://github.com/cert-manager/cert-manager/releases/download/${CERT_MANAGER_VERSION}/cert-manager.yaml"
 
     log "Waiting for cert-manager to be ready..."
-    kubectl --kubeconfig="${hub_kubeconfig}" rollout status deployment/cert-manager -n cert-manager --timeout=120s
-    kubectl --kubeconfig="${hub_kubeconfig}" rollout status deployment/cert-manager-webhook -n cert-manager --timeout=120s
+    retry kubectl --kubeconfig="${hub_kubeconfig}" rollout status deployment/cert-manager -n cert-manager --timeout=120s
+    retry kubectl --kubeconfig="${hub_kubeconfig}" rollout status deployment/cert-manager-webhook -n cert-manager --timeout=120s
```
sridhargaddam: Same here, we can remove the retry from these commands.

hack/dev-env.sh
```diff
@@ -310,22 +307,22 @@ setup_mesh() {
     fi
 
     log "Waiting for cert-manager-cainjector to be ready..."
-    kubectl --kubeconfig="${hub_kubeconfig}" rollout status deployment/cert-manager-cainjector \
+    retry kubectl --kubeconfig="${hub_kubeconfig}" rollout status deployment/cert-manager-cainjector \
```
sridhargaddam: Same here...

hack/dev-env.sh
```diff
@@ -115,7 +114,7 @@ create_clusters() {
             --wait 120s
 
         log "Waiting for cluster ${cluster} API to be ready..."
-        kubectl --kubeconfig="${kubeconfig}" wait --for=condition=Ready nodes --all --timeout=120s
+        retry kubectl --kubeconfig="${kubeconfig}" wait --for=condition=Ready nodes --all --timeout=120s
```
sridhargaddam: The previous command `kind create cluster --wait 120s` already waits for nodes to be Ready. By the time we are here, nodes are fully initialized with conditions populated, so using the `retry` option here is not required.

mkolesnik:
@coderabbitai configuration

sridhargaddam (approved):
LGTM, thanks

PR #152, opened 2026-06-22
Extract on() and require_clusters() helpers in dev-env

mkolesnik:
Every kubectl, clusteradm, and kind call repeated the same
`--kubeconfig="${DEV_KUBE_DIR}/${cluster}.config"` argument, burying
the actual command in boilerplate. Similarly, every function had
its own kubeconfig file-existence check with slightly different
error messages.

`on()` appends `--kubeconfig` for a given cluster, letting calls read
naturally: `on "${cluster}" retry kubectl wait ...`
`require_clusters()` fails early if any cluster config is missing.

Depends-On: #151


mkolesnik:
/hold
Depends on #151

hack/dev-env.sh
```diff
@@ -279,60 +253,54 @@ join_clusters() {
...
+    on "${HUB}" retry kubectl wait issuer/mesh-root-ca \
         -n mesh-system --for=condition=Ready --timeout=60s
 
     log "Creating MultiClusterMesh CR"
-    kubectl --kubeconfig="${hub_kubeconfig}" apply -f "${SCRIPT_DIR}/samples/basic.yaml"
+    on "${HUB}" kubectl apply -f "${SCRIPT_DIR}/samples/basic.yaml"
 
     log "Mesh setup complete. The controller will now reconcile the mesh."
-    log "Monitor progress: kubectl --kubeconfig=${hub_kubeconfig} get multiclustermesh -n mesh-system"
+    log "Monitor progress: $(on "${HUB}" echo kubectl get multiclustermesh -n mesh-system)"
```
sridhargaddam, line 303: nit:

```suggestion
    log "Monitor progress: on ${HUB} kubectl get multiclustermesh -n mesh-system"
```

sridhargaddam (commented):
One small nit, otherwise LGTM. @FilipB do you want to take a second look?

hack/dev-env.sh
```diff
@@ -104,119 +119,87 @@ create_clusters() {
...
+        if on "${cluster}" kubectl get deployment olm-operator -n olm &>/dev/null; then
             log "OLM already installed on ${cluster}, skipping"
             continue
         fi
 
         log "Installing OLM ${OLM_VERSION} on ${cluster}..."
 
-        kubectl --kubeconfig="${kubeconfig}" apply --server-side -f "${olm_base_url}/crds.yaml"
-        # kubectl wait errors immediately when .status.conditions is nil
-        # (before the API server has processed the CRD), so retry a few times.
```
FilipB: Let's not remove the comment if it's still true. It's confusing without that comment why you need to retry the wait command. Is it necessary to retry all other wait commands as well? I have never seen to retry wait command before. It's a bit confusing.

sridhargaddam: Not for all. See https://github.com/stolostron/multicluster-mesh-addon/pull/151

FilipB: Alternative solution without retries might be:
```
kubectl wait --for=create ...
kubectl wait --for=condition=  ...
```

mkolesnik:
Merged upstream/main (includes #151), resolved conflict in `hack/dev-env.sh` (kept `on()` helper pattern over inline kubeconfig passing, no logical changes).

hack/dev-env.sh
mkolesnik: The comment removal and the retry pattern itself are from #151, which this PR depends on.
This PR only replaces `kubectl --kubeconfig=... wait` with `on cluster kubectl wait`, it does not change the retry logic.

The `kubectl wait --for=create` idea is worth looking into as a follow-up to #151, which was merged.

hack/dev-env.sh
mkolesnik: The `$(on "${HUB}" echo ...)` is intentional here.
It runs `echo` through `on()` which appends `--kubeconfig=path`, so the log shows a copy-pasteable command like:

```
kubectl get multiclustermesh -n mesh-system --kubeconfig=/path/to/hub.config
```

Your suggestion would print `on hub1 kubectl get ...` which is not a command the user can run outside of dev-env.sh (on() is a script-local function).

sridhargaddam (approved):
LGTM

PR #157, opened 2026-06-22
Add ManagedServiceAccount addon installation in dev-env

yxun:
This PR fixes an e2e test job error by adding the ManagedServiceAccount addon installation in the dev-env script.
 
Currently, we are getting `no matches for kind "ManagedServiceAccount"` error when running e2e job in https://github.com/stolostron/multicluster-mesh-addon/pull/130 .

```
status:
  conditions:
  - lastTransitionTime: "2026-06-22T16:29:11Z"
    message: 'failed to create ManagedServiceAccounts: failed to get ManagedServiceAccount
      cluster1/test-ns-k87bkn-istio-reader-test-mesh-xv4m9m: no matches for kind "ManagedServiceAccount"
      in version "authentication.open-cluster-management.io/v1beta1"'
    observedGeneration: 2
    reason: ReconcileError
    status: "False"
```

Because it cannot delete a mesh CR in that situation. `[It] cleans up resources on mesh deletion fails.`

Comparing to the envtest suite setup, this PR installs the required addon dependency in the dev-env KinD hub cluster. The step is following the doc below:
https://github.com/open-cluster-management-io/managed-serviceaccount#installation-steps

This helps fixing the e2e test errors when creating or deleting a mesh with the ManagedServiceAccount CRD.


test/e2e/suite_test.go
```diff
@@ -20,17 +21,25 @@ import (
 	clusterv1beta2 "open-cluster-management.io/api/cluster/v1beta2"
 	workv1 "open-cluster-management.io/api/work/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/envtest"
```
FilipB: I thought envtest is not needed at all for e2e tests. It's used to mock etcd and the Kubernetes API server for integration tests isn't it?

sridhargaddam: Good point. He seems to be using the utility function `envtest.InstallCRDs` to install CRDs. I feel, importing a heavy weight envtest package just to install the CRDs is an overkill. 

IMO, a better fix would be to install the missing CRDs as part of `make dev-env` setup. Currently, the dev-env already creates Kind clusters and installs OCM. The ManagedServiceAccount CRD should be installed there as part of the cluster setup, alongside the other OCM components. The e2e test suite shouldn't need to bootstrap the cluster's CRDs, it should assume the environment is already setup correctly.

@mkolesnik do you have any thoughts on this?

yxun: > He seems to be using the utility function envtest.InstallCRDs to install CRDs. 

Correct, there is an existing method for installing CRDs there so that we can make all steps working correctly. There are details if we choose to implement those steps by ourselves.

> a better fix would be to install the missing CRDs as part of make dev-env setup.

To be clear, the issue is the `hubclient` is missing those CRDs and the `hubclient` will call `get` or `list` CRDs later. the `hubclient` can run from a local golang client or a remote one depends on how the job is running.

> I feel, importing a heavy weight envtest package just to install the CRDs is an overkill.

The trade off is code simplicity vs importing weight package

>  The e2e test suite shouldn't need to bootstrap the cluster's CRDs, it should assume the environment is already setup correctly.

I would like to know how e2e tests cert-manager CRDs, I will follow that pattern if it can call `get` or `list` cert-manager CRDs from its hubclient.

.github/workflows/pr-e2e.yml
```diff
@@ -41,5 +41,8 @@ jobs:
       - name: Set up dev environment
         run: make dev-env
 
+      - name: Update test crds
```
mkolesnik: This isn't needed in a dev env

test/e2e/suite_test.go
mkolesnik: Yes as stated this is mixing up integ and e2e tests.

Integ tests bootstrap a fake env with `envtest` with a real API server in memory but no controllers other than our own, so no 3rd party CRDs get registered automatically.

E2E uses the dev env, which is a simulated "real environment" running several kind clusters with real K8s API servers and real 3rd party controllers.
It doesn't come with everything pre-installed, these have to be installed explicitly by us.

So, E2E needs everything setup beforehand by the `dev-env` target.

You can see PR #113 for reference

mkolesnik (changes requested):

test/e2e/suite_test.go
yxun: I see. I will follow PR #113 and update this PR with the `dev-env` target change.

yxun:
~The correct fix is in the commit [Add msa addon installation in dev-env](https://github.com/stolostron/multicluster-mesh-addon/pull/130/changes/96b3d80dafa2380e9e66146a9fdcfe01b5ef73da)
That's part of the https://github.com/stolostron/multicluster-mesh-addon/pull/130~

mkolesnik:
Not sure why you closed this, I think it was better to split off the setup from PR #130 seeing its as big as it is.

You should've resolved issues here and merged this, then you could merge main into #130

hack/dev-env.sh
```diff
@@ -297,6 +297,29 @@ join_clusters() {
     kubectl --kubeconfig="${hub_kubeconfig}" get managedclustersets
 }
 
+install_managed_serviceaccount() {
+    local hub_kubeconfig
```
mkolesnik: This function doesn't follow the patterns established by the other functions in this script:

1. No `require_clusters "${HUB}"` (manually checks kubeconfig instead)
2. No `on "${HUB}"` helper for kubectl (uses `kubectl --kubeconfig=` directly)

For reference, see `install_cert_manager`.

hack/dev-env.sh
```diff
@@ -297,6 +297,29 @@ join_clusters() {
...
+    if [[ ! -f "${hub_kubeconfig}" ]]; then
+        err "Hub kubeconfig not found at ${hub_kubeconfig}. Run 'make init-ocm' first."
+    fi
+
+    local ocm_repo_url
+    ocm_repo_url="https://open-cluster-management.io/helm-charts/"
+
+    log "Installing managed-serviceaccount addon on hub..."
+    ${HELM} repo add ocm "${ocm_repo_url}"
+    ${HELM} upgrade --install managed-serviceaccount ocm/managed-serviceaccount \
```
mkolesnik: No `--version` flag.
This pulls whatever version is currently latest in the chart repo.

Every other component in this script pins its version via a Makefile variable (`CERT_MANAGER_VERSION`, `OLM_VERSION`, `K8S_VERSION`).

Without a pin here:
- A breaking upstream release silently breaks `make dev-env` for everyone
- A supply chain compromise of the chart repo propagates automatically
- Builds are not reproducible

Add a `MSA_VERSION` variable in the Makefile and pass it here:

```suggestion
    ${HELM} upgrade --install managed-serviceaccount ocm/managed-serviceaccount \
        --version "${MSA_VERSION}" \
```

hack/dev-env.sh
```diff
@@ -297,6 +297,29 @@ join_clusters() {
...
+        --kubeconfig="${hub_kubeconfig}" \
+        --create-namespace \
+        --namespace open-cluster-management-addon \
+        --wait --timeout 180s
+
+    log "managed-serviceaccount addon installed on hub"
+    kubectl --kubeconfig="${hub_kubeconfig}" get managedclusteraddon -A | grep managed-serviceaccount || true
```
mkolesnik: This returns:
```
==> managed-serviceaccount addon installed on hub
cluster1    managed-serviceaccount                          False
cluster2    managed-serviceaccount                          False
```

Was this intentional?
Did you mean to wait for them to become available?

mkolesnik:
Please update the PR name and description to match the updated commit contents

hack/dev-env.sh
yxun: Reason of this change:
There was a failure in the job of the previous commits:
https://github.com/stolostron/multicluster-mesh-addon/actions/runs/28170953327/job/83434683995

```
/home/runner/work/multicluster-mesh-addon/multicluster-mesh-addon/hack/dev-env.sh: line 273: kubeconfig_for: command not found
```

I could never reproduce this error from my local run. Still not clear why the github workflow job was not able to find that function name. So I have to update this from copying the exact part of 

```
local hub_kubeconfig
hub_kubeconfig="$(kubeconfig_for "${HUB}")"
```

I can copy it back and let's see how the e2e job log works

hack/dev-env.sh
yxun: > Was this intentional?

Yes, those False are expected. The column name for that is `PROGRESSING` 

I can remove the grep part. The full result should be :

```
==> managed-serviceaccount addon installed on hub
NAMESPACE   NAME                     AVAILABLE   DEGRADED   PROGRESSING
cluster1    managed-serviceaccount                          False
cluster2    managed-serviceaccount                          False
```

yxun: The expected result:
https://github.com/open-cluster-management-io/managed-serviceaccount#verification

mkolesnik: Based on your example:
```
NAMESPACE        NAME                     AVAILABLE   DEGRADED   PROGRESSING
<your-cluster>   managed-serviceaccount   True        False      False
```

So it is expected to be available. which is not the case ATM.
This should wait for them to become available?

Also waiting is in line with other 3rd party install commands we have.

hack/dev-env.sh
mkolesnik: No this changed please follow the new pattern introduced in #152

hack/dev-env.sh
```diff
@@ -268,6 +268,36 @@ join_clusters() {
...
 
+install_managed_serviceaccount() {
+    require_clusters "${HUB}" "${CLUSTER1}" "${CLUSTER2}"
+    if on "${HUB}" kubectl get deployment managed-serviceaccount-addon-manager -n open-cluster-management-addon &>/dev/null; then
+        log "managed-serviceaccount addon already installed on hub, skipping"
+        return
+    fi
+
+    local ocm_repo_url
+    ocm_repo_url="https://open-cluster-management.io/helm-charts/"
```
mkolesnik: nit: This can be inlined, theres no need for a variable

hack/dev-env.sh
```diff
@@ -268,6 +268,36 @@ join_clusters() {
...
+        --create-namespace \
+        --namespace open-cluster-management-addon \
+        --wait --timeout 180s
+
+    log "Waiting for managed-serviceaccount addon to be ready..."
+    for cluster in "${CLUSTER1}" "${CLUSTER2}"; do
+        on "${HUB}" retry kubectl wait managedclusteraddon/managed-serviceaccount \
+            -n "${cluster}" \
+            --for='jsonpath={.status.conditions[?(@.type=="Available")].status}=True' \
+            --timeout=60s
```
mkolesnik: nit: `kubectl wait` has built-in condition syntax.
This also aligns with most other `kubectl wait` calls in the script

```suggestion
        on "${HUB}" retry kubectl wait managedclusteraddon/managed-serviceaccount \
            -n "${cluster}" --for=condition=Available --timeout=60s
```

mkolesnik (commented):
Couple of nits otherwise LGTM

mkolesnik (approved):

PR #158, opened 2026-06-23
Tune CodeRabbit for this repository

mkolesnik:
This tunes CodeRabbit for this repository so it produces less noisy feedback.

It adds a few repo-specific hints for Go, test, and shell files, disables docstring and generic test-quality checks that were generating unrelated comments, and keeps the sensitive-data check focused on actual secrets.

### DIFF
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -16,5 +16,39 @@ reviews:
         - Skip repetition — if a pattern repeats, mention it once at a summary level only.
         - Do not post comments on code style, formatting, or non-critical improvements.
         - Before flagging a deviation from best practices, check if the PR follows an existing pattern in the codebase. If the same approach is used elsewhere in the repo, it's intentional, not a bug.
+    - path: '**/*.go'
+      instructions: >
+        - This project uses klog (printf-style); do not suggest structured
+          logging just because fmt-style logging is used.
+        - Do not require copyright headers.
+    - path: '**/*_test.go'
+      instructions: >
+        - This project sets default Eventually/Consistently timeouts in suite
+          setup; do not require per-assertion timeouts unless a test is
+          genuinely ambiguous or flaky.
+        - Do not require failure messages on every Expect(); only flag them
+          when an assertion would otherwise be unclear.
+    - path: '**/*.sh'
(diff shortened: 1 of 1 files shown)


sridhargaddam (commented):
LGTM. I'll let @FilipB to take a look and approve the PR.

.coderabbit.yaml
```diff
@@ -16,5 +16,39 @@ reviews:
...
+          kubeconfig paths as sensitive data by themselves.
+  finishing_touches:
+    docstrings:
+      enabled: false
+  pre_merge_checks:
+    docstrings:
+      mode: off
+    custom_checks:
+      - name: Test Structure and Quality
+        mode: off
```
FilipB, line 45: Why disabling this? Those are just warnings in https://github.com/stolostron/coderabbit/blob/main/.coderabbit.yaml

.coderabbit.yaml
```diff
@@ -16,5 +16,39 @@ reviews:
...
+      - name: no-sensitive-data-in-logs
+        mode: error
+        instructions: |
```
FilipB, line 48: Just curious if the org wide config was causing false positives?

.coderabbit.yaml
mkolesnik: The org-wide check requires explicit timeouts on every Eventually/Consistently and failure messages on every Expect().
The path_instructions added in this same PR for test files say the opposite: skip per-assertion timeouts (we set defaults in suite setup) and only flag missing failure messages when an assertion would genuinely be unclear.

With the check at warning, it fires on things the path_instructions already say to ignore, which creates contradictory feedback.
Disabling the pre-merge check avoids that.
The path_instructions still guide the review comments themselves.

FilipB (approved):

.coderabbit.yaml
mkolesnik: Yeah, the org-wide version flags "internal hostnames" and "customer data" which can false-positive on cluster names and resource references that the controller logs normally.
This is not disabling the check, just narrowing the instructions to focus on actual secrets (passwords, tokens, API keys, private keys, credentials in URLs, etc.).

PR #161, opened 2026-06-23
Add make start/start-console local dev workflow for the console plugin

jmazzitelli:


jshaughn:
/lgtm

This should be useful for faster dev

PR #167, opened 2026-06-24
Add Konflux Tekton pipeline configurations for ACM 5.1

FilipB:
Add Pipelines-as-Code configurations for automated builds:
- Pull request pipeline: x86_64 only, 5-day image expiry
- Push pipeline: multi-arch (x86_64, ppc64le, s390x, arm64)

Both use Dockerfile.konflux and the stolostron-common-build-pipeline which includes hermetic builds, dependency prefetching, SAST scanning, and source image generation.


FilipB:
Depends on https://github.com/stolostron/multicluster-mesh-addon/pull/166

PR #170, opened 2026-06-24
POC: Replace operator ManifestWork with OperatorPolicy

mkolesnik:
Quick and dirty proof of concept to explore what operator installation via OperatorPolicy would look like.
Not production quality, just enough to validate the approach end-to-end on a dev environment.

Replaces per-cluster ManifestWorks for operator installation with the OCM governance policy framework's OperatorPolicy.
Instead of creating a ManifestWork per cluster containing OLM Subscription, OperatorGroup, and Namespace manifests, the controller now creates a single Policy + Placement + PlacementBinding that distributes an OperatorPolicy to all clusters in the ClusterSet.

Key changes:
- Controller creates Policy/Placement/PlacementBinding/ManagedClusterSetBinding instead of per-cluster ManifestWorks
- Status reads Policy compliance per-cluster instead of ManifestWork feedback for installedCSV
- Deletion sets OperatorPolicy to mustnothave and waits for compliance before cleaning up (last mesh only)
- Multi-tenancy: shared Policy uses addon-level labels only, non-last mesh deletion leaves the Policy intact
- Dev-env installs governance-policy-framework and enables OperatorPolicy on managed clusters
- Cacerts ManifestWork path unchanged

Limitations:
- Single operator package (sailoperator) for all platforms
- GW API conflict on OCP 4.19-4.21 documented as known limitation (fixed in 4.22+, see open-cluster-management-io/config-policy-controller#479)


hack/dev-env.sh
```diff
@@ -303,6 +303,41 @@ setup_mesh() {
...
+
+    log "Waiting for policy framework hub components..."
+    on "${HUB}" kubectl wait --for=condition=Available \
+        deployment/governance-policy-propagator \
+        deployment/governance-policy-addon-controller \
+        -n open-cluster-management --timeout=120s
+
+    for cluster in "${CLUSTER1}" "${CLUSTER2}"; do
+        log "Enabling governance-policy-framework on ${cluster}"
+        on "${HUB}" "${CLUSTERADM}" addon enable --names governance-policy-framework --clusters "${cluster}"
```
FilipB, line 320: What would be a workflow for RH customers who want to use our addon? They would have to manually enabled governance-policy-framework addon as prereq or it's enabled by default?

mkolesnik: From my research this should be on by default in ACM, with the rest or policy engine bits.
I didn't test it on OCP though so there might be surprises there..

PR #172, opened 2026-06-25
Add ClusterManagementAddOn resource and E2E test

FilipB:
Adds ClusterManagementAddOn resource to register the addon with OCM's addon-manager. The resource is deployed via Helm and declares this as a hub-only addon with manual installation strategy.

- Set installStrategy.type=Manual (no auto-deployment to clusters)
- Add E2E test to verify ClusterManagementAddOn is created with correct metadata

### DIFF
--- a/chart/templates/clustermanagementaddon.yaml
+++ b/chart/templates/clustermanagementaddon.yaml
@@ -1,0 +1,10 @@
+apiVersion: addon.open-cluster-management.io/v1beta1
+kind: ClusterManagementAddOn
+metadata:
+  name: multicluster-mesh-addon
+spec:
+  addOnMeta:
+    displayName: Multi-Cluster Mesh Add-on
+    description: "Hub-side controller for orchestrating multi-cluster Istio service mesh deployments"
+  installStrategy:
+    type: Manual
--- a/docs/design.md
+++ b/docs/design.md
@@ -31,6 +31,8 @@ The add-on follows OCM's hub-and-spoke model:
 - **Hub**: The Mesh Add-on controller watches `MultiClusterMesh` resources and creates [ManifestWorks][ManifestWork], orchestrates cert-manager and ManagedServiceAccount
 - **Spoke** (managed clusters): Receives ManifestWorks from the hub, runs the Sail/OSSM operator and Istio control plane
 
+A [ClusterManagementAddOn] resource is deployed to register this addon with OCM's addon manager, but the addon uses manual installation strategy and does not leverage the framework's lifecycle management features (auto-deployment, per-cluster enable/disable via `ManagedClusterAddOn`).
+
 ```mermaid
 flowchart TD
     subgraph Hub Cluster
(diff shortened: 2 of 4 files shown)


chart/templates/clustermanagementaddon.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: addon.open-cluster-management.io/v1beta1
+kind: ClusterManagementAddOn
```
sridhargaddam, line 2: Do we have to include the following annotation to indicate that this addon handles its own lifecycle?
Ref: https://open-cluster-management.io/docs/developer-guides/addon/

```
annotations:
  addon.open-cluster-management.io/lifecycle: "self"
```

chart/templates/clustermanagementaddon.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: addon.open-cluster-management.io/v1beta1
```
mkolesnik, line 1: docs/design.md line 113 says "Does not integrate with ACM addon lifecycle" and line 246 lists `ClusterManagementAddOn` as a potential future addition.
This PR adds the resource, so the design doc should be updated to reflect the current state.

chart/templates/clustermanagementaddon.yaml
FilipB: We are in a bit confusing situation as this annotation makes sense only for the add-on agent. And we are not using it. We are hub only addon. `self` should mean you take care of the add-on agent installation yourself and `addon-manager` means addon-manager will take care of it. But there is no add-on agent in our case so we should probably not include it at all. It will default to `addon-manager` but explicitly saying `self` is also incorrect and moreover it's deprecated.

chart/templates/clustermanagementaddon.yaml
FilipB: Updated the design doc

docs/design.md
mkolesnik, line 1: Line 115 now contradicts line 34. The addon *does* integrate with ACM addon lifecycle (via `ClusterManagementAddOn` registration), it just doesn't use lifecycle management features.

Suggestion:
```suggestion
- Does not use ACM addon lifecycle management (no auto-deployment, no per-cluster enable/disable via `ManagedClusterAddOn`)
```

FilipB: ` Does not integrate with ACM addon lifecycle (enable/disable via ManagedClusterAddOn and such)` does not contradict `A [ClusterManagementAddOn] resource is deployed to register this addon with OCM's addon manager, but the addon uses manual installation strategy and does not leverage the framework's lifecycle management features (auto-deployment, per-cluster enable/disable via ManagedClusterAddOn).`

ACM addon lifecycle is not used at all. Existence of `ClusterManagementAddOn` has no effect on addon lifecycle in our case, it's just there for visibility.

mkolesnik (commented):
/hold waiting for rereview from @sridhargaddam 
/lgtm

sridhargaddam (approved):

PR #173, opened 2026-06-28
Use dedicated operator namespace for all platforms

### LINKED ISSUE #124: Clean up OLM CSV when deleting operator ManifestWork on OCP
When targeting OCP clusters with the default namespace (`openshift-operators`), the operator ManifestWork only contains the Subscription since the namespace and OperatorGroup already exist.
When the ManifestWork is deleted, the work agent deletes the Subscription but OLM does not delete the CSV.
The CSV stays with the operator deployment still running.

If a new mesh is created for the same cluster, the new Subscription fails to resolve:
`constraints not satisfiable: @existing/<ns>//<csv> and <catalog>/<channel>/<csv> provide <API>`

When using a custom operator namespace, the MW includes the Namespace resource and its deletion cascades to everything inside it.

We need to either explicitly delete the CSV during operator cleanup, or change the OCP default to a dedicated namespace so the MW owns the full lifecycle.

Options:
- Change the default operator namespace to a dedicated one (e.g. `multicluster-mesh-operator`) for all platforms. This simplifies the CRD (one default instead of platform-specific defaults) and the controller code (no special-casing for `openshift-operators`). Tradeoff: on OCP, operators in `openshift-operators` appear in the console's "Installed Operators" view by default, a custom namespace might require an extra click.
- Delete the CSV as a separate cleanup step before deleting the operator ManifestWork.
- Add the CSV to the ManifestWork once its name is known (#119). Risk: work agent could fight OLM upgrades.

Related: #115 (simplify operator config)

Ref: https://olm.operatorframework.io/docs/tasks/uninstall-operator/

mkolesnik:
The default openshift-operators namespace already exists on OCP, so the
ManifestWork only contained the Subscription. Deleting it left the CSV
behind, blocking re-installation.

Use a dedicated namespace so the ManifestWork owns the full lifecycle.
Namespace deletion cascades to all operator resources including the CSV.

Fixes #124


docs/design.md
```diff
@@ -128,7 +128,7 @@ The resource name (`metadata.name`) is limited to 63 characters because it is us
 |-------|----------|-------------|
 | `spec.clusterSet` | Yes | Name of the [ManagedClusterSet] defining cluster membership (immutable after creation) |
 | `spec.controlPlane.namespace` | No | Namespace where Istio is installed on each cluster (default: `istio-system`) |
-| `spec.operator.namespace` | No | Namespace where the operator is installed (default: `openshift-operators` on OCP, `sail-operator` on K8s) |
+| `spec.operator.namespace` | No | Namespace where the operator is installed (default: `multicluster-mesh-operator`) |
```
sridhargaddam, line 134: The [OSSM docs](https://docs.redhat.com/en/documentation/red_hat_openshift_service_mesh/3.2/html-single/installing/index) explicitly include a note that the operator will be installed in the `openshift-operators` namespace by default.

So, every OCP admin who has ever installed OSSM will have the operator in `openshift-operators`. The addon deviating from this creates a parallel, non-standard installation that contradicts the product docs end users are trained to follow. Are we creating an inconsistent user experience with the broader OCP ecosystem?

docs/design.md
```diff
@@ -186,7 +186,7 @@ The add-on follows a **Do No Harm** strategy: it never forcibly uninstalls or do
 
 1. **Pre-existing operator detection**: The controller creates a [ManagedClusterView] to check if a Sail/OSSM Subscription already exists on the managed cluster. This is necessary because ManifestWork claims ownership of any resource it applies, and deleting the ManifestWork would remove a pre-existing Subscription, potentially disrupting other components that depend on it (e.g., OpenShift Gateway API).
```
sridhargaddam, line 191: Since there is a plan to support pre-existing operators as well in the controller, what would happen if the user already has an existing operator in the `openshift-operators` ns?

pkg/apis/mesh/v1alpha1/types.go
```diff
@@ -99,11 +99,13 @@ type ControlPlaneConfig struct {
...
-// OperatorConfig defines the Sail Operator installation settings
+// OperatorConfig defines the Mesh Operator installation settings
 type OperatorConfig struct {
-	// Namespace is the namespace where the Sail Operator will be installed
-	// Defaults to "openshift-operators" on OpenShift, "sail-operator" on vanilla Kubernetes
+	// Namespace is the namespace where the Mesh Operator will be installed.
+	// This namespace may be deleted when the mesh is removed, so avoid
+	// using a namespace that contains other resources.
 	// +optional
+	// +kubebuilder:default="multicluster-mesh-operator"
```
sridhargaddam, line 114: How would the controller handle the situation when users explicitly specify `openshift-operators`?

sridhargaddam (changes requested):
Are we fixing a symptom which can potentially create new problems, instead of addressing the root cause?

docs/design.md
mkolesnik: The OSSM docs describe manual installation by users.
This addon manages the full lifecycle automatically, so the user never interacts with the operator namespace directly.

Also AFAIK users aren't prohibited from using a custom ns when installing the operator on OCP.

This is the same pattern used by other OCM addons and operators that need clean uninstall. The addon creates a namespace it fully owns, so removing the mesh removes everything cleanly.

The field is overridable, so users who prefer openshift-operators can set it explicitly (but with the documented caveat about cleanup).

docs/design.md
mkolesnik: The design already covers this.
The Installation Workflow section (right above this line) describes three steps:

1. Pre-existing operator detection via ManagedClusterView checks if a Subscription already exists.
2. If a compatible Subscription is found, the addon skips ManifestWork creation entirely (adoption, no parallel install).
3. If no Subscription is found, the addon installs in the dedicated namespace.

So if a user already has OSSM in openshift-operators, the addon would still detect it and adopts it.
No second installation is created.
This PR does not change any of that logic.

pkg/apis/mesh/v1alpha1/types.go
mkolesnik: The namespace field is overridable.
If a user sets it to `openshift-operators` (or any other value), ManifestWork includes Namespace + OperatorGroup + Subscription for that namespace.

The old code had a hardcoded skip for the literal string `openshift-operators`, but it never actually checked whether the namespace exists on the spoke. It just assumed it did on OCP.
For any other pre-existing namespace, the old code had the same behavior as this PR.

If we wish to, we can add a spoke-side existence check (via MCV) before including the Namespace manifest, which would handle any pre-existing namespace correctly, not just `openshift-operators`.
That is out of scope for this PR.

I will add a CEL validation to exclude system ns (`openshift-*` or `kube-*`) to svae users from themselves.

mkolesnik:
> Are we fixing a symptom which can potentially create new problems, instead of addressing the root cause?

The root cause is that `openshift-operators` is a shared namespace that cannot be deleted for cleanup.
OLM (by design) does not remove CSVs when a Subscription is deleted in a namespace that still exists, so the CSV gets orphaned and blocks re-installation.
A dedicated namespace is the simplest fix, and it also ensures any other resources created by the mesh in that namespace are cleaned up on removal.

Installing AllNamespaces operators in a custom namespace is fully supported by OLM. You create a global OperatorGroup (empty spec) + Subscription in any namespace.
The RH OSSM docs show `openshift-operators` as the default for manual end-user installation, but there is no restriction against other namespaces.
AFAICT there's nothing OCP-specific beyond the OperatorGroup that openshift-operators provides.
OLM handles everything through the CSV regardless of which namespace the Subscription lives in.

There is actually an [open issue against the OCP ingress operator](https://github.com/openshift/cluster-ingress-operator/issues/1365) reporting that its automated `servicemeshoperator3` installation in `openshift-operators` causes upgrade conflicts with other operators in the same namespace.

I added CEL validation rejecting `openshift-*`, `kube-*`, and `default` namespaces so users cannot accidentally point at a system namespace.

chart/crds/mesh.open-cluster-management.io_multiclustermeshes.yaml
```diff
@@ -59,7 +59,7 @@ spec:
                     type: string
                 type: object
               operator:
-                description: Operator defines the Sail Operator installation configuration
+                description: Operator defines the Mesh Operator installation configuration
```
sridhargaddam: "Service Mesh Operator" would be more appropriate IMO instead of "Mesh Operator". If you agree, can you please make similar changes at other places too.

```suggestion
                description: Operator defines the Service Mesh Operator installation configuration
```

mkolesnik: Agree, Service Mesh Operator is more explicit.

chart/crds/mesh.open-cluster-management.io_multiclustermeshes.yaml
```diff
@@ -59,7 +59,8 @@ spec:
                     type: string
                 type: object
               operator:
-                description: Operator defines the Sail Operator installation configuration
+                description: Operator defines the Service Mesh Operator installation
```
sridhargaddam: Please update this line as well https://github.com/stolostron/multicluster-mesh-addon/blob/d1aaa7a5fcc1e23167359593dde8aaf4c52c916d/test/e2e/mesh_lifecycle_test.go#L123

mkolesnik: No, this is unrelated it's describing exactly what the test does - it updates the mesh spec operator channel..

```go
                Step("Updating mesh operator channel")
                Expect(getMesh(ctx, mesh)).To(Succeed())
                mesh.Spec.Operator.Channel = "candidate"
```

sridhargaddam: aah i see..

sridhargaddam (commented):
LGTM. 

@jewertow can you pls take a look at this PR?

@mkolesnik do we need any changes in the makefile/scripts that are used for deploying/validating the addon?

mkolesnik:
> @mkolesnik do we need any changes in the makefile/scripts that are used for deploying/validating the addon?

None that I can see, everything should've been handled by the PR

jewertow:
> @jewertow can you pls take a look at this PR?

Sorry, but I don't feel an expert in this area. @dgn, could you take a look?

dgn:
Honestly, I think we have to do both: 1. allow different namespace (and default to something other than `openshift-operators`) *and* 2. clean up the CSV ourselves. Let me explain:

1. As Mike has already pointed out, due to OLM's strange design decisions, update strategies of different operators in the same namespace can overwrite/influence each other. Because of that, it's always safer to use dedicated namespaces for different operators. To me that's a bug in OLM but somehow they don't treat it as such.

2. We still need to actively cleanup the CSV because the namespace might already exist anyway. No matter what name we choose as default (and IMO it should be selectable anyway), that namespace might already exist in the cluster and create the same problem we wanted to fix here.

mkolesnik:
Thanks for the review @dgn do you see any problem with it right now or can we push this in?

>     2. We still need to actively cleanup the CSV because the namespace might already exist anyway. No matter what name we choose as default (and IMO it should be selectable anyway), that namespace might already exist in the cluster and create the same problem we wanted to fix here.

Yes that's the plan for the future but right now for the MVP this is the easiest and simplest path.
Going forward, the `OperatorPolicy` might provide a solution to his (and more problems) but this is still being looked into.

dgn (approved):

sridhargaddam (approved):

PR #177, opened 2026-06-29
Make operator status reasons implementation-agnostic

mkolesnik:
Rename ReasonManifestWorkCreated to ReasonInstallationPending and
update its message to not reference ManifestWork internals.

### DIFF
--- a/pkg/apis/mesh/v1alpha1/types.go
+++ b/pkg/apis/mesh/v1alpha1/types.go
@@ -194,8 +194,8 @@ const (
 	// ReasonClustersNotReady indicates that not all clusters have confirmed operator installation
 	ReasonClustersNotReady = "ClustersNotReady"
 
-	// ReasonManifestWorkCreated indicates the operator ManifestWork has been created
-	ReasonManifestWorkCreated = "ManifestWorkCreated"
+	// ReasonInstallationPending indicates the operator installation has been requested
+	ReasonInstallationPending = "InstallationPending"
 
 	// ReasonOperatorInstalled indicates the operator CSV has been successfully installed
 	ReasonOperatorInstalled = "Installed"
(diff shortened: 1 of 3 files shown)


pkg/apis/mesh/v1alpha1/types.go
```diff
@@ -194,8 +194,8 @@ const (
 	// ReasonClustersNotReady indicates that not all clusters have confirmed operator installation
 	ReasonClustersNotReady = "ClustersNotReady"
 
-	// ReasonManifestWorkCreated indicates the operator ManifestWork has been created
-	ReasonManifestWorkCreated = "ManifestWorkCreated"
```
sridhargaddam, line 198: Looks like we will require some additional changes in the frontend code as well. For example https://github.com/stolostron/multicluster-mesh-addon/blob/e4188fe2dfd28d0c2b1b0943ad24643d1aedb139/frontend/src/components/MeshStatus.tsx#L11

Can you please modify the frontend code as well?

CC: @jmazzitelli (for awareness)

sridhargaddam (commented):
I like the idea to make the status reasons implementation-agnostic by not exposing the internal ManifestWork details to the users. One small observation regarding the missing frontend changes. Otherwise, LGTM.

pkg/apis/mesh/v1alpha1/types.go
mkolesnik: I discussed this with the FE team. 
They prefer a separate issue filed for this so they can handle it on their side. 
The frontend currently shows "Manifest Work Created" for the old reason, which will just show the raw string "InstallationPending" until they update the `friendlyReasons` map.

I filed #178 for this

sridhargaddam (approved):

PR #186, opened 2026-07-01
Add CodeRabbit instructions to catch backend-frontend drift

sridhargaddam:
This PR adds path_instructions for types.go and controller.go that tell CodeRabbit which frontend files depend on Go structs, constants, label keys, and defaults. Using the backend-frontend-consistency pre-merge check, codeRabbit can now flag PRs touching those Go files without corresponding frontend updates.

### DIFF
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -16,6 +16,68 @@ reviews:
         - Skip repetition — if a pattern repeats, mention it once at a summary level only.
         - Do not post comments on code style, formatting, or non-critical improvements.
         - Before flagging a deviation from best practices, check if the PR follows an existing pattern in the codebase. If the same approach is used elsewhere in the repo, it's intentional, not a bug.
+    - path: 'pkg/apis/mesh/v1alpha1/types.go'
+      instructions: >
+        This file defines the CRD types whose Go structs, constants, and
+        default values are mirrored by hand in the frontend console plugin.
+        When ANY of the following change, the PR MUST include matching
+        frontend updates — post a blocking comment if it does not.
+
+
+        Struct / field shape — `MultiClusterMeshSpec`, `MultiClusterMeshStatus`,
+        `ClusterMeshStatus` and their nested types are mirrored in
+        `frontend/src/types/multiClusterMesh.ts`. Adding, removing, renaming,
+        or re-typing a field here requires the same change there.
+
+
+        Condition type constants (`Condition*`) — `ConditionReady`,
+        `ConditionOperatorInstalled`, etc. are referenced as string literals
+        in `frontend/src/components/MeshDetailPage.tsx` (cluster
+        categorization, status display) and `MeshStatus.tsx` (default
+        condition type). A rename or removal breaks those look-ups.
+
+
+        Reason constants (`Reason*`) — every `Reason*` string value has a
+        corresponding key in the `friendlyReasons` map in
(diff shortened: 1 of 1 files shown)


mkolesnik (commented):
Reviewed together with Claude (opus-4-6). Three comments below.

.coderabbit.yaml
```diff
@@ -43,6 +105,30 @@ reviews:
     custom_checks:
       - name: Test Structure and Quality
         mode: off
+      - name: backend-frontend-consistency
+        mode: error
```
mkolesnik, line 109: Change `mode: error` to `mode: warn`. The FE team prefers to handle frontend changes on their own timeline, so blocking BE PRs is the wrong tool here. An informational comment gives visibility without gating the merge.

Also update the instructions to suggest opening a tracking issue with the `area/frontend` label (or checking if one already exists) instead of just flagging. The CodeRabbit sandbox has `gh` CLI access, so the instructions could say:

> Before flagging, check if a GitHub issue with `area/frontend` label already tracks this change. If not, suggest the PR author (or a maintainer) open one. Do not block the PR.

.coderabbit.yaml
```diff
@@ -43,6 +105,30 @@ reviews:
...
+        instructions: |
+          This repo has a Go backend (pkg/) and a TypeScript frontend
+          (frontend/) that share contract surfaces. When the PR changes any
+          of the Go files below WITHOUT also changing the corresponding
+          frontend files, flag it as an error:
+
+          pkg/apis/mesh/v1alpha1/types.go changes → must check:
+            - frontend/src/types/multiClusterMesh.ts (struct mirror)
+            - frontend/src/components/MeshStatus.tsx (friendlyReasons map)
```
mkolesnik, line 118: nit: The path_instructions (L33-37) correctly note MeshStatus.tsx uses condition type constants AND friendlyReasons, but here it only says "friendlyReasons map." Add "default condition type" to keep them in sync:

```suggestion
            - frontend/src/components/MeshStatus.tsx (friendlyReasons map,
              default condition type)
```

.coderabbit.yaml
```diff
@@ -16,6 +16,68 @@ reviews:
...
+        Default values — kubebuilder defaults like `istio-system`,
+        `stable`, `Automatic`, `Issuer`, `360h` have fallback display logic
+        in `frontend/src/components/MeshDetailPage.tsx`. Changing a default
+        here may require updating the frontend fallback.
+
+
+        Enum constraints — kubebuilder `+kubebuilder:validation:Enum` values
+        (e.g., `Automatic|Manual`, `Issuer|ClusterIssuer`) are mirrored as
+        TypeScript union types in `frontend/src/types/multiClusterMesh.ts`.
+    - path: 'pkg/hub/mesh/controller.go'
```
mkolesnik, line 56: The label keys and MW name constants in controller.go rarely change, so this will mostly fire on unrelated controller logic changes. The "only flag shared surfaces" clause should filter those, but it relies on AI judgment.

Consider a follow-up to move user-facing constants (label keys, MW names) from controller.go to types.go where the condition/reason constants already live. That would consolidate the contract surface into one file and reduce noise from this check.

sridhargaddam:
Closing this PR as there is a plan to remove the frontend code from the repo - https://github.com/stolostron/multicluster-mesh-addon/pull/237

PR #189, opened 2026-07-01
Add update existing ManagedServiceAccount method

yxun:
This PR adds a `ManagedServiceAccount` update method and integration test when a `MultiClusterMesh` is updated with custom TokenValidity .

It resolves the #120 

hold until merging #130


pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -19,8 +19,7 @@ func (r *Reconciler) ensureManagedServiceAccountCreated(ctx context.Context, mes
 	msaName := fmt.Sprintf("%s-%s-%s", mesh.Namespace, "istio-reader", mesh.Name)
 	existing := &msav1beta1.ManagedServiceAccount{}
 	if err := r.Get(ctx, key.Of(msaName, cluster.Name), existing); err == nil {
-		klog.V(4).Infof("Cluster %s already has ManagedServiceAccount %s, skipping", cluster.Name, msaName)
-		return nil
+		return r.ensureManagedServiceAccountUpdated(ctx, mesh, existing)
```
mkolesnik, line 23: The function now delegates to `ensureManagedServiceAccountUpdated` on the existing-resource path, but the name still says "Created".

A reader at the call site (controller.go L290) and the error message ("failed to create ManagedServiceAccounts" at L291) both reinforce the incorrect assumption that this function only creates.

Suggestion: rename the `ensureManagedServiceAccountCreated` func to `ensureManagedServiceAccount` and update the caller:
```go
if err := r.ensureManagedServiceAccount(ctx, mesh, &cluster); err != nil {
    return reconcile.Result{}, fmt.Errorf("failed to ensure ManagedServiceAccount on cluster %s: %w", cluster.Name, err)
}
```

pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -87,3 +86,22 @@ func (r *Reconciler) deleteAllManagedServiceAccounts(ctx context.Context, mesh *
 
 	return nil
 }
+
+// ensureManagedServiceAccountUpdated updates an existing ManagedServiceAccount with the mesh's spec.Security.Discovery.TokenValidity value
```
mkolesnik: nit: This comment is pointless, it just restates the function name and signature.
The function name `ensureManagedServiceAccountUpdated` plus the parameters already tell the reader what it does. 

```suggestion
```

pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -87,3 +86,22 @@ func (r *Reconciler) deleteAllManagedServiceAccounts(ctx context.Context, mesh *
...
+	}
+	existing.Spec.Rotation.Validity = *mesh.Spec.Security.Discovery.TokenValidity
+
+	if err := r.Update(ctx, existing); err != nil {
+		return fmt.Errorf("failed to update a ManagedServiceAccount %s/%s: %w", existing.Namespace, existing.Name, err)
+	}
+
+	klog.V(4).Infof("Successfully updated a ManagedServiceAccount %s/%s", existing.Namespace, existing.Name)
+	return nil
+}
```
mkolesnik, line 108: The update path only reconciles `Spec.Rotation.Validity` but not labels. ManifestWorks and Certificates both apply the full desired state on every reconcile (via WorkApplier and SSA respectively), so tampered labels self-heal. MSAs should do the same.

Reconcile labels alongside Validity:
```suggestion
func (r *Reconciler) ensureManagedServiceAccountUpdated(ctx context.Context, mesh *meshv1alpha1.MultiClusterMesh, existing *msav1beta1.ManagedServiceAccount) error {
	desiredLabels := meshOwnedLabels(mesh, existing.Namespace)
	desiredValidity := *mesh.Spec.Security.Discovery.TokenValidity

	if maps.Equal(existing.Labels, desiredLabels) && existing.Spec.Rotation.Validity == desiredValidity {
		return nil
	}
	existing.Labels = desiredLabels
	existing.Spec.Rotation.Validity = desiredValidity

	if err := r.Update(ctx, existing); err != nil {
		return fmt.Errorf("failed to update ManagedServiceAccount %s/%s: %w", existing.Namespace, existing.Name, err)
	}

	klog.V(4).Infof("Updated ManagedServiceAccount %s/%s", existing.Namespace, existing.Name)
	return nil
}
```
This also removes the ownership check since reconciling labels makes it redundant: if the MSA exists with the same deterministic name, we own it and restore our labels.

test/integration/controller_test.go
```diff
@@ -719,6 +715,23 @@ var _ = Describe("MultiClusterMesh Controller", func() {
...
+
+				msa := expectManagedServiceAccount(testNs, meshName, clusterName)
+				Expect(msa.Spec.Rotation.Validity).To(Equal(metav1.Duration{Duration: 360 * time.Hour}))
+
+				updateMesh(meshName, testNs, func(mesh *meshv1alpha1.MultiClusterMesh) {
+					mesh.Spec.Security.Discovery.TokenValidity = &metav1.Duration{Duration: 15 * time.Minute}
+				})
+				expectMeshNotReady(meshName, testNs)
+				msa = expectManagedServiceAccount(testNs, meshName, clusterName)
+				Expect(msa.Spec.Rotation.Validity).To(Equal(metav1.Duration{Duration: 15 * time.Minute}))
```
mkolesnik: `expectManagedServiceAccount` returns immediately here because the MSA already exists. 
The cache might not have the updated validity yet.

You can wrap both the Get and the validity assertion in a single `Eventually`:

```suggestion
			Eventually(func(g Gomega) {
				msa := &msav1beta1.ManagedServiceAccount{}
				g.Expect(k8sClient.Get(ctx, key.Of(expectedManagedServiceAccountName(testNs, meshName), clusterName), msa)).To(Succeed())
				g.Expect(msa.Spec.Rotation.Validity).To(Equal(metav1.Duration{Duration: 15 * time.Minute}))
			}).Should(Succeed())
```

test/integration/controller_test.go
```diff
@@ -719,6 +715,26 @@ var _ = Describe("MultiClusterMesh Controller", func() {
...
+				util.CreateManagedCluster(ctx, k8sClient, clusterName, testClusterSet)
+				util.CreateMultiClusterMesh(ctx, k8sClient, meshName, testNs, testClusterSet)
+
+				msa := expectManagedServiceAccount(testNs, meshName, clusterName)
+				Expect(msa.Spec.Rotation.Validity).To(Equal(metav1.Duration{Duration: 360 * time.Hour}))
+
+				updateMesh(meshName, testNs, func(mesh *meshv1alpha1.MultiClusterMesh) {
+					mesh.Spec.Security.Discovery.TokenValidity = &metav1.Duration{Duration: 15 * time.Minute}
+				})
+				expectMeshNotReady(meshName, testNs)
```
mkolesnik: This isn't really needed since you use `Eventually` to test the update on the MSA itself.

```suggestion
```

mkolesnik (approved):

PR #192, opened 2026-07-02
Handle missing discovery.tokenValidity in MCM spec

sridhargaddam:
The ensureManagedServiceAccountCreated function dereferences `mesh.Spec.Security.Discovery.TokenValidity `without checking whether the Discovery configuration is present. As a result, when a `MultiClusterMesh` resource omits the discovery section, such as in the basic.yaml manifest used by `make setup-mesh`, `TokenValidity` is nil, causing the controller to panic on every reconciliation attempt.

This PR applies the default value of 360h in code to align with the kubebuilder CRD default. This is necessary because CRD defaulting only applies to omitted leaf fields and does not populate defaults when the parent object itself is absent from the resource manifest.

### DIFF
--- a/pkg/hub/mesh/endpoint_discovery.go
+++ b/pkg/hub/mesh/endpoint_discovery.go
@@ -3,6 +3,7 @@ package mesh
 import (
 	"context"
 	"fmt"
+	"time"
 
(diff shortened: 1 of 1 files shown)


sridhargaddam:
Probably it was some setup issue, its working fine without this fix.

PR #193, opened 2026-07-02
Prepare full release process for a dev preview

FilipB:
Included changes:
- Adopted sail-operator pattern: VERSION in Makefile → TAG (0.1-latest), auto-propagates to Chart.yaml via make gen
- GitHub Actions workflow: manually triggered, builds/pushes image, creates git tag, publishes Helm chart to gh-pages, creates GitHub release
- Helm chart README with install instructions, release process guide


.github/workflows/release.yml
```diff
@@ -0,0 +1,106 @@
...
+          go-version-file: go.mod
+          cache: true
+
+      - name: Extract version from Chart.yaml
+        id: version
+        run: |
+          VERSION=$(sed -n 's/^version: *//p' chart/Chart.yaml)
+          echo "version=$VERSION" >> $GITHUB_OUTPUT
+          echo "tag=v$VERSION" >> $GITHUB_OUTPUT
+          echo "Release version: v$VERSION"
```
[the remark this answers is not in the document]
FilipB: We can expect valid version from the chart.yaml

FilipB:
If we agree on this approach, we still need to enable gh-pages and create quay secrets.

Makefile
```diff
@@ -14,8 +15,8 @@ $(BIN_DIR):
 	mkdir -p $(BIN_DIR)
 
 # Build ldflags for version injection
-LDFLAGS := -X github.com/stolostron/multicluster-mesh-addon/pkg/version.versionFromGit=$(GIT_VERSION)
-LDFLAGS += -X github.com/stolostron/multicluster-mesh-addon/pkg/version.commitFromGit=$(GIT_COMMIT)
+LDFLAGS := -X github.com/stolostron/multicluster-mesh-addon/pkg/version.buildVersion=$(VERSION)
```
mkolesnik, line 18: Maybe I'm missing something, but these ldflags target `buildVersion` and `buildGitRevision`, while `pkg/version/version.go` defines `versionFromGit` and `commitFromGit`.
The linker silently ignores the mismatch, so the binary has no version info.

Either update `pkg/version/version.go` to match:
```go
var (
	buildVersion     string
	buildGitRevision string
	buildDate        string
)
```
(and update `Get()` and `init()` to use the new names), or keep the ldflags matching the existing Go variable names.

FilipB: Thanks. Silly mistake

.github/workflows/release.yml
```diff
@@ -0,0 +1,94 @@
...
+      - name: Log in to Quay.io
+        uses: docker/login-action@9780b0c442fbb1117ed29e0efdff1e18412f7567 # v3.3.0
+        with:
+          registry: quay.io
+          username: ${{ secrets.QUAY_USERNAME }}
+          password: ${{ secrets.QUAY_PASSWORD }}
+
+      - name: Build and push image
+        run: |
+          make images push
```
sridhargaddam, line 57: The [Dockerfile](https://github.com/stolostron/multicluster-mesh-addon/blob/d915309883c4f34d26d8caf75fa26770253c5054/Dockerfile#L2) uses `registry.ci.openshift.org/stolostron/builder:go1.25-linux` as the builder stage which typically requires authentication and may not work with the github action which may not have the necessary access permissions. Can you please check this once.

.github/workflows/release.yml
```diff
@@ -0,0 +1,94 @@
...
+        run: |
+          git config user.name "$GITHUB_ACTOR"
+          git config user.email "$GITHUB_ACTOR@users.noreply.github.com"
+          git tag -a "v${{ steps.version.outputs.version }}" -m "Release v${{ steps.version.outputs.version }}"
+          git push origin "v${{ steps.version.outputs.version }}"
+
+      - name: Install Helm
+        uses: azure/setup-helm@b7246b12e77f7134dc2d460a3d5bad15bbe29390 # v4.1.0
+        with:
+          version: v3.16.3
```
sridhargaddam: The Makefile uses [HELM_VERSION v4.2.2](https://github.com/stolostron/multicluster-mesh-addon/blob/d915309883c4f34d26d8caf75fa26770253c5054/Makefile#L37) whereas the release process is using `v3.16.3` 

Should we pick one version and use it consistently?

.github/workflows/release.yml
```diff
@@ -0,0 +1,94 @@
...
+
+      - name: Run chart-releaser
+        uses: helm/chart-releaser-action@a917fd15b20e8b64b94d9158ad54cd6345335584 # v1.6.0
+        with:
+          charts_dir: .
```
sridhargaddam, line 69: The charts_dir is set to `.` here whereas in the cr.yaml file we have `charts-dir: chart`. Please check if they should be consistent.

.github/workflows/release.yml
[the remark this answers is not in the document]
sridhargaddam: I think we can add a validate version step to handle any misconfigurations. Something like...

```
  - name: Validate version
    run: |
      if ! echo "${{ steps.version.outputs.version }}" | grep -qE '^[0-9]+\.[0-9]+\.[0-9]+$'; then
        echo "Error: Version '${{ steps.version.outputs.version }}' is not valid semver"
        exit 1
      fi
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,94 @@
...
+      - name: Checkout code
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          ref: ${{ inputs.branch }}
+          fetch-depth: 0
+
+      - name: Extract version from Chart.yaml
+        id: version
+        run: |
+          VERSION=$(sed -n 's/^version: *//p' chart/Chart.yaml)
```
sridhargaddam, line 28: We have VERSION in the Makefile which is propagated to Chart.yaml via make gen-chart. But the release workflow reads version from Chart.yaml, not the Makefile. If someone updates the Makefile VERSION but forgets make gen, the release will use the stale Chart.yaml version.

The workflow should use either:
  - VERSION from the Makefile directly, or
  - add a validation that Makefile VERSION matches Chart.yaml version

sridhargaddam (commented):
Some initial comments from a quick pass

.github/workflows/release.yml
FilipB: `podman pull registry.ci.openshift.org/stolostron/builder:go1.25-linux` works without auth and VPN

.github/workflows/release.yml
FilipB: This is correct:
 charts_dir in release.yml (line 69): This is the helm/chart-releaser-action GitHub Action input. It specifies the parent directory to scan for charts. The action walks this directory looking for subdirectories
  that contain a Chart.yaml. With charts_dir: . (repo root), it will find chart/Chart.yaml.

  charts-dir in cr.yaml (line 3): This is a flag passed directly to the underlying cr (chart-releaser) CLI tool. It serves the same purpose — telling cr where to look for charts.
 
To avoid confusion, I removed charts-dir from cr.yaml as it's redundant.

.github/workflows/release.yml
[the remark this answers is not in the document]
FilipB: Added

.github/workflows/release.yml
FilipB: This is already covered by [ci/prow/verify](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/stolostron_multicluster-mesh-addon/193/pull-ci-stolostron-multicluster-mesh-addon-main-verify/2084291514405490688) it's not possible to merge a PR if version in Makefile and Chart.yaml are out of sync.

FilipB:
@sridhargaddam thank you for the review. Changes are pushed.

Makefile
```diff
@@ -1,8 +1,9 @@
 .DEFAULT_GOAL := default
 default: all
 
+VERSION ?= 0.1.0
```
sridhargaddam, line 4: Just curious: are there any plans to build and push daily images to quay (in a future PR)?

.github/workflows/release.yml
```diff
@@ -0,0 +1,93 @@
...
+          git push origin "v${{ steps.version.outputs.version }}"
+
+      - name: Log in to Quay.io
+        uses: docker/login-action@9780b0c442fbb1117ed29e0efdff1e18412f7567 # v3.3.0
+        with:
+          registry: quay.io
+          username: ${{ secrets.QUAY_USERNAME }}
+          password: ${{ secrets.QUAY_PASSWORD }}
+
+      - name: Build and push image
```
sridhargaddam, line 55: Do we want to use [PREVENT_IMAGE_OVERWRITE](https://github.com/istio-ecosystem/sail-operator/blob/c41b3a9c574f3b99c12a48cf27b9d4fbde275e74/.github/workflows/release.yaml#L51) similar to the Sail Operator repo?

.github/workflows/release.yml
```diff
@@ -0,0 +1,93 @@
...
+
+      # Tag push is atomic — fails if the tag already exists on the remote,
+      # we need to be sure not to overwrite existing images in quay which are using given tag.
+      # Must happen before chart-releaser, which scans remote tags to discover
+      # which chart versions to package and publish to the gh-pages branch.
+      - name: Create and push tag
+        run: |
+          git config user.name "$GITHUB_ACTOR"
+          git config user.email "$GITHUB_ACTOR@users.noreply.github.com"
+          git tag -a "v${{ steps.version.outputs.version }}" -m "Release v${{ steps.version.outputs.version }}"
```
sridhargaddam, line 45: Shall we use signed tags? (i.e., git tag -s ...)

sridhargaddam (commented):
Thanks @FilipB. Mostly good, please take a look at some minor comments/questions.

Makefile
FilipB: No plan yet. Do you know about any demands for daily images? I guess it only makes sense for external teams? We have daily images for sail operator and nobody is using them AFAIK.

.github/workflows/release.yml
FilipB: While it's not configurable here, the images will never be overwritten because the workflow will fail earlier on git push tags if the tag exists.

.github/workflows/release.yml
FilipB: It will not work with the current setup. We would have to import users GPG keys to the runner from secrets or use a bot to create a tag but we don't have any bot yet.

Makefile
sridhargaddam: Since we do it in Sail Operator, I wanted to hear your thoughts on it. Agree that we may not need it currently.

sridhargaddam:
LGTM @FilipB. Let's give others some time to take a look at the PR. If we don't hear anything by tomorrow, we can merge it.

.github/workflows/release.yml
```diff
@@ -0,0 +1,93 @@
...
+          VERSION=$(sed -n 's/^version: *//p' chart/Chart.yaml)
+          if ! printf '%s\n' "$VERSION" | grep -Eq '^[0-9]+\.[0-9]+\.[0-9]+(-[0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*)?$'; then
+            echo "Error: '$VERSION' is not a valid semver version"
+            exit 1
+          fi
+          echo "version=$VERSION" >> $GITHUB_OUTPUT
+          echo "tag=v$VERSION" >> $GITHUB_OUTPUT
+          echo "Release version: v$VERSION"
+
+      # Tag push is atomic — fails if the tag already exists on the remote,
```
mkolesnik: nit: em-dash.
```suggestion
      # Tag push is atomic - fails if the tag already exists on the remote,
```

RELEASE-PROCESS.md
```diff
@@ -0,0 +1,121 @@
...
+   - **Branch to release from**: Select the release branch (e.g., `release-0.2`)
+5. Click **Run workflow**
+
+The workflow will:
+1. Extract version from `chart/Chart.yaml`
+2. Check if tag already exists (fails if it does)
+3. Build and push container image: `quay.io/sail-dev/multicluster-mesh-addon:v0.2.0`
+4. Create and push git tag: `v0.2.0`
+5. Package and publish Helm chart to `gh-pages` branch
+6. Create GitHub release with auto-generated notes
```
mkolesnik: These steps don't match the actual workflow order. The tag is now pushed before the image build (not after), and there's no separate "check if tag exists" step. The push itself fails atomically if the tag exists.

```suggestion
1. Extract and validate version from `chart/Chart.yaml`
2. Create and push git tag: `v0.2.0` (fails if tag already exists)
3. Build and push container image: `quay.io/sail-dev/multicluster-mesh-addon:v0.2.0`
4. Package and publish Helm chart to `gh-pages` branch
5. Create GitHub release with auto-generated notes
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,93 @@
...
+      # Tag push is atomic - fails if the tag already exists on the remote,
+      # we need to be sure not to overwrite existing images in quay which are using given tag.
+      # Must happen before chart-releaser, which scans remote tags to discover
+      # which chart versions to package and publish to the gh-pages branch.
+      - name: Create and push tag
+        run: |
+          git config user.name "$GITHUB_ACTOR"
+          git config user.email "$GITHUB_ACTOR@users.noreply.github.com"
+          git tag -a "v${{ steps.version.outputs.version }}" -m "Release v${{ steps.version.outputs.version }}"
+          git push origin "v${{ steps.version.outputs.version }}"
```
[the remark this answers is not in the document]
FilipB: This scenario should be handled manually

FilipB:
@mkolesnik Thank you for review. We can merge now

PR #212, opened 2026-07-09
Fix log message when issuer is removed from the MCM CR

sridhargaddam:
Currently, when the spec.security.trust.certManager.issuerRef is removed
from the MCM CR, we see the following log message which is a bit misleading
as the cluster is still part of the clusterSet.

```
controller.go:428] Deleting Certificate mesh-system/cacerts-cluster1 (cluster cluster1 no longer in ClusterSet mesh-cluster-set) controller.go:428] Deleting Certificate mesh-system/cacerts-cluster2 (cluster cluster2 no longer in ClusterSet mesh-cluster-set)
```

This PR modifies the log message as follows.

When the issuer is removed, the following message is logged. 
```
Deleting Certificate mesh-system/cacerts-cluster1 (issuer configuration removed)
```

and when a cluster is actually removed from the ClusterSet, the following message is logged. 
```
Deleting Certificate mesh-system/cacerts-cluster1 (cluster cluster1 no longer in ClusterSet mesh-cluster-set)
```

### DIFF
--- a/pkg/hub/mesh/controller.go
+++ b/pkg/hub/mesh/controller.go
@@ -419,13 +419,17 @@ func (r *Reconciler) cleanupCertificates(ctx context.Context, mesh *meshv1alpha1
 		return fmt.Errorf("failed to list Certificates: %w", err)
 	}
 
+	deleteReason := "issuer configuration removed"
 	for _, cert := range certList.Items {
 		clusterName := cert.Labels[ClusterNameLabel]
 		if !forceCleanupAll && clusterNames[clusterName] {
 			continue
 		}
 
-		klog.Infof("Deleting Certificate %s/%s (cluster %s no longer in ClusterSet %s)", cert.Namespace, cert.Name, clusterName, mesh.Spec.ClusterSet)
+		if !forceCleanupAll {
+			deleteReason = fmt.Sprintf("cluster %s no longer in ClusterSet %s", clusterName, mesh.Spec.ClusterSet)
+		}
+		klog.Infof("Deleting Certificate %s/%s (%s)", cert.Namespace, cert.Name, deleteReason)
 		if err := client.IgnoreNotFound(r.Delete(ctx, &cert)); err != nil {
 			return fmt.Errorf("failed to delete Certificate %s/%s: %w", cert.Namespace, cert.Name, err)
 		}
--- a/pkg/hub/mesh/controller.go
+++ b/pkg/hub/mesh/controller.go
@@ -274,9 +274,14 @@ func (r *Reconciler) doReconcile(ctx context.Context, mesh *meshv1alpha1.MultiCl
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/hub/mesh/controller.go
```diff
@@ -419,13 +419,17 @@ func (r *Reconciler) cleanupCertificates(ctx context.Context, mesh *meshv1alpha1
...
 	for _, cert := range certList.Items {
 		clusterName := cert.Labels[ClusterNameLabel]
 		if !forceCleanupAll && clusterNames[clusterName] {
 			continue
 		}
 
-		klog.Infof("Deleting Certificate %s/%s (cluster %s no longer in ClusterSet %s)", cert.Namespace, cert.Name, clusterName, mesh.Spec.ClusterSet)
+		if !forceCleanupAll {
+			deleteReason = fmt.Sprintf("cluster %s no longer in ClusterSet %s", clusterName, mesh.Spec.ClusterSet)
+		}
```
mkolesnik: nit: `deleteReason` is only used inside the loop body. Defining it outside means the initial value leaks across iterations (harmless here since `forceCleanupAll` is constant, but the scoping is misleading). I'd move it inside:

```suggestion
	for _, cert := range certList.Items {
		clusterName := cert.Labels[ClusterNameLabel]
		if !forceCleanupAll && clusterNames[clusterName] {
			continue
		}

		deleteReason := "issuer configuration removed"
		if !forceCleanupAll {
			deleteReason = fmt.Sprintf("cluster %s no longer in ClusterSet %s", clusterName, mesh.Spec.ClusterSet)
		}
```

pkg/hub/mesh/controller.go
```diff
@@ -419,13 +419,17 @@ func (r *Reconciler) cleanupCertificates(ctx context.Context, mesh *meshv1alpha1
...
 
+	deleteReason := "issuer configuration removed"
 	for _, cert := range certList.Items {
 		clusterName := cert.Labels[ClusterNameLabel]
 		if !forceCleanupAll && clusterNames[clusterName] {
 			continue
 		}
 
-		klog.Infof("Deleting Certificate %s/%s (cluster %s no longer in ClusterSet %s)", cert.Namespace, cert.Name, clusterName, mesh.Spec.ClusterSet)
+		if !forceCleanupAll {
```
mkolesnik: The fix works for the log message, but the underlying design has `forceCleanupAll` doing double duty: controlling which certs to delete AND which reason to log. The caller already knows the reason (L277: `issuerRef.Name == ""`), so the function could take the reason as a parameter, or be split into two dedicated functions. 

Worth a follow-up PR to simplify.

sridhargaddam: Let me update it in the current PR itself.

mkolesnik (approved):

PR #213, opened 2026-07-09
Restructure MSA integration tests

mkolesnik:
The MSA tests duplicated mesh creation inside every It block instead of using BeforeEach for shared setup. Restructured to mirror the operator tests: BeforeEach establishes steady state, each It tests one aspect.

Added missing lifecycle tests for MSA cleanup on cluster deletion and ClusterSet deletion.

### DIFF
--- a/test/integration/controller_test.go
+++ b/test/integration/controller_test.go
@@ -663,31 +663,23 @@ var _ = Describe("MultiClusterMesh Controller", func() {
 	})
 
 	Context("Endpoint discovery", func() {
-		var cluster1, cluster2, cluster3 string
-
-		When("a MultiClusterMesh is created", func() {
+		When("referencing a set with a cluster", func() {
 			BeforeEach(func() {
-				cluster1 = util.UniqueName("cluster1")
-				cluster2 = util.UniqueName("cluster2")
-				util.CreateManagedCluster(ctx, k8sClient, cluster1, testClusterSet)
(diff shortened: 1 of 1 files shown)


test/integration/controller_test.go
```diff
@@ -696,41 +688,52 @@ var _ = Describe("MultiClusterMesh Controller", func() {
...
 			})
 
-			It("should create a ManagedServiceAccount after adding a cluster to the ClusterSet", func() {
+			It("should create ManagedServiceAccount for newly added cluster", func() {
 				util.CreateMultiClusterMesh(ctx, k8sClient, meshName, testNs, testClusterSet)
-				cluster3 = util.UniqueName("cluster3")
-				util.CreateManagedCluster(ctx, k8sClient, cluster3, testClusterSet)
-				expectManagedServiceAccount(testNs, meshName, cluster3)
-			})
+				expectManagedServiceAccount(testNs, meshName, clusterName)
```
yxun, line 697: nit: Is this expected check duplicated as the one in `It("should create ManagedServiceAccount with correct labels and default validity")` ?
The `"should create ManagedServiceAccount for newly added cluster"` test case result may focus on the newly added cluster `cluster2` .

yxun (approved):

test/integration/controller_test.go
mkolesnik: This was used as the sync point to see that the 1st MSA already exists, and the controller only handles the new cluster addition.

PR #215, opened 2026-07-13
ManifestWorkReplicaSet: Add remote secrets distribution change

### LINKED ISSUE #214: Implement remote access secret construction and ManifestWork distribution
This issue tracks the implementation of building an Istio remote API server access secret and distributing it via ManifestWork.
The secret need required ownership labels and Istio remote secret label and annotation. It should use the data from a ManagedServiceAccount secret.

The implementation should also include integration tests and e2e tests coverage.

yxun:
This PR builds remote cluster API server access secrets from each cluster's `ManagedServiceAccount` secret. And it creates a `ManifestWorkReplicaSet` for distributing those secrets.

The `ManifestWorkReplicaSet` feature is **Not** enabled by default. See ref:
https://open-cluster-management.io/docs/concepts/work-distribution/manifestworkreplicaset/

Enabling this feature is a prerequisite manual configuration step at the moment.

After setting up N-to-N remote access secrets distribution, users can create Istio control planes and Istio will pick up those secrets in the control plane namespace. All the RBAC requirements are handled by `istio-reader` ClusterRole and service account. Those are created by default when users create an Istio control plane.

This PR resolves #214 

WIP: updating e2e tests.

### DIFF
--- a/hack/dev-env.sh
+++ b/hack/dev-env.sh
@@ -188,6 +188,12 @@ init_ocm() {
     log "Waiting for OCM hub components to be ready..."
     on "${HUB}" retry kubectl wait --for=condition=Available \
         deployment/cluster-manager -n open-cluster-management --timeout=120s
+
+    log "Enabling feature ManifestWorkReplicaSet on cluster: ${HUB}"
+    on "${HUB}" kubectl patch clustermanager cluster-manager --type=merge \
+        -p '{"spec":{"workConfiguration":{"featureGates":[{"feature":"ManifestWorkReplicaSet","mode":"Enable"}]}}}'
+    on "${HUB}" retry kubectl wait --for=condition=Available \
+        deployment/cluster-manager -n open-cluster-management --timeout=120s
 }
 
 join_clusters() {
--- a/pkg/hub/mesh/controller.go
+++ b/pkg/hub/mesh/controller.go
@@ -145,7 +145,7 @@ func RegisterController(mgr manager.Manager) error {
 //+kubebuilder:rbac:groups=work.open-cluster-management.io,resources=manifestworks,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=cert-manager.io,resources=certificates,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=authentication.open-cluster-management.io,resources=managedserviceaccounts,verbs=get;list;watch;create;update;delete
-//+kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch;create;update;delete
+//+kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch;delete
 
 // Reconcile implements the reconcile loop for MultiClusterMesh resources
 func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
--- a/pkg/hub/mesh/controller.go
+++ b/pkg/hub/mesh/controller.go
@@ -46,8 +46,9 @@ const (
 	// When the operator targets this namespace, the controller skips creating Namespace and OperatorGroup.
 	BuiltinOperatorNamespace = "openshift-operators"
 
-	OperatorManifestWorkName = "multicluster-mesh-operator"
-	ManifestWorkNameCacerts  = "multicluster-mesh-cacerts"
+	OperatorManifestWorkName      = "multicluster-mesh-operator"
+	ManifestWorkNameCacerts       = "multicluster-mesh-cacerts"
+	ManifestWorkNameRemoteSecrets = "multicluster-mesh-remote-secrets"
 
 	FeedbackInstalledCSV = "installedCSV"
 
@@ -144,7 +145,7 @@ func RegisterController(mgr manager.Manager) error {
 //+kubebuilder:rbac:groups=work.open-cluster-management.io,resources=manifestworks,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=cert-manager.io,resources=certificates,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=authentication.open-cluster-management.io,resources=managedserviceaccounts,verbs=get;list;watch;create;update;delete
-//+kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch
+//+kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch;create;update;delete
 
 // Reconcile implements the reconcile loop for MultiClusterMesh resources
 func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
@@ -273,6 +274,10 @@ func (r *Reconciler) doReconcile(ctx context.Context, mesh *meshv1alpha1.MultiCl
 		}
 	}
 
+	if err := r.ensureRemoteSecretDistributed(ctx, mesh, clusters); err != nil {
+		return fmt.Errorf("failed to ensure remote secret distribution for mesh %s: %w", mesh.Name, err)
+	}
+
 	forceCleanupAll := mesh.Spec.Security.Trust.CertManager.IssuerRef.Name == ""
 	if err := r.cleanupCertificates(ctx, mesh, clusters, forceCleanupAll); err != nil {
 		return fmt.Errorf("failed to cleanup Certificates: %w", err)
@@ -282,9 +287,12 @@ func (r *Reconciler) doReconcile(ctx context.Context, mesh *meshv1alpha1.MultiCl
 		return fmt.Errorf("failed to cleanup ManifestWorks: %w", err)
 	}
 
-	// Cleanup ManagedServiceAccount when the cluster(s) are removed from the ClusterSet.
 	if err := r.cleanupManagedServiceAccounts(ctx, mesh, clusters); err != nil {
 		return fmt.Errorf("failed to cleanup ManagedServiceAccounts: %w", err)
+	}
+
+	if err := r.cleanupRemoteSecrets(ctx, mesh, clusters); err != nil {
+		return fmt.Errorf("failed to cleanup Istio remote secrets: %w", err)
 	}
 
 	return nil
@@ -363,6 +371,10 @@ func (r *Reconciler) handleDeletion(ctx context.Context, mesh *meshv1alpha1.Mult
 
 	if err := r.deleteAllManagedServiceAccounts(ctx, mesh); err != nil {
 		return fmt.Errorf("failed to cleanup ManagedServiceAccount resources: %w", err)
+	}
+
+	if err := r.deleteAllRemoteSecrets(ctx, mesh); err != nil {
+		return fmt.Errorf("failed to cleanup Istio remote access secrets: %w", err)
 	}
 
 	// Trigger reconciliation for other meshes targeting the same cluster set.
@@ -724,7 +736,6 @@ func (r *Reconciler) ensureCacertsManifestWork(ctx context.Context, mesh *meshv1
 	secretName := getCacertsName(cluster.Name)
 	secret := &corev1.Secret{}
 	err := r.Get(ctx, key.Of(secretName, mesh.Namespace), secret)
-
 	if err != nil {
 		if apierrors.IsNotFound(err) {
 			klog.V(4).Infof("Secret %s/%s not found yet, waiting for cert-manager to create it", mesh.Namespace, secretName)
--- a/pkg/hub/mesh/controller.go
+++ b/pkg/hub/mesh/controller.go
@@ -42,10 +42,10 @@ import (
 )
 
 const (
-	OperatorManifestWorkName      = "multicluster-mesh-operator"
-	ManifestWorkNameCacerts       = "multicluster-mesh-cacerts"
-	ManifestWorkReplicaSetName    = "multicluster-mesh-mwrset"
-	ManifestWorkNameCPNSPrefix    = "multicluster-mesh-cp-ns-"
+	OperatorManifestWorkName   = "multicluster-mesh-operator"
+	ManifestWorkNameCacerts    = "multicluster-mesh-cacerts"
+	ManifestWorkReplicaSetName = "multicluster-mesh-mwrset"
+	ManifestWorkNameCPNSPrefix = "multicluster-mesh-cp-ns-"
 
 	FeedbackInstalledCSV = "installedCSV"
 
@@ -311,10 +311,6 @@ func (r *Reconciler) doReconcile(ctx context.Context, mesh *meshv1alpha1.MultiCl
 
 	if err := r.ensureManifestWorkReplicaSet(ctx, mesh); err != nil {
 		return fmt.Errorf("failed to ensure ManifestWorkReplicaSet for mesh %s: %w", mesh.Name, err)
-	}
-
-	if err := r.cleanupRemoteSecrets(ctx, mesh, clusters); err != nil {
-		return fmt.Errorf("failed to cleanup Istio remote secrets: %w", err)
 	}
 
 	return nil
--- a/pkg/hub/mesh/controller.go
+++ b/pkg/hub/mesh/controller.go
@@ -141,7 +141,7 @@ func RegisterController(mgr manager.Manager) error {
 //+kubebuilder:rbac:groups=cluster.open-cluster-management.io,resources=managedclusters,verbs=get;list;watch
 //+kubebuilder:rbac:groups=cluster.open-cluster-management.io,resources=managedclustersets,verbs=get;list;watch
 //+kubebuilder:rbac:groups=work.open-cluster-management.io,resources=manifestworks,verbs=get;list;watch;create;update;patch;delete
-//+kubebuilder:rbac:groups=work.open-cluster-management.io,resources=manifestworkreplicasets,verbs=get;list;create;update;delete
+//+kubebuilder:rbac:groups=work.open-cluster-management.io,resources=manifestworkreplicasets,verbs=get;list;watch;create;update;delete
 //+kubebuilder:rbac:groups=cert-manager.io,resources=certificates,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=authentication.open-cluster-management.io,resources=managedserviceaccounts,verbs=get;list;watch;create;update;delete
 //+kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch;delete
--- a/pkg/hub/mesh/controller.go
+++ b/pkg/hub/mesh/controller.go
@@ -327,7 +327,11 @@ func (r *Reconciler) doReconcile(ctx context.Context, mesh *meshv1alpha1.MultiCl
 	}
 
 	if err := r.ensureManifestWorkReplicaSet(ctx, mesh); err != nil {
-		return fmt.Errorf("failed to ensure ManifestWorkReplicaSet for mesh %s/%s: %w", mesh.Namespace, mesh.Name, err)
+		if apierrors.IsNotFound(err) {
+			klog.V(4).Infof("managedServiceAccount secret not found yet, waiting for ManagedServiceAccount controller to create it: %s", err)
+		} else {
+			return fmt.Errorf("failed to ensure ManifestWorkReplicaSet for mesh %s/%s: %w", mesh.Namespace, mesh.Name, err)
+		}
 	}
 
 	return nil
--- a/pkg/hub/mesh/controller.go
+++ b/pkg/hub/mesh/controller.go
@@ -45,7 +45,6 @@ import (
 const (
 	OperatorManifestWorkName   = "multicluster-mesh-operator"
 	ManifestWorkNameCacerts    = "multicluster-mesh-cacerts"
-	ManifestWorkReplicaSetName = "multicluster-mesh-mwrset"
 	ManifestWorkNameCPNSPrefix = "multicluster-mesh-cp-ns-"
 
 	FeedbackInstalledCSV = "installedCSV"
@@ -145,7 +144,7 @@ func RegisterController(mgr manager.Manager) error {
 //+kubebuilder:rbac:groups=cluster.open-cluster-management.io,resources=managedclustersets/bind,verbs=create
 //+kubebuilder:rbac:groups=cluster.open-cluster-management.io,resources=placements,verbs=get;list;watch;create;update
 //+kubebuilder:rbac:groups=work.open-cluster-management.io,resources=manifestworks,verbs=get;list;watch;create;update;patch;delete
-//+kubebuilder:rbac:groups=work.open-cluster-management.io,resources=manifestworkreplicasets,verbs=get;list;watch;create;update;delete
+//+kubebuilder:rbac:groups=work.open-cluster-management.io,resources=manifestworkreplicasets,verbs=get;list;watch;create;update
 //+kubebuilder:rbac:groups=cert-manager.io,resources=certificates,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=authentication.open-cluster-management.io,resources=managedserviceaccounts,verbs=get;list;watch;create;update;delete
 //+kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch;delete
@@ -328,7 +327,7 @@ func (r *Reconciler) doReconcile(ctx context.Context, mesh *meshv1alpha1.MultiCl
 	}
 
 	if err := r.ensureManifestWorkReplicaSet(ctx, mesh); err != nil {
-		return fmt.Errorf("failed to ensure ManifestWorkReplicaSet for mesh %s: %w", mesh.Name, err)
+		return fmt.Errorf("failed to ensure ManifestWorkReplicaSet for mesh %s/%s: %w", mesh.Namespace, mesh.Name, err)
 	}
 
 	return nil
--- a/pkg/hub/mesh/controller.go
+++ b/pkg/hub/mesh/controller.go
@@ -147,7 +147,7 @@ func RegisterController(mgr manager.Manager) error {
 //+kubebuilder:rbac:groups=work.open-cluster-management.io,resources=manifestworkreplicasets,verbs=get;list;watch;create;update
 //+kubebuilder:rbac:groups=cert-manager.io,resources=certificates,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=authentication.open-cluster-management.io,resources=managedserviceaccounts,verbs=get;list;watch;create;update;delete
-//+kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch;delete
+//+kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch
 
 // Reconcile implements the reconcile loop for MultiClusterMesh resources
 func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
@@ -415,10 +415,6 @@ func (r *Reconciler) handleDeletion(ctx context.Context, mesh *meshv1alpha1.Mult
 
 	if err := r.cleanupManagedClusterSetBinding(ctx, mesh); err != nil {
 		return fmt.Errorf("failed to cleanup ManagedClusterSetBinding: %w", err)
-	}
-
-	if err := r.deleteAllRemoteSecrets(ctx, mesh); err != nil {
-		return fmt.Errorf("failed to cleanup Istio remote access secrets: %w", err)
 	}
 
 	// Trigger reconciliation for other meshes targeting the same cluster set.
--- a/pkg/hub/mesh/controller.go
+++ b/pkg/hub/mesh/controller.go
@@ -45,7 +45,6 @@ const (
 	OperatorManifestWorkName      = "multicluster-mesh-operator"
 	ManifestWorkNameCacerts       = "multicluster-mesh-cacerts"
 	ManifestWorkReplicaSetName    = "multicluster-mesh-mwrset"
-	ManifestWorkNameRemoteSecrets = "multicluster-mesh-remote-secrets"
 	ManifestWorkNameCPNSPrefix    = "multicluster-mesh-cp-ns-"
 
 	FeedbackInstalledCSV = "installedCSV"
@@ -288,10 +287,6 @@ func (r *Reconciler) doReconcile(ctx context.Context, mesh *meshv1alpha1.MultiCl
 		}
 	}
 
-	if err := r.ensureRemoteSecretDistributed(ctx, mesh, clusters); err != nil {
-		return fmt.Errorf("failed to ensure remote secret distribution for mesh %s: %w", mesh.Name, err)
-	}
-
 	if mesh.Spec.Security.Trust.CertManager.IssuerRef.Name == "" {
 		if err := r.deleteAllCertificates(ctx, mesh); err != nil {
 			return fmt.Errorf("failed to cleanup Certificates: %w", err)
@@ -312,6 +307,10 @@ func (r *Reconciler) doReconcile(ctx context.Context, mesh *meshv1alpha1.MultiCl
 
 	if err := r.cleanupManagedServiceAccounts(ctx, mesh, clusters); err != nil {
 		return fmt.Errorf("failed to cleanup ManagedServiceAccounts: %w", err)
+	}
+
+	if err := r.ensureManifestWorkReplicaSet(ctx, mesh); err != nil {
+		return fmt.Errorf("failed to ensure ManifestWorkReplicaSet for mesh %s: %w", mesh.Name, err)
 	}
 
 	if err := r.cleanupRemoteSecrets(ctx, mesh, clusters); err != nil {
--- a/pkg/hub/mesh/controller.go
+++ b/pkg/hub/mesh/controller.go
@@ -42,9 +42,11 @@ import (
 )
 
 const (
-	OperatorManifestWorkName   = "multicluster-mesh-operator"
-	ManifestWorkNameCacerts    = "multicluster-mesh-cacerts"
-	ManifestWorkNameCPNSPrefix = "multicluster-mesh-cp-ns-"
+	OperatorManifestWorkName      = "multicluster-mesh-operator"
+	ManifestWorkNameCacerts       = "multicluster-mesh-cacerts"
+	ManifestWorkReplicaSetName    = "multicluster-mesh-mwrset"
+	ManifestWorkNameRemoteSecrets = "multicluster-mesh-remote-secrets"
+	ManifestWorkNameCPNSPrefix    = "multicluster-mesh-cp-ns-"
 
 	FeedbackInstalledCSV = "installedCSV"
 
@@ -140,6 +142,7 @@ func RegisterController(mgr manager.Manager) error {
 //+kubebuilder:rbac:groups=cluster.open-cluster-management.io,resources=managedclusters,verbs=get;list;watch
 //+kubebuilder:rbac:groups=cluster.open-cluster-management.io,resources=managedclustersets,verbs=get;list;watch
 //+kubebuilder:rbac:groups=work.open-cluster-management.io,resources=manifestworks,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=work.open-cluster-management.io,resources=manifestworkreplicasets,verbs=get;list;create;update;delete
 //+kubebuilder:rbac:groups=cert-manager.io,resources=certificates,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=authentication.open-cluster-management.io,resources=managedserviceaccounts,verbs=get;list;watch;create;update;delete
 //+kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch;delete
@@ -283,6 +286,10 @@ func (r *Reconciler) doReconcile(ctx context.Context, mesh *meshv1alpha1.MultiCl
 				return fmt.Errorf("failed to ensure cacerts ManifestWork for cluster %s: %w", cluster.Name, err)
 			}
 		}
+	}
+
+	if err := r.ensureRemoteSecretDistributed(ctx, mesh, clusters); err != nil {
+		return fmt.Errorf("failed to ensure remote secret distribution for mesh %s: %w", mesh.Name, err)
 	}
 
 	if mesh.Spec.Security.Trust.CertManager.IssuerRef.Name == "" {
@@ -795,6 +802,7 @@ func (r *Reconciler) ensureCacertsManifestWork(ctx context.Context, mesh *meshv1
 	secretName := getCacertsName(cluster.Name)
 	secret := &corev1.Secret{}
 	err := r.Get(ctx, key.Of(secretName, mesh.Namespace), secret)
+
 	if err != nil {
 		if apierrors.IsNotFound(err) {
 			klog.V(4).Infof("Secret %s/%s not found yet, waiting for cert-manager to create it", mesh.Namespace, secretName)
--- a/pkg/hub/mesh/endpoint_discovery.go
+++ b/pkg/hub/mesh/endpoint_discovery.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"maps"
+	"sort"
 
 	meshv1alpha1 "github.com/stolostron/multicluster-mesh-addon/pkg/apis/mesh/v1alpha1"
 	"github.com/stolostron/multicluster-mesh-addon/pkg/key"
@@ -122,18 +123,20 @@ func (r *Reconciler) ensureRemoteSecretDistributed(ctx context.Context, mesh *me
 		if err != nil {
 			if errors.IsNotFound(err) {
 				klog.V(4).Infof("ManagedServiceAccount secret %s/%s not found yet, waiting for its controller to create it", cluster.Name, msaSecretName)
-				return nil
 			}
+			msaSecret = nil
 		}
 
-		remoteSecret := buildMeshRemoteSecret(mesh, &cluster, msaSecret)
-		msaSecrets[remoteSecret.Name] = remoteSecret
+		if msaSecret != nil {
+			remoteSecret := buildMeshRemoteSecret(mesh, &cluster, msaSecret)
+			msaSecrets[remoteSecret.Name] = remoteSecret
+		}
 	}
 
 	for _, cluster := range clusters {
 		work, err := r.workApplier.Apply(ctx, r.buildRemoteSecretManifestWork(mesh, &cluster, msaSecrets))
 		if err != nil {
-			return fmt.Errorf("failed to apply remote secret ManifestWork %s/%s", work.Namespace, work.Name)
+			return fmt.Errorf("failed to apply remote secret ManifestWork %s/%s: %w", work.Namespace, work.Name, err)
 		}
 	}
 
@@ -167,11 +170,16 @@ func buildMeshRemoteSecret(mesh *meshv1alpha1.MultiClusterMesh, cluster *cluster
 
 // buildRemoteSecretManifestWork builds a ManifestWork using remote access secrets from other clusters.
 func (r *Reconciler) buildRemoteSecretManifestWork(mesh *meshv1alpha1.MultiClusterMesh, cluster *clusterv1.ManagedCluster, remoteSecrets map[string]*corev1.Secret) *workv1.ManifestWork {
-	manifests := []workv1.Manifest{}
+	names := make([]string, 0, len(remoteSecrets))
+	for name := range remoteSecrets {
+		names = append(names, name)
+	}
+	sort.Strings(names)
 
-	for _, secret := range remoteSecrets {
+	manifests := []workv1.Manifest{}
+	for _, name := range names {
 		manifests = append(manifests, workv1.Manifest{
-			RawExtension: runtime.RawExtension{Object: secret},
+			RawExtension: runtime.RawExtension{Object: remoteSecrets[name]},
 		})
 	}
 
--- a/pkg/hub/mesh/endpoint_discovery.go
+++ b/pkg/hub/mesh/endpoint_discovery.go
@@ -134,9 +134,9 @@ func (r *Reconciler) ensureRemoteSecretDistributed(ctx context.Context, mesh *me
 	}
 
 	for _, cluster := range clusters {
-		work, err := r.workApplier.Apply(ctx, r.buildRemoteSecretManifestWork(mesh, &cluster, msaSecrets))
+		_, err := r.workApplier.Apply(ctx, r.buildRemoteSecretManifestWork(mesh, &cluster, msaSecrets))
 		if err != nil {
-			return fmt.Errorf("failed to apply remote secret ManifestWork %s/%s: %w", work.Namespace, work.Name, err)
+			return fmt.Errorf("failed to apply remote secret ManifestWork for cluster %s: %w", cluster.Name, err)
 		}
 	}
 
--- a/pkg/hub/mesh/endpoint_discovery.go
+++ b/pkg/hub/mesh/endpoint_discovery.go
@@ -7,10 +7,13 @@ import (
 
 	meshv1alpha1 "github.com/stolostron/multicluster-mesh-addon/pkg/apis/mesh/v1alpha1"
 	"github.com/stolostron/multicluster-mesh-addon/pkg/key"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/klog/v2"
 	clusterv1 "open-cluster-management.io/api/cluster/v1"
+	workv1 "open-cluster-management.io/api/work/v1"
 	msav1beta1 "open-cluster-management.io/managed-serviceaccount/apis/authentication/v1beta1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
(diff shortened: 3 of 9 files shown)


pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -106,3 +112,113 @@ func (r *Reconciler) ensureManagedServiceAccountUpdated(ctx context.Context, mes
...
+			Name:      istioRemoteSecretName,
+			Namespace: mesh.GetControlPlaneNamespace(),
+			Labels:    istioRemoteSecretLabels,
+			Annotations: map[string]string{
+				"networking.istio.io/cluster": cluster.Name,
+			},
+			OwnerReferences: msaSecret.OwnerReferences,
+		},
+		Type: corev1.SecretTypeOpaque,
+		Data: msaSecret.Data,
```
mkolesnik: `msaSecret.Data` has `ca.crt` and `token` keys. Istio's multicluster secret controller expects a single data key named after the cluster with a kubeconfig YAML as the value:
```yaml
data:
  <cluster-name>: <base64-kubeconfig>
```
The kubeconfig needs to be constructed from the CA cert, token, and the spoke's API server URL (from `ManagedCluster.spec.managedClusterClientConfigs[0].url` or similar).

Verified on dev-env: spoke secrets have `ca.crt`/`token` keys. Istio won't recognize this format, so remote endpoint discovery won't work.

pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -106,3 +112,113 @@ func (r *Reconciler) ensureManagedServiceAccountUpdated(ctx context.Context, mes
...
+			Kind:       "Secret",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      istioRemoteSecretName,
+			Namespace: mesh.GetControlPlaneNamespace(),
+			Labels:    istioRemoteSecretLabels,
+			Annotations: map[string]string{
+				"networking.istio.io/cluster": cluster.Name,
+			},
+			OwnerReferences: msaSecret.OwnerReferences,
```
mkolesnik: `msaSecret.OwnerReferences` points to the ManagedServiceAccount on the hub (by UID).
When this secret is distributed to spokes via MWRS, those UIDs don't exist on the spoke cluster.

The spoke's GC may delete the secret as an orphan, and a `controller: true` OwnerReference conflicts with the work agent's own controller ref.

Drop this field. Spoke-side ownership is managed by the work agent.

```suggestion
```

pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -106,3 +112,113 @@ func (r *Reconciler) ensureManagedServiceAccountUpdated(ctx context.Context, mes
 	klog.V(4).Infof("Updated ManagedServiceAccount %s/%s", existing.Namespace, existing.Name)
 	return nil
 }
+
+// ensureManifestWorkReplicaSet creates a ManifestWorkReplicaSet to distribute remote access secrets for clusters selected by a Placement
+func (r *Reconciler) ensureManifestWorkReplicaSet(ctx context.Context, mesh *meshv1alpha1.MultiClusterMesh) error {
```
mkolesnik: The dev-env setup (`hack/dev-env.sh` `init_ocm` action) doesn't enable the ManifestWorkReplicaSet feature gate.
The MWRS CRD exists on the hub after `clusteradm init`, but the work-controller that processes MWRS resources only starts when the feature is explicitly enabled.

Without it, this function creates the MWRS successfully, but no ManifestWorks are generated from it and no remote secrets reach the spokes.

`hack/dev-env.sh` needs to patch the ClusterManager after init:
```bash
kubectl patch clustermanager cluster-manager --type=merge \
  -p '{"spec":{"workConfiguration":{"featureGates":[{"feature":"ManifestWorkReplicaSet","mode":"Enable"}]}}}'
```

pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -106,3 +112,113 @@ func (r *Reconciler) ensureManagedServiceAccountUpdated(ctx context.Context, mes
...
+	placement := &clusterv1beta1.Placement{}
+	if err := r.Get(ctx, key.Of(mesh.Name, mesh.Namespace), placement); err != nil {
+		return fmt.Errorf("failed to get Placement %s/%s: %w", mesh.Namespace, mesh.Name, err)
+	}
+
+	workTemplate, err := r.buildManifestWorkSpec(ctx, mesh)
```
mkolesnik: The Placement name is always `mesh.Name` (set in `ensurePlacement`).
This Get fetches it just to use `placement.Name`, which is `mesh.Name`.

Drop the Get and use `mesh.Name` directly.
This also avoids errors when the Placement doesn't exist yet (e.g. when the ClusterSet is missing and `ensurePlacement` was skipped).

Verified on dev-env: an MWRS referencing a non-existent Placement sets `PlacementVerified: False` with reason `PlacementDecisionNotFound` and waits.
When the Placement is created later, the MWRS controller picks it up immediately and distributes to all selected clusters.

```suggestion
	workTemplate, err := r.buildManifestWorkSpec(ctx, mesh)
```

And on line 139:
```
mwrset.Spec.PlacementRefs = []workv1alpha1.LocalPlacementReference{{Name: mesh.Name}}
```

pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -106,3 +112,113 @@ func (r *Reconciler) ensureManagedServiceAccountUpdated(ctx context.Context, mes
...
+
+	for _, secret := range secretList.Items {
+		klog.V(4).Infof("Deleting Istio remote secret %s/%s", secret.Namespace, secret.Name)
+		if err := client.IgnoreNotFound(r.Delete(ctx, &secret)); err != nil {
+			return fmt.Errorf("failed to delete Istio remote secret %s/%s: %w", secret.Namespace, secret.Name, err)
+		}
+	}
+
+	return nil
+}
```
mkolesnik, line 230: This lists secrets with `istio/multiCluster` label in the hub's `controlPlane.namespace`.
But those remote secrets are only created on spokes by the work agent via MWRS.
They never exist on the hub.

MSA secrets are separate (cluster namespaces, `is-managed-serviceaccount` label) and cleaned up by `deleteAllManagedServiceAccounts`.

Spoke-side cleanup is automatic: mesh deletion -> GC deletes MWRS (owner ref) -> MWRS controller deletes ManifestWorks -> work agent removes spoke secrets.

This function and the `delete` verb on Secrets RBAC (chart/templates/clusterrole.yaml line 14, controller.go line 150) can be removed.

```suggestion
```

pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -106,3 +112,113 @@ func (r *Reconciler) ensureManagedServiceAccountUpdated(ctx context.Context, mes
...
+	}
+
+	klog.Infof("Successfully created a ManifestWorkReplicaSet %s/%s", mesh.Namespace, mesh.Name)
+	return nil
+}
+
+// buildMeshRemoteSecret builds a remote API server access secret.
+// The secret includes required label and annotation for Istio remote endpoint discovery and data from a ManageServiceAccount secret.
+func buildMeshRemoteSecret(mesh *meshv1alpha1.MultiClusterMesh, cluster *clusterv1.ManagedCluster, msaSecret *corev1.Secret) *corev1.Secret {
+	istioRemoteSecretName := fmt.Sprintf("%s-%s-%s-%s", mesh.Namespace, mesh.Name, "istio-remote-secret", cluster.Name)
```
mkolesnik: The name includes `meshNamespace-meshName-istio-remote-secret-clusterName`.
The controller enforces one mesh per `controlPlane.namespace` per ClusterSet, and clusters are exclusive to one set.
The cluster name alone is unique within the target namespace.
Istio identifies remote secrets by the `istio/multiCluster` label and `networking.istio.io/cluster` annotation, not by name.

Either just the cluster name:
```
istioRemoteSecretName := cluster.Name
```

Or with a prefix if you want it clearer when listing secrets in the namespace:
```
istioRemoteSecretName := "istio-remote-secret-" + cluster.Name
```

The namespace is already `istio-system` so the purpose is clear either way.

test/integration/controller_test.go
```diff
@@ -839,6 +840,46 @@ var _ = Describe("MultiClusterMesh Controller", func() {
...
+				util.CreateNamespace(ctx, k8sClient, "istio-system")
+				util.CreateIstioRemoteSecret(ctx, k8sClient, clusterName, meshName, testNs, "istio-system")
+				util.CreateIstioRemoteSecret(ctx, k8sClient, cluster2Name, meshName, testNs, "istio-system")
+				util.DeleteResource(ctx, k8sClient, &meshv1alpha1.MultiClusterMesh{}, meshName, testNs)
+
+				util.ExpectResourceDeleted(ctx, k8sClient, &corev1.Secret{},
+					expectedIstioRemoteSecretName(testNs, meshName, clusterName), "istio-system")
+				util.ExpectResourceDeleted(ctx, k8sClient, &corev1.Secret{},
+					expectedIstioRemoteSecretName(testNs, meshName, cluster2Name), "istio-system")
+			})
```
mkolesnik: This creates hub-side secrets manually and checks `deleteAllRemoteSecrets` removes them.
But in production, remote secrets only exist on spokes (created by work agent via MWRS), never on the hub.
This test validates a dead code path.

Replace with a test that verifies the MWRS is cleaned up after mesh deletion:
```suggestion
It("should cleanup ManifestWorkReplicaSet when mesh is deleted", func() {
    expectManifestWorkReplicaSet(testNs, meshName)
    util.DeleteResource(ctx, k8sClient, &meshv1alpha1.MultiClusterMesh{}, meshName, testNs)
    expectNoManifestWorkReplicaSet(testNs)
})
```

test/integration/controller_test.go
```diff
@@ -857,12 +898,19 @@ var _ = Describe("MultiClusterMesh Controller", func() {
...
+
 			It("should reconcile when the ClusterSet is created", func() {
 				expectMeshNotReady(meshName, testNs)
 				util.CreateManagedCluster(ctx, k8sClient, clusterName, otherClusterSet)
+				util.CreateMsaSecret(ctx, k8sClient, clusterName, meshName, testNs)
 				util.CreateManagedClusterSet(ctx, k8sClient, otherClusterSet)
 				expectManagedClusterSetBinding(testNs, otherClusterSet)
 				expectPlacement(meshName, testNs)
+				expectManifestWorkReplicaSet(testNs, meshName)
```
mkolesnik: This creates an MSA secret inline (line 909) and checks the MWRS exists, mixing two concerns.
The cacerts tests (line 652) follow a cleaner pattern: they separate "controller creates Certificate" from "controller reacts to cert-manager creating the secret."

Move the MWRS assertion into the "MSA secrets exist" context where it belongs.
This test should verify the late-ClusterSet path creates MSAs, Placement, and binding, without also testing MWRS distribution.

test/integration/controller_test.go
```diff
@@ -839,6 +840,46 @@ var _ = Describe("MultiClusterMesh Controller", func() {
 			})
 		})
 
+		When("two clusters have ManagedServiceAccount secrets", func() {
```
mkolesnik: Restructure this context to cover the full MWRS lifecycle.
Rename to something like `When("MSA secrets exist")` and add lifecycle tests that create/remove clusters alongside their MSA secrets:

- **Cluster exists**: Start with 1 cluster + MSA secret, verify MWRS has 1 manifest with correct content.
- **Add cluster**: Add second cluster + MSA secret, trigger reconcile, verify MWRS updates to 2 manifests.
- **Remove cluster**: start with 2 clusters, remove one from the ClusterSet, verify MWRS drops to 1 manifest for the remaining cluster.
- **Secret data update**: change the MSA secret data (token rotation), verify the MWRS manifest content is updated.
- **Partial availability**: 2 clusters but only 1 has an MSA secret, verify MWRS has 1 manifest.
- **MWRS cleanup on mesh deletion**: `expectNoManifestWorkReplicaSet` after deletion (replaces the current hub-side secret cleanup test which exercises dead code).

This separates concerns the same way the cacerts tests do (line 652): MSA creation is tested in the MSA section, this context tests the controller's reaction to MSA secrets existing.

test/integration/controller_test.go
```diff
@@ -839,6 +840,46 @@ var _ = Describe("MultiClusterMesh Controller", func() {
...
+			It("should create ManifestWorkReplicaSet with two Manifests", func() {
+				placement := expectPlacement(meshName, testNs)
+				mwrset := expectManifestWorkReplicaSet(testNs, meshName)
+				Expect(mwrset.OwnerReferences).To(HaveLen(1))
+				Expect(mwrset.OwnerReferences[0].Name).To(Equal(meshName))
+				Expect(mwrset.Labels[meshcontroller.ManagedByLabel]).To(Equal(meshcontroller.ManagedByValue))
+				Expect(mwrset.Labels[meshcontroller.MeshNameLabel]).To(Equal(meshName))
+				Expect(mwrset.Labels[meshcontroller.MeshNamespaceLabel]).To(Equal(testNs))
+				Expect(mwrset.Spec.PlacementRefs[0].Name).To(Equal(placement.Name))
+				Expect(mwrset.Spec.ManifestWorkTemplate.Workload.Manifests).To(HaveLen(2))
```
mkolesnik: `HaveLen(2)` doesn't verify what's inside the manifests.
Unmarshal each manifest and assert on the remote secret properties: name, data keys, labels (`istio/multiCluster`, cluster name), and `networking.istio.io/cluster` annotation.

pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -106,3 +112,113 @@ func (r *Reconciler) ensureManagedServiceAccountUpdated(ctx context.Context, mes
...
+		mwrset.Spec.PlacementRefs = []workv1alpha1.LocalPlacementReference{{Name: placement.Name}}
+		mwrset.Spec.ManifestWorkTemplate = *workTemplate
+		return controllerutil.SetControllerReference(mesh, mwrset, r.Scheme)
+	})
+
+	if err != nil {
+		return fmt.Errorf("failed to ensure ManifestWorkReplicaSet %s/%s: %w", mesh.Namespace, mesh.Name, err)
+	}
+
+	klog.Infof("Successfully created a ManifestWorkReplicaSet %s/%s", mesh.Namespace, mesh.Name)
```
mkolesnik: nit: `CreateOrUpdate` handles both cases.
Consider using the returned `OperationResult` to log the actual operation, or change to "Ensured ManifestWorkReplicaSet".

pkg/hub/mesh/endpoint_discovery.go
yxun: I just found a problem getting the spoke's API server URL from `ManagedCluster.spec.managedClusterClientConfigs[0].url` 

By default `len(cluster.Spec.ManagedClusterClientConfigs)` is 0, this url may be generated from an OCM controller.  I will find a fix later today.

pkg/hub/mesh/controller.go
```diff
@@ -968,7 +1010,7 @@ func (r *Reconciler) ensurePlacement(ctx context.Context, mesh *meshv1alpha1.Mul
 		placement.Spec.ClusterSets = []string{mesh.Spec.ClusterSet}
 		return controllerutil.SetControllerReference(mesh, placement, r.Scheme)
 	})
-	if err != nil {
+	if err != nil && result != controllerutil.OperationResultNone {
```
mkolesnik: This silently swallows all errors.
`CreateOrUpdate` returns `OperationResultNone` on error, so `result != OperationResultNone` is always false when `err != nil`.

```suggestion
	if err != nil {
```

What was the intent here?
If you wanted to suppress no-op updates, that's already handled by `CreateOrUpdate` internally.

test/integration/controller_test.go
```diff
@@ -839,6 +842,93 @@ var _ = Describe("MultiClusterMesh Controller", func() {
...
+
+			It("should update ManifestWorkReplicaSet for newly added cluster", func() {
+				cluster2Name := util.UniqueName("cluster")
+				util.CreateManagedCluster(ctx, k8sClient, cluster2Name, testClusterSet)
+				util.SetManagedClusterURL(ctx, k8sClient, cluster2Name)
+				expectMsaSecret(testNs, meshName, cluster2Name)
+
+				mwrset := expectManifestWorkReplicaSet(meshName, testNs)
+				manifests := mwrset.Spec.ManifestWorkTemplate.Workload.Manifests
+				Expect(manifests).To(HaveLen(2))
```
mkolesnik: `expectManifestWorkReplicaSet` returns immediately because the MWRS already exists from BeforeEach.
The `HaveLen(2)` on line 876 sees stale data (1 manifest from BeforeEach, not 2 yet).

Keep `expectManifestWorkReplicaSet` for existence checks.
Add a separate helper for content assertions that polls until the content matches:

```go
func expectManifestWorkReplicaSetContent(meshName, meshNamespace string, assert func(Gomega, *workv1alpha1.ManifestWorkReplicaSet)) {
    Eventually(func(g Gomega) {
        mwrset := &workv1alpha1.ManifestWorkReplicaSet{}
        g.Expect(k8sClient.Get(ctx, key.Of(meshName, meshNamespace), mwrset)).To(Succeed())
        assert(g, mwrset)
    }).Should(Succeed())
}
```

Then use it in tests that check content after state changes (add cluster, remove cluster, update secret).
Affected lines: 874, 898, 912.

pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -106,3 +115,199 @@ func (r *Reconciler) ensureManagedServiceAccountUpdated(ctx context.Context, mes
...
+	}
+
+	msaName := fmt.Sprintf("%s-%s-%s", mesh.Namespace, "istio-reader", mesh.Name)
+	manifests := []workv1.Manifest{}
+	for _, cluster := range clusters {
+		tokenSecret, err := r.getServiceAccountSecret(ctx, msaName, cluster.Name)
+		if err != nil {
+			if apierrors.IsNotFound(err) {
+				klog.V(4).Infof("managedServiceAccount secret not found yet, waiting for ManagedServiceAccount controller to create it: %s", err)
+				return nil
```
mkolesnik: `return nil` bails out of the entire function.
If cluster1 has a secret but cluster2 doesn't, NO MWRS is created or updated.
During cluster addition, existing clusters' remote secrets are not distributed.

The test at line 921 passes by accident: it checks stale state from BeforeEach, not intentional partial creation.

```suggestion
			if apierrors.IsNotFound(err) {
				klog.V(4).Infof("managedServiceAccount secret not found yet for cluster %s, skipping", cluster.Name)
				continue
```

test/integration/controller_test.go
```diff
@@ -839,6 +842,93 @@ var _ = Describe("MultiClusterMesh Controller", func() {
...
+				slices.SortFunc(manifests, func(a, b workv1.Manifest) int {
+					secret := &corev1.Secret{}
+					Expect(unmarshalManifest(a, secret)).To(Succeed())
+					secret2 := &corev1.Secret{}
+					Expect(unmarshalManifest(b, secret2)).To(Succeed())
+					return cmp.Compare(secret2.Name, secret.Name)
+				})
+
+				expectRemoteSecret(mwrset.Spec.ManifestWorkTemplate.Workload.Manifests[0], clusterName, "istio-system")
+				expectRemoteSecret(mwrset.Spec.ManifestWorkTemplate.Workload.Manifests[1], cluster2Name, "istio-system")
```
mkolesnik: Sorting manifests by random names and asserting by index is a 50% flake.
Don't sort and index. Iterate the expected clusters and find the matching manifest:

```suggestion
				for _, cluster := range []string{clusterName, cluster2Name} {
					found := false
					for _, m := range manifests {
						secret := &corev1.Secret{}
						Expect(unmarshalManifest(m, secret)).To(Succeed())
						if secret.Name == "istio-remote-secret-"+cluster {
							expectRemoteSecret(m, cluster, "istio-system")
							found = true
							break
						}
					}
					Expect(found).To(BeTrue(), "no remote secret found for cluster %s", cluster)
				}
```

pkg/hub/mesh/controller.go
```diff
@@ -125,6 +126,10 @@ func RegisterController(mgr manager.Manager) error {
 				return obj.GetLabels()[MeshNameLabel] != "" && obj.GetLabels()[MeshNamespaceLabel] != ""
 			})),
 		).
+		Watches(&corev1.Secret{},
+			handler.EnqueueRequestsFromMapFunc(reconciler.mapSecretToMSA),
+			builder.WithPredicates(predicate.GenerationChangedPredicate{}),
+		).
```
mkolesnik: This watches every Secret in the cluster and reverse-lookups which MSA references it via a List call per event.
The MSA already has mesh labels and `Status.TokenSecretRef`.
Watch MSA status directly, drop `mapSecretToMSA` entirely:

```suggestion
		Watches(&msav1beta1.ManagedServiceAccount{},
			handler.EnqueueRequestsFromMapFunc(reconciler.findMeshesForMSA),
			builder.WithPredicates(predicate.NewPredicateFuncs(func(obj client.Object) bool {
				return obj.GetLabels()[MeshNameLabel] != "" && obj.GetLabels()[MeshNamespaceLabel] != ""
			})),
		).
```

And add the mapper (same style as `mapSecretToMesh`):

```go
func (r *Reconciler) findMeshesForMSA(_ context.Context, obj client.Object) []reconcile.Request {
	meshName := obj.GetLabels()[MeshNameLabel]
	meshNamespace := obj.GetLabels()[MeshNamespaceLabel]

	klog.V(4).Infof("ManagedServiceAccount %s/%s triggered reconcile for mesh %s/%s",
		obj.GetNamespace(), obj.GetName(), meshNamespace, meshName)

	return []reconcile.Request{{NamespacedName: key.Of(meshName, meshNamespace)}}
}
```

`mapSecretToMSA` can be deleted.

pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -106,3 +115,199 @@ func (r *Reconciler) ensureManagedServiceAccountUpdated(ctx context.Context, mes
...
+	secret := &corev1.Secret{}
+	if err := r.Get(ctx, key.Of(serviceAccountName, clusterName), secret); err != nil {
+		return nil, err
+	}
+
+	if err := secretReferencesServiceAccount(sa, secret); err != nil {
+		return nil, err
+	}
+	return secret, nil
+}
```
mkolesnik: This fetches a ServiceAccount on the hub, but MSA creates the SA on the spoke, not the hub.
The test creates a fake SA to make this pass.

Replace with MSA's `TokenSecretRef`:

```suggestion
func (r *Reconciler) getMSATokenSecret(ctx context.Context, msaName, clusterName string) (*corev1.Secret, error) {
	msa := &msav1beta1.ManagedServiceAccount{}
	if err := r.Get(ctx, key.Of(msaName, clusterName), msa); err != nil {
		return nil, err
	}
	if msa.Status.TokenSecretRef == nil {
		return nil, nil
	}
	secret := &corev1.Secret{}
	return secret, r.Get(ctx, key.Of(msa.Status.TokenSecretRef.Name, clusterName), secret)
}
```

The caller skips the cluster when `secret == nil` (same as the `continue` on NotFound).
`secretReferencesServiceAccount` can be deleted.

pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -106,3 +115,199 @@ func (r *Reconciler) ensureManagedServiceAccountUpdated(ctx context.Context, mes
 	klog.V(4).Infof("Updated ManagedServiceAccount %s/%s", existing.Namespace, existing.Name)
 	return nil
 }
+
+// ensureManifestWorkReplicaSet creates a ManifestWorkReplicaSet to distribute remote access secrets for clusters selected by a Placement
+func (r *Reconciler) ensureManifestWorkReplicaSet(ctx context.Context, mesh *meshv1alpha1.MultiClusterMesh) error {
+	clusters, err := r.getClustersFromSet(ctx, mesh.Spec.ClusterSet)
+	if err != nil {
+		return fmt.Errorf("failed to get clusters from set %s: %w", mesh.Spec.ClusterSet, err)
+	}
```
mkolesnik: `doReconcile` already called `getClustersFromSet` and has the clusters list.
Every other function in the reconcile loop takes `clusters` as a parameter.
This function should too, instead of making a redundant API/cache call.

pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -106,3 +115,199 @@ func (r *Reconciler) ensureManagedServiceAccountUpdated(ctx context.Context, mes
...
+
+	if len(clusters) == 0 {
+		return nil
+	}
+
```
mkolesnik: nit: redundant.
If clusters is empty, the loop doesn't execute, `manifests` stays empty, and `len(manifests) == 0` on line 158 returns nil anyway.

```suggestion
```

pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -106,3 +115,199 @@ func (r *Reconciler) ensureManagedServiceAccountUpdated(ctx context.Context, mes
 	klog.V(4).Infof("Updated ManagedServiceAccount %s/%s", existing.Namespace, existing.Name)
 	return nil
 }
+
+// ensureManifestWorkReplicaSet creates a ManifestWorkReplicaSet to distribute remote access secrets for clusters selected by a Placement
+func (r *Reconciler) ensureManifestWorkReplicaSet(ctx context.Context, mesh *meshv1alpha1.MultiClusterMesh) error {
```
mkolesnik: `ensureManifestWorkReplicaSet` doesn't say what it actually does.
It builds Istio remote discovery secrets from MSA tokens and distributes them via MWRS.

```suggestion
func (r *Reconciler) ensureRemoteSecretDistribution(ctx context.Context, mesh *meshv1alpha1.MultiClusterMesh) error {
```

Same downstream:
- `createRemoteSecret` -> `buildIstioRemoteSecret` (it builds, doesn't create via API)
- `getServiceAccountSecret` -> `getMSATokenSecret` (it's the MSA token, not a generic SA secret)

test/integration/controller_test.go
```diff
@@ -839,6 +842,93 @@ var _ = Describe("MultiClusterMesh Controller", func() {
...
+			})
+
+			It("should drop one ManifestWorkReplicaSet manifest when removing one cluster from the ManagedClusterSet", func() {
```
mkolesnik, line 893: The "remove one cluster" test is here, but there's no test for removing ALL clusters.
When `len(clusters) == 0`, the code returns nil without updating or deleting the MWRS.
The MWRS stays with stale manifests pointing to clusters that no longer exist.

Add a test that removes all clusters and verifies the MWRS is cleaned up (or at least updated to have zero manifests).

pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -106,3 +115,199 @@ func (r *Reconciler) ensureManagedServiceAccountUpdated(ctx context.Context, mes
...
+				"istio/multiCluster": "true",
+			},
+		},
+		Data: map[string][]byte{
+			key: data.Bytes(),
+		},
+		Type: corev1.SecretTypeOpaque,
+	}
+	return out, nil
+}
```
mkolesnik: The helper chain is 4 levels deep:
`createRemoteSecret` -> `createRemoteSecretFromTokenAndServer` -> `tokenDataFromSecret` + `createBearerTokenKubeconfig` -> `createBaseKubeconfig` + `createRemoteServiceAccountSecret`

This is "extract token, build kubeconfig, serialize to secret."
Flatten into one function:

```go
func buildIstioRemoteSecret(tokenSecret *corev1.Secret, clusterName, server, namespace string) (*corev1.Secret, error) {
	ca, ok := tokenSecret.Data[corev1.ServiceAccountRootCAKey]
	if !ok {
		return nil, fmt.Errorf("no %q data found", corev1.ServiceAccountRootCAKey)
	}
	token, ok := tokenSecret.Data[corev1.ServiceAccountTokenKey]
	if !ok {
		return nil, fmt.Errorf("no %q data found", corev1.ServiceAccountTokenKey)
	}

	kubeconfig := &api.Config{
		Clusters:       map[string]*api.Cluster{clusterName: {CertificateAuthorityData: ca, Server: server}},
		AuthInfos:      map[string]*api.AuthInfo{clusterName: {Token: string(token)}},
		Contexts:       map[string]*api.Context{clusterName: {Cluster: clusterName, AuthInfo: clusterName}},
		CurrentContext: clusterName,
	}
	if err := clientcmd.Validate(*kubeconfig); err != nil {
		return nil, fmt.Errorf("invalid kubeconfig: %w", err)
	}

	var buf bytes.Buffer
	if err := latest.Codec.Encode(kubeconfig, &buf); err != nil {
		return nil, fmt.Errorf("failed to encode kubeconfig: %w", err)
	}

	return &corev1.Secret{
		ObjectMeta: metav1.ObjectMeta{
			Name:        "istio-remote-secret-" + clusterName,
			Namespace:   namespace,
			Annotations: map[string]string{"networking.istio.io/cluster": clusterName},
			Labels:      map[string]string{"istio/multiCluster": "true"},
		},
		Data: map[string][]byte{clusterName: buf.Bytes()},
		Type: corev1.SecretTypeOpaque,
	}, nil
}
```

One function, no intermediaries.

test/util/k8s.go
```diff
@@ -65,6 +65,41 @@ func CreateCacertsSecret(ctx context.Context, k8sClient client.Client, namespace
...
 
+// CreateMsaSecret creates a ServiceAccount and a secret that simulates what ManagedServiceAccount controller would create.
+func CreateMsaSecret(ctx context.Context, k8sClient client.Client, msaName, clusterName string) {
+	sa := &corev1.ServiceAccount{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      msaName,
+			Namespace: clusterName,
+		},
+	}
+	Expect(k8sClient.Create(ctx, sa)).To(Succeed())
```
mkolesnik: This fake SA only exists because `getServiceAccountSecret` looks up an SA on the hub.
Once that's fixed to use `msa.Status.TokenSecretRef`, drop it:

```suggestion
func CreateMsaSecret(ctx context.Context, k8sClient client.Client, msaName, clusterName string) {
```

The test flow in `expectMsaSecret` is already correct:
1. `expectManagedServiceAccount` - wait for controller to create the MSA
2. `SetMsaStatus` - simulate MSA controller setting `TokenSecretRef`
3. `CreateMsaSecret` - simulate MSA controller creating the token Secret

test/integration/controller_test.go
```diff
@@ -1100,6 +1195,10 @@ func expectedManagedServiceAccountName(meshNamespace, meshName string) string {
 	return fmt.Sprintf("%s-istio-reader-%s", meshNamespace, meshName)
 }
 
+func expectedIstioRemoteSecretName(meshNamespace, meshName, clusterName string) string {
+	return fmt.Sprintf("%s-%s-%s-%s", meshNamespace, meshName, "istio-remote-secret", clusterName)
+}
+
```
mkolesnik: nit: dead code, no longer called.

```suggestion
```

pkg/hub/mesh/controller.go
```diff
@@ -752,6 +762,38 @@ func (r *Reconciler) mapSecretToMesh(_ context.Context, obj client.Object) []rec
...
+		// Check if this MSA uses the secret that just triggered the event
+		if msa.Status.TokenSecretRef == nil {
+			return nil
+		}
+
+		if msa.Status.TokenSecretRef.Name == secret.Name {
+			meshName := msa.Labels[MeshNameLabel]
+			meshNamespace := msa.Labels[MeshNamespaceLabel]
+
+			klog.V(4).Infof("ManagedServiceAccount Secret %s%s triggered reconcile for mesh %s/%s",
```
mkolesnik: nit: missing `/` between namespace and name.
```suggestion
			klog.V(4).Infof("ManagedServiceAccount Secret %s/%s triggered reconcile for mesh %s/%s",
```

pkg/hub/mesh/controller.go
yxun: ~When there is an existing Placement, mesh reconcile will ensure and call the CreateOrUpdate placement. If the `CreateOrUpdate` result is `OperationResultNone` (unchanged), that `CreateOrUpdate` will also return an error.
We want to capture errors except existing Placement unchanged case.~

I misunderstood the following error return description. I added that when I was fixing a test error of existing Placement issue. I will remove this.

```
If the object did exist, MutateFn will be called, and if it changed the object, it will be updated. Otherwise, it will be left unchanged. The executed operation (and an error) will be returned.
```

test/integration/controller_test.go
```diff
@@ -839,6 +840,102 @@ var _ = Describe("MultiClusterMesh Controller", func() {
...
+				manifests := mwrset.Spec.ManifestWorkTemplate.Workload.Manifests
+				oldSec := &corev1.Secret{}
+				Expect(unmarshalManifest(manifests[0], oldSec)).To(Succeed())
+				oldData := oldSec.Data[clusterName]
+				expectMsaSecretUpdated(testNs, meshName, clusterName)
+
+				expectManifestWorkReplicaSetContent(meshName, testNs, func(g Gomega, mwrset *workv1alpha1.ManifestWorkReplicaSet) {
+					manifests = mwrset.Spec.ManifestWorkTemplate.Workload.Manifests
+					newSec := &corev1.Secret{}
+					Expect(unmarshalManifest(manifests[0], newSec)).To(Succeed())
```
mkolesnik: Bare `Expect` inside an Eventually callback aborts the test instead of retrying.
Use `g.Expect`:

```suggestion
					g.Expect(unmarshalManifest(manifests[0], newSec)).To(Succeed())
```

test/util/ocm.go
```diff
@@ -34,9 +36,37 @@ func CreateManagedCluster(ctx context.Context, k8sClient client.Client, name, cl
...
+func UpdateMsaTokenRotation(ctx context.Context, k8sClient client.Client, msaName, clusterName string) {
+	msa := &msav1beta1.ManagedServiceAccount{}
+	Expect(k8sClient.Get(ctx, key.Of(msaName, clusterName), msa)).To(Succeed())
+	msa.Spec.Rotation.Validity = metav1.Duration{Duration: 10 * time.Minute}
+	msa.Status = msav1beta1.ManagedServiceAccountStatus{
+		ExpirationTimestamp: &metav1.Time{Time: time.Now().Add(10 * time.Minute)},
+	}
+	Expect(k8sClient.Update(ctx, msa)).To(Succeed())
+	Expect(k8sClient.Status().Update(ctx, msa)).To(Succeed())
+}
```
mkolesnik: This replaces the entire MSA status, wiping `TokenSecretRef` and conditions.
After this, `getMSATokenSecret` finds `TokenSecretRef == nil` and returns nil.
The token rotation test can't actually verify the MWRS picks up the new token.

This function doesn't need to exist.
`SetMsaStatus` already sets `TokenSecretRef` with `LastRefreshTimestamp: metav1.Now()`.
On rotation, the same fields change.
Just call `SetMsaStatus` again.

In `expectMsaSecretUpdated`:

```go
func expectMsaSecretUpdated(meshNamespace, meshName, clusterName string) {
	msa := expectManagedServiceAccount(meshNamespace, meshName, clusterName)
	util.UpdateMsaSecret(ctx, k8sClient, msa.Name, clusterName)
	util.SetMsaStatus(ctx, k8sClient, msa.Name, clusterName)
	expectMeshNotReady(meshName, meshNamespace)
}
```

Delete `UpdateMsaTokenRotation` entirely.

pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -106,3 +115,125 @@ func (r *Reconciler) ensureManagedServiceAccountUpdated(ctx context.Context, mes
 	klog.V(4).Infof("Updated ManagedServiceAccount %s/%s", existing.Namespace, existing.Name)
 	return nil
 }
+
+// ensureRemoteSecretDistribution builds Istio remote discovery secrets from ManagedServiceAccount tokens and distributes them via ManifestWorkReplicaSet.
+func (r *Reconciler) ensureRemoteSecretDistribution(ctx context.Context, mesh *meshv1alpha1.MultiClusterMesh, clusters []clusterv1.ManagedCluster) error {
+	msaName := fmt.Sprintf("%s-%s-%s", mesh.Namespace, "istio-reader", mesh.Name)
```
mkolesnik: `fmt.Sprintf("%s-%s-%s", mesh.Namespace, "istio-reader", mesh.Name)` is duplicated from `ensureManagedServiceAccount` (line 29).
Extract a helper:

```go
func msaName(mesh *meshv1alpha1.MultiClusterMesh) string {
	return fmt.Sprintf("%s-istio-reader-%s", mesh.Namespace, mesh.Name)
}
```

pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -106,3 +115,125 @@ func (r *Reconciler) ensureManagedServiceAccountUpdated(ctx context.Context, mes
...
+	manifests := []workv1.Manifest{}
+	for _, cluster := range clusters {
+		if len(cluster.Spec.ManagedClusterClientConfigs) == 0 {
```
mkolesnik, line 128: A cluster with no client configs is a transient state (just registered, hub hasn't received the config yet).
Returning an error fails the entire reconcile for all clusters.

```suggestion
		if len(cluster.Spec.ManagedClusterClientConfigs) == 0 {
			klog.V(4).Infof("Cluster %s has no client configs yet, skipping", cluster.Name)
			continue
		}
```

pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -106,3 +115,125 @@ func (r *Reconciler) ensureManagedServiceAccountUpdated(ctx context.Context, mes
...
+			RawExtension: runtime.RawExtension{Object: &corev1.Secret{
+				TypeMeta: metav1.TypeMeta{
+					APIVersion: "v1",
+					Kind:       "Secret",
+				},
+				ObjectMeta: remoteSecret.ObjectMeta,
+				Data:       remoteSecret.Data,
+				Type:       remoteSecret.Type,
+			}}})
+	}
```
mkolesnik, line 153: `buildIstioRemoteSecret` doesn't set TypeMeta, so the caller wraps the result in a new Secret object just to add it.
This double-wrapping silently drops any fields added to `buildIstioRemoteSecret` in the future.

Set TypeMeta inside `buildIstioRemoteSecret` (matching the existing pattern in `buildOperatorManifestWork`, `buildCacertsManifestWork`):

```go
return &corev1.Secret{
    TypeMeta: metav1.TypeMeta{APIVersion: "v1", Kind: "Secret"},
    ObjectMeta: ...
```

Then the caller simplifies to:

```go
manifests = append(manifests, workv1.Manifest{
    RawExtension: runtime.RawExtension{Object: remoteSecret},
})
```

pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -106,3 +115,125 @@ func (r *Reconciler) ensureManagedServiceAccountUpdated(ctx context.Context, mes
...
+
+	if err := r.Get(ctx, key.Of(mesh.Name, mesh.Namespace), mwrset); err == nil && len(clusters) == 0 {
+		klog.Infof("Deleting ManifestWorkReplicaSet %s/%s (clusterSet is empty)", mesh.Namespace, mesh.Name)
+		return r.Delete(ctx, mwrset)
+	}
+
+	if len(manifests) == 0 {
+		return nil
+	}
```
mkolesnik: The MWRS uses `SetControllerReference` for lifecycle management.
Explicitly deleting it here is inconsistent with the ownership model.

Drop the deletion path and the `len(manifests) == 0` early return.
Let `CreateOrUpdate` handle it: when `manifests` is empty, the MWRS gets updated with an empty manifest list.
When the mesh is deleted, GC cleans up the MWRS.

```suggestion
	mwrset := &workv1alpha1.ManifestWorkReplicaSet{
		ObjectMeta: metav1.ObjectMeta{Name: mesh.Name, Namespace: mesh.Namespace},
	}

	result, err := controllerutil.CreateOrUpdate(ctx, r.Client, mwrset, func() error {
```

No `delete` RBAC needed either.
The test "should cleanup ManifestWorkReplicaSet" should check for empty manifests instead of deletion.

pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -106,3 +115,125 @@ func (r *Reconciler) ensureManagedServiceAccountUpdated(ctx context.Context, mes
...
+	if msa.Status.TokenSecretRef == nil {
+		return nil, nil
+	}
+	secret := &corev1.Secret{}
+	return secret, r.Get(ctx, key.Of(msa.Status.TokenSecretRef.Name, clusterName), secret)
+}
+
+// buildIstioRemoteSecret builds a remote API server access secret.
+// The secret includes required label and annotation for Istio remote endpoint discovery and data from a ManageServiceAccount secret.
```
mkolesnik: nit: typo.

```suggestion
// buildIstioRemoteSecret builds a remote API server access secret.
// The secret includes required label and annotation for Istio remote endpoint discovery and data from a ManagedServiceAccount secret.
```

test/util/k8s.go
```diff
@@ -65,6 +65,33 @@ func CreateCacertsSecret(ctx context.Context, k8sClient client.Client, namespace
...
+func CreateMsaSecret(ctx context.Context, k8sClient client.Client, msaName, clusterName string) {
+	secret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      msaName,
+			Namespace: clusterName,
+			Annotations: map[string]string{
+				"kubernetes.io/service-account.name": msaName,
+			},
+		},
+		Type: corev1.SecretTypeServiceAccountToken,
```
mkolesnik: nit: the real MSA controller creates an `Opaque` secret with label `authentication.open-cluster-management.io/is-managed-serviceaccount: "true"` (verified on dev-env).
The test creates `SecretTypeServiceAccountToken` with SA annotations.
Doesn't affect the controller (it doesn't check type), but it's an inaccurate simulation.

```suggestion
		Type: corev1.SecretTypeOpaque,
```

test/util/ocm.go
yxun: Initially I only used the `SetMsaStatus` . However, MSA status change didn't trigger mesh's MSA watch . See I pushed another commit in the fix test error commit.
I will try your `g.Expect(unmarshalManifest` suggestion and this soon.

test/util/k8s.go
yxun: I got ServiceAccount type MSA secret from ACM OpenShift hub cluster before. Let me double check that today

test/e2e/multi_primary_test.go
```diff
@@ -144,10 +144,6 @@ var _ = Describe("Multi-primary data plane", Ordered, Serial, func() {
 				Success("istiod is ready on %s", cluster)
 			}
 
-			// TODO(endpoint-discovery): Remove once the controller distributes remote secrets.
-			Step("Creating remote secrets for cross-cluster discovery")
-			util.CreateAndDistributeRemoteSecrets(ctx, hubClient, spokeClients, clusters, cpNamespace)
-
```
sridhargaddam, line 173: Since this PR is not yet creating the ClusterRole/ClusterRoleBinding for the ServiceAccount associated with the MSA, the actual data-plane validation fails. So, **lets not remove** this code until you implement the RBAC.

JFYI, when i cloned this PR and ran the test-e2e-multicluster, it fails (as expected).

```
• [FAILED] [213.298 seconds]
Multi-primary data plane when Istio multi-cluster is deployed [It] should have cross-cluster data plane traffic working [Serial]
/home/sgaddam/goprojects/istio/multicluster-mesh-addon/test/e2e/multi_primary_test.go:180

  [FAILED] Timed out after 180.051s.
  The function passed to Eventually failed at /home/sgaddam/goprojects/istio/multicluster-mesh-addon/test/e2e/multi_primary_test.go:229 with:
  never saw v2 from cluster1
  Expected
      <bool>: false
  to be true
  In [It] at: /home/sgaddam/goprojects/istio/multicluster-mesh-addon/test/e2e/multi_primary_test.go:230 @ 08/13/26 21:11:41.82
------------------------------

Summarizing 1 Failure:
  [FAIL] Multi-primary data plane when Istio multi-cluster is deployed [It] should have cross-cluster data plane traffic working [Serial]
  /home/sgaddam/goprojects/istio/multicluster-mesh-addon/test/e2e/multi_primary_test.go:230

Ran 2 of 2 Specs in 349.059 seconds
FAIL! -- 1 Passed | 1 Failed | 0 Pending | 0 Skipped
```

yxun: ok, I will revert this commit

test/util/ocm.go
yxun: Hi @mkolesnik , right, that `UpdateMsaTokenRotation` was my attempt to fix the test error of `"should update ManifestWorkReplicaSet when ManagedServiceAccount secret is updated"` 

I tried several redundant checks such as the len(manifests) and return errors as well. 

Now the root cause and fix of that is in the last fix integration test error commit. Thanks for your help and I have updated all suggestions and removed all those redundant complexities.

yxun:
@sridhargaddam I have reverted my e2e change in this PR. It looks like the existing setup does not clean remote secret in a Kind environment. 

Is this `[E2E Multicluster Tests (Kind)]` job required here ? I will validate next PR CR and CRB change with e2e tests but this PR has no impact on existing e2e tests.

The remote secrets cleanup flow was discussed in https://github.com/stolostron/multicluster-mesh-addon/pull/215/changes#r3728276710
So we don't have a cleanup method for that in this PR.

sridhargaddam:
> @sridhargaddam I have reverted my e2e change in this PR. It looks like the existing setup does not clean remote secret in a Kind environment.

The `remoteSecret` creation in the current e2e-multicluster testsuite is just a stopgap solution until the addon controller implements the necessary changes. Coming back to your question, since the `remoteSecrets` are created on the spoke clusters in the `controlPlane` namespace, deleting the `MultiClusterMesh` CR also deletes the `controlPlane` namespace and, consequently, the `remoteSecrets`. Because of this, we did not include an explicit step to delete the `remoteSecrets` as part of this temporary solution. Hope this clarifies your question.

> Is this `[E2E Multicluster Tests (Kind)]` job required here ? I will validate next PR CR and CRB change with e2e tests but this PR has no impact on existing e2e tests.

Sure.

test/util/ocm.go
```diff
@@ -34,9 +36,26 @@ func CreateManagedCluster(ctx context.Context, k8sClient client.Client, name, cl
...
 
+// SetMsaStatus updates a ManagedServiceAccount's status TokenSecretRef,
+// simulating what the ManagedServiceAccount controller does
+func SetMsaStatus(ctx context.Context, k8sClient client.Client, msaName, clusterName string, testDuration time.Duration) {
+	msa := &msav1beta1.ManagedServiceAccount{}
+	Expect(k8sClient.Get(ctx, key.Of(msaName, clusterName), msa)).To(Succeed())
+	msa.Status = msav1beta1.ManagedServiceAccountStatus{
+		TokenSecretRef: &msav1beta1.SecretRef{
+			Name:                 msaName,
+			LastRefreshTimestamp: metav1.NewTime(metav1.Now().Add(testDuration)),
```
yxun: I added a testDurtion for the `LastRefreshTimestamp` . This is for fixing an integration test issue.

The issue was:
When updating MSA , it sets status `LastRefreshTimestamp : metav1.Now()` value type metav1.Time.
However, metav1.Time only has second-level precision. 
When integration test runs, it's fast and set same timestamp in updates. So MSA status update didn't trigger watch reconciliation .

test/integration/controller_test.go
```diff
@@ -844,6 +845,113 @@ var _ = Describe("MultiClusterMesh Controller", func() {
...
+				expectRemoteSecret(mwrset.Spec.ManifestWorkTemplate.Workload.Manifests[0], cluster2Name, "istio-system")
+			})
+
+			It("should cleanup ManifestWorkReplicaSet Manifests when removing all clusters from the ManagedClusterSet", func() {
+				updateClusterSetLabel(clusterName, "")
+				Eventually(func() []workv1.Manifest {
+					mwrset := &workv1alpha1.ManifestWorkReplicaSet{}
+					Expect(k8sClient.Get(ctx, key.Of(meshName, testNs), mwrset))
+					return mwrset.Spec.ManifestWorkTemplate.Workload.Manifests
+				}).Should(BeEmpty())
```
mkolesnik: Bare `Expect` inside `Eventually` aborts the test instead of retrying.
Use the `func(g Gomega)` form:

```suggestion
				Eventually(func(g Gomega) {
					mwrset := &workv1alpha1.ManifestWorkReplicaSet{}
					g.Expect(k8sClient.Get(ctx, key.Of(meshName, testNs), mwrset)).To(Succeed())
					g.Expect(mwrset.Spec.ManifestWorkTemplate.Workload.Manifests).To(BeEmpty())
				}).Should(Succeed())
```

test/integration/controller_test.go
```diff
@@ -844,6 +845,113 @@ var _ = Describe("MultiClusterMesh Controller", func() {
...
+				})
+			})
+
+			It("should create ManifestWorkReplicaSet with one manifest when only one cluster has a ManagedServiceAccount secret", func() {
+				cluster2Name := util.UniqueName("cluster")
+				util.CreateManagedCluster(ctx, k8sClient, cluster2Name, testClusterSet)
+
+				mwrset := expectManifestWorkReplicaSet(meshName, testNs)
+				Expect(mwrset.Spec.ManifestWorkTemplate.Workload.Manifests).To(HaveLen(1))
+				expectRemoteSecret(mwrset.Spec.ManifestWorkTemplate.Workload.Manifests[0], clusterName, "istio-system")
```
mkolesnik: `expectManifestWorkReplicaSet` returns immediately because the MWRS already exists from BeforeEach. The `HaveLen(1)` check may run before the reconcile for cluster2 even triggers, so it passes vacuously.

This is a negative assertion (adding a cluster without an MSA secret should not add a manifest). Use `Consistently`:

```go
Consistently(func(g Gomega) {
    mwrset := &workv1alpha1.ManifestWorkReplicaSet{}
    g.Expect(k8sClient.Get(ctx, key.Of(meshName, testNs), mwrset)).To(Succeed())
    g.Expect(mwrset.Spec.ManifestWorkTemplate.Workload.Manifests).To(HaveLen(1))
    expectRemoteSecret(mwrset.Spec.ManifestWorkTemplate.Workload.Manifests[0], clusterName, "istio-system")
}).Should(Succeed())
```

Also in general I'm not sure if this is very interesting to test, thinking about it some more it seems to me we care about the secrets being distributed when needed.. I'm not sure we care about intermediary states such as this since it's not really adding much value (there's no real state we care about here).

pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -106,3 +119,111 @@ func (r *Reconciler) ensureManagedServiceAccountUpdated(ctx context.Context, mes
...
+
+// ensureRemoteSecretDistribution builds Istio remote discovery secrets from ManagedServiceAccount tokens and distributes them via ManifestWorkReplicaSet.
+func (r *Reconciler) ensureRemoteSecretDistribution(ctx context.Context, mesh *meshv1alpha1.MultiClusterMesh, clusters []clusterv1.ManagedCluster) error {
+	msaName := msaName(mesh)
+	manifests := []workv1.Manifest{}
+	for _, cluster := range clusters {
+		if len(cluster.Spec.ManagedClusterClientConfigs) == 0 {
+			klog.V(4).Infof("Cluster %s has no client configs yet, skipping", cluster.Name)
+			continue
+		}
```
mkolesnik, line 131: `ManagedClusterClientConfigs` is set during OCM registration, not populated asynchronously. A cluster without it may never get one, so "yet" is misleading.

More importantly, this skip is a prerequisite gap that's invisible to the user. The mesh status should reflect per-cluster endpoint discovery state (e.g., "secret distribution pending: no API endpoint for cluster X").

Please file a follow-up issue for this, so we don't forget to handle this state.

hack/dev-env.sh
```diff
@@ -185,6 +185,12 @@ init_ocm() {
     log "Waiting for OCM hub components to be ready..."
     on "${HUB}" retry kubectl wait --for=condition=Available \
         deployment/cluster-manager -n open-cluster-management --timeout=120s
+
+    log "Enabling feature ManifestWorkReplicaSet on cluster: ${HUB}"
+    on "${HUB}" kubectl patch clustermanager cluster-manager --type=merge \
+        -p '{"spec":{"workConfiguration":{"featureGates":[{"feature":"ManifestWorkReplicaSet","mode":"Enable"}]}}}'
+    on "${HUB}" retry kubectl wait --for=condition=Available \
+        deployment/cluster-manager -n open-cluster-management --timeout=120s
```
mkolesnik: `clusteradm init` already supports `--feature-gates=ManifestWorkReplicaSet=true` so you can simply add this argument at L183.

This is more straightforward and removes the need for `kubectl patch` + second `kubectl wait` here.

```suggestion
```

test/integration/controller_test.go
```diff
@@ -844,6 +845,113 @@ var _ = Describe("MultiClusterMesh Controller", func() {
...
+					msa = &msav1beta1.ManagedServiceAccount{}
+					g.Expect(k8sClient.Get(ctx, key.Of(expectedManagedServiceAccountName(testNs, meshName), clusterName), msa)).To(Succeed())
+					g.Expect(msa.Status.TokenSecretRef.LastRefreshTimestamp).NotTo(Equal(oldTime))
+
+					manifests = mwrset.Spec.ManifestWorkTemplate.Workload.Manifests
+					newSec := &corev1.Secret{}
+					g.Expect(unmarshalManifest(manifests[0], newSec)).To(Succeed())
+					newData := newSec.Data[clusterName]
+					g.Expect(newData).NotTo(Equal(oldData))
+				})
```
mkolesnik: This Eventually checks two independent things: MSA status changed (L933-935) and MWRS content changed (L937-941). If the MSA status doesn't update, the failure message is about the MSA, not the MWRS content this test is named for.

Separate the concerns: wait for MSA status change first, then wait for MWRS content change. Each Eventually should verify one thing.

test/integration/controller_test.go
```diff
@@ -844,6 +845,113 @@ var _ = Describe("MultiClusterMesh Controller", func() {
...
+						secret := &corev1.Secret{}
+						Expect(unmarshalManifest(m, secret)).To(Succeed())
+						if secret.Name == "istio-remote-secret-"+cluster {
+							expectRemoteSecret(m, cluster, "istio-system")
+							found = true
+							break
+						}
+					}
+					Expect(found).To(BeTrue(), "no remote secret found for cluster %s", cluster)
+				}
```
mkolesnik: The HaveLen(2) check is a proxy for what the test actually cares about: the right remote secrets are distributed. Put the full assertion inside the `expectManifestWorkReplicaSetContent` callback instead of checking count, re-fetching, and asserting content separately:

```go
expectManifestWorkReplicaSetContent(meshName, testNs, func(g Gomega, mwrset *workv1alpha1.ManifestWorkReplicaSet) {
    g.Expect(mwrset.Spec.ManifestWorkTemplate.Workload.Manifests).To(HaveLen(2))
    for _, cluster := range []string{clusterName, cluster2Name} {
        found := false
        for _, m := range mwrset.Spec.ManifestWorkTemplate.Workload.Manifests {
            secret := &corev1.Secret{}
            g.Expect(unmarshalManifest(m, secret)).To(Succeed())
            if secret.Name == "istio-remote-secret-"+cluster {
                found = true
                break
            }
        }
        g.Expect(found).To(BeTrue(), "no remote secret for %s", cluster)
    }
})
```
Same approach for L894-906 (remove one cluster) and L917-943 (secret update).

test/integration/controller_test.go
```diff
@@ -1168,6 +1276,47 @@ func expectPlacement(meshName, meshNamespace string) *clusterv1beta1.Placement {
...
+	Expect(secret.Type).To(Equal(corev1.SecretTypeOpaque))
+	Expect(secret.Data).To(HaveKey(clusterName))
+}
+
+func expectMsaSecret(meshNamespace, meshName, clusterName string) {
+	msa := expectManagedServiceAccount(meshNamespace, meshName, clusterName)
+	util.CreateMsaSecret(ctx, k8sClient, msa.Name, clusterName)
+	util.SetMsaStatus(ctx, k8sClient, msa.Name, clusterName, time.Duration(0))
+	expectMeshNotReady(meshName, meshNamespace)
+}
```
mkolesnik: `expectMsaSecret` does: wait for MSA, create fake secret,  MSA status, wait for mesh not-ready.

That's multi-step setup, not assertion.
Same for `expectMsaSecretUpdated`.

The "expect" prefix suggests verifying pre-existing state.

For reference, the cacerts tests do this inline: `util.CreateCacertsSecret(...)` then `expectCacertsManifestWork(...)` — each step visible.

Either rename to `setupMsaTokenSecret` / `simulateTokenRotation`, or inline the steps like the cacerts pattern does.

pkg/hub/mesh/controller.go
```diff
@@ -772,6 +784,16 @@ func (r *Reconciler) mapSecretToMesh(_ context.Context, obj client.Object) []rec
 	return []reconcile.Request{{NamespacedName: key.Of(meshName, meshNamespace)}}
 }
 
+func (r *Reconciler) findMeshesForMSA(_ context.Context, obj client.Object) []reconcile.Request {
```
mkolesnik: This is a direct 1:1 label-to-request mapping, same shape as `mapSecretToMesh` right above it. The other `findMeshesForX` functions List and can return zero/one/many. Rename to `mapMsaToMesh` for consistency:

```suggestion
func (r *Reconciler) mapMsaToMesh(_ context.Context, obj client.Object) []reconcile.Request {
```
(and update the `Watches` call site at line 131)

pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -106,3 +119,111 @@ func (r *Reconciler) ensureManagedServiceAccountUpdated(ctx context.Context, mes
...
+	return &corev1.Secret{
+		TypeMeta: metav1.TypeMeta{APIVersion: "v1", Kind: "Secret"},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        "istio-remote-secret-" + clusterName,
+			Namespace:   namespace,
+			Annotations: map[string]string{"networking.istio.io/cluster": clusterName},
+			Labels:      map[string]string{"istio/multiCluster": "true"},
+		},
+		Data: map[string][]byte{clusterName: buf.Bytes()},
+		Type: corev1.SecretTypeOpaque,
```
mkolesnik: nit: no need to set `Type` because `Opaque` is the default. 

```suggestion
```

test/integration/controller_test.go
```diff
@@ -1168,6 +1276,47 @@ func expectPlacement(meshName, meshNamespace string) *clusterv1beta1.Placement {
...
+	Expect(secret.Labels["istio/multiCluster"]).To(Equal("true"))
+	Expect(secret.Annotations["networking.istio.io/cluster"]).To(Equal(clusterName))
+	Expect(secret.Type).To(Equal(corev1.SecretTypeOpaque))
+	Expect(secret.Data).To(HaveKey(clusterName))
+}
+
+func expectMsaSecret(meshNamespace, meshName, clusterName string) {
+	msa := expectManagedServiceAccount(meshNamespace, meshName, clusterName)
+	util.CreateMsaSecret(ctx, k8sClient, msa.Name, clusterName)
+	util.SetMsaStatus(ctx, k8sClient, msa.Name, clusterName, time.Duration(0))
```
mkolesnik: nit: `time.Duration(0)` is just `0`, and `time.Duration(1*time.Minute)` on line 1316 is just `time.Minute`.

test/util/k8s.go
```diff
@@ -65,6 +65,33 @@ func CreateCacertsSecret(ctx context.Context, k8sClient client.Client, namespace
 	Expect(k8sClient.Create(ctx, secret)).To(Succeed())
 }
 
+// CreateMsaSecret creates a ServiceAccount and a secret that simulates what ManagedServiceAccount controller would create.
```
mkolesnik: nit: `CreateMsaSecret`, `UpdateMsaSecret` (k8s.go) and `SetMsaStatus` (ocm.go) are only used by the MWRS integration tests. Move to the test file as private helper methods.

test/util/k8s.go
```diff
@@ -65,6 +65,33 @@ func CreateCacertsSecret(ctx context.Context, k8sClient client.Client, namespace
...
+			Name:      msaName,
+			Namespace: clusterName,
+			Labels: map[string]string{
+				"authentication.open-cluster-management.io/is-managed-serviceaccount": "true",
+			},
+		},
+		Type: corev1.SecretTypeOpaque,
+		Data: map[string][]byte{
+			"ca.crt": []byte("test-ca-data"),
+			"token":  []byte("test-token-data"),
```
mkolesnik: nit: hardcoded `"ca.crt"` / `"token"` strings — use `corev1.ServiceAccountRootCAKey` / `corev1.ServiceAccountTokenKey` for consistency with MSA itself.

```suggestion
			corev1.ServiceAccountRootCAKey: []byte("test-ca-data"),
			corev1.ServiceAccountTokenKey:  []byte("test-token-data"),
```

pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -106,3 +119,111 @@ func (r *Reconciler) ensureManagedServiceAccountUpdated(ctx context.Context, mes
...
+		mwrset.Spec.PlacementRefs = []workv1alpha1.LocalPlacementReference{{Name: mesh.Name}}
+		mwrset.Spec.ManifestWorkTemplate = workv1.ManifestWorkSpec{Workload: workv1.ManifestsTemplate{Manifests: manifests}}
+		return controllerutil.SetControllerReference(mesh, mwrset, r.Scheme)
+	})
+
+	if err != nil {
+		return fmt.Errorf("failed to ensure ManifestWorkReplicaSet %s/%s: %w", mesh.Namespace, mesh.Name, err)
+	}
+
+	klog.Infof("Ensured ManifestWorkReplicaSet %s/%s state: %s", mesh.Namespace, mesh.Name, result)
```
mkolesnik: nit: this fires on every reconcile even when state is `unchanged`, adding noise.

```suggestion
	if result != controllerutil.OperationResultNone {
		klog.Infof("Ensured ManifestWorkReplicaSet %s/%s state: %s", mesh.Namespace, mesh.Name, result)
	}
```

pkg/hub/mesh/endpoint_discovery.go
yxun: https://github.com/stolostron/multicluster-mesh-addon/issues/253

test/integration/controller_test.go
yxun: I see. I removed this test case in last commit.

mkolesnik (commented):
Overall looks good to me, I have a few things I don't like with the tests but since we don't want to stall this any further I'll submit a PR to fix them.

@sridhargaddam PTAL and approve it if you agree

sridhargaddam (approved):

PR #216, opened 2026-07-13
Add e2e data-plane tests for multi-primary multi-network mesh topology

### LINKED ISSUE #187: Implement e2e data-plane tests for MultiClusterMesh addon
We need to implement end-to-end tests that validate the complete deployment and validation flow of the MultiClusterMesh addon, including control plane deployment on the hub and data plane validation with Istio CR creation on spoke clusters.

**Data Plane Validation**

- [ ] Create Istio Custom Resources on spoke clusters
- [ ] Validate the mesh control plane becomes ready
- [ ] Verify endpoint discovery works across clusters
- [ ] Test mTLS communication between services in different clusters

sridhargaddam:
This adds an end-to-end test that validates a full multi-primary, multi-network Istio deployment across two Kind spoke clusters. The test covers the complete flow: addon plumbing (operator + cacerts + MSA), Istio control plane installation via Sail Operator, east-west gateway setup using Gateway API, and cross-cluster traffic verification with the helloworld sample app.

Steps to execute the tests:
1. make dev-env
2. make test-e2e-multicluster

Fixes: https://github.com/stolostron/multicluster-mesh-addon/issues/187

Depends on: https://github.com/stolostron/multicluster-mesh-addon/pull/221


hack/dev-env.sh
```diff
@@ -276,6 +278,64 @@ install_managed_serviceaccount() {
...
+    local idx=0
+    for cluster in "${CLUSTER1}" "${CLUSTER2}"; do
+        local range_start="${base_prefix}.255.$((idx * 10 + 1))"
+        local range_end="${base_prefix}.255.$((idx * 10 + 10))"
+        idx=$((idx + 1))
+
+        if on "${cluster}" kubectl get deployment controller -n metallb-system &>/dev/null; then
+            log "MetalLB already installed on ${cluster}, skipping"
+            continue
+        fi
```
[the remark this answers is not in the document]
sridhargaddam: This adds more complexity to the code for an edge case. So we can revisit this in future if required.

hack/dev-env.sh
```diff
@@ -276,6 +278,64 @@ install_managed_serviceaccount() {
...
+    kind_subnet="$(docker network inspect kind -f '{{(index .IPAM.Config 0).Subnet}}' 2>/dev/null)" \
+        || err "Failed to determine Kind Docker network subnet. Is Docker running?"
+
+    local base_prefix
+    base_prefix="$(echo "${kind_subnet}" | cut -d'.' -f1-2)"
+
+    local idx=0
+    for cluster in "${CLUSTER1}" "${CLUSTER2}"; do
+        local range_start="${base_prefix}.255.$((idx * 10 + 1))"
+        local range_end="${base_prefix}.255.$((idx * 10 + 10))"
```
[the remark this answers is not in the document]
sridhargaddam: "/16" is used in the standard KIND deployments. We can revisit this in a future PR (if required) to support non standard deployments.

test/e2e/multi_primary_test.go
```diff
@@ -0,0 +1,317 @@
...
+					continue
+				}
+				targetClient := spokeClients[target]
+				secret := &corev1.Secret{}
+				Expect(targetClient.Get(ctx, key.Of("istio-remote-secret-"+source, cpNamespace), secret)).
+					To(Succeed(), "remote secret for %s not found on %s", source, target)
+				Expect(secret.Labels).To(HaveKeyWithValue("istio/multiCluster", "true"))
+			}
+		}
+	}, SpecTimeout(2*time.Minute))
```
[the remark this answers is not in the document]
sridhargaddam: The tests run sequentially and the Secret validation is already done in a previous step, so this additional validation is not really required.

hack/dev-env.sh
```diff
@@ -276,6 +278,67 @@ install_managed_serviceaccount() {
...
+    local kind_subnet
+    kind_subnet="$(docker network inspect kind -f '{{(index .IPAM.Config 0).Subnet}}' 2>/dev/null)" \
+        || err "Failed to determine Kind Docker network subnet. Is Docker running?"
+
+    local base_prefix
+    base_prefix="$(echo "${kind_subnet}" | cut -d'.' -f1-2)"
+
+    local idx=0
+    for cluster in "${CLUSTER1}" "${CLUSTER2}"; do
+        local range_start="${base_prefix}.255.$((idx * 10 + 1))"
```
yxun: nit: I was comparing this part with the upstream istio script. This PR is clear. If we need to make it consistent with upstream Istio script, we may reuse their function.
https://github.com/istio/istio/blob/master/common/scripts/kind_provisioner.sh#L463

test/e2e/testdata/helloworld.yaml
```diff
@@ -0,0 +1,29 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
```
yxun, line 3: Do you need sidecar injection label in the sample deployment ? I didn't find adding namespace sidecar injection label in the test.

test/e2e/multi_primary_test.go
```diff
@@ -0,0 +1,318 @@
...
+		for _, source := range clusters {
+			for _, target := range clusters {
+				if source == target {
+					continue
+				}
+				targetClient := spokeClients[target]
+				secret := &corev1.Secret{}
+				Expect(targetClient.Get(ctx, key.Of("istio-remote-secret-"+source, cpNamespace), secret)).
+					To(Succeed(), "remote secret for %s not found on %s", source, target)
+				Expect(secret.Labels).To(HaveKeyWithValue("istio/multiCluster", "true"))
```
yxun: We may also check secret has an annotation `"networking.istio.io/cluster": source` . This tells which cluster is the source for Istio.

hack/dev-env.sh
sridhargaddam: Since the current code is working fine, lets take it up in some future PR.

test/e2e/testdata/helloworld.yaml
sridhargaddam: Yes its already done in the multi_primary_test.go file Line 253

test/e2e/multi_primary_test.go
sridhargaddam: Sure, let me add it.

hack/dev-env.sh
```diff
@@ -316,6 +384,9 @@ case "${ACTION}" in
     init-ocm)                        init_ocm ;;
     join-clusters)                   join_clusters ;;
     setup-mesh)                      setup_mesh ;;
+    install-metallb)                 install_metallb ;;
+    install-gateway-api)             install_gateway_api ;;
+    clean)                           clean ;;
```
mkolesnik: The `clean` case is added here but I don't see a `clean` function defined in this PR..

hack/dev-env.sh
```diff
@@ -159,6 +159,13 @@ install_olm() {
     log "Granting klusterlet-work-sa OLM permissions on ${cluster}"
     on "${cluster}" kubectl apply -f "${SCRIPT_DIR}/hack/kind/klusterlet-work-olm.yaml"
 
+    for cluster in "${CLUSTER1}" "${CLUSTER2}"; do
+        log "Granting klusterlet-work-sa OLM permissions on ${cluster}"
+        on "${cluster}" kubectl apply -f "${SCRIPT_DIR}/hack/kind/klusterlet-work-olm.yaml"
```
mkolesnik: This is already applied on L160?

hack/dev-env.sh
```diff
@@ -271,6 +278,67 @@ install_managed_serviceaccount() {
...
+    local idx=0
+    for cluster in "${CLUSTER1}" "${CLUSTER2}"; do
+        local range_start="${base_prefix}.255.$((idx * 10 + 1))"
+        local range_end="${base_prefix}.255.$((idx * 10 + 10))"
+        idx=$((idx + 1))
+
+        if on "${cluster}" kubectl get deployment controller -n metallb-system &>/dev/null; then
+            log "MetalLB already installed on ${cluster}, skipping"
+            continue
+        fi
```
mkolesnik: This loops both clusters sequentially, but installing MetalLB on cluster1 and cluster2 is fully independent.
These are network-heavy operations (downloading manifests, waiting for rollouts).

Should parallelize across clusters the same way the dev-env does it.
Same applies to `install_gateway_api` below.

test/e2e/testdata/curl.yaml
```diff
@@ -0,0 +1,20 @@
...
+    matchLabels:
+      app: curl
+  template:
+    metadata:
+      labels:
+        app: curl
+    spec:
+      containers:
+      - name: curl
+        image: curlimages/curl:latest
```
mkolesnik: nit: pin to a specific version instead of `latest` for reproducible builds (e.g. `curlimages/curl:8.13.0`).

Makefile
```diff
@@ -249,9 +265,13 @@ $(CLUSTERADM): | $(BIN_DIR)
 		rm -rf $$tmp_dir; \
 	fi
 
+METALLB_VERSION ?= v0.14.9
+GATEWAY_API_VERSION ?= v1.2.1
```
mkolesnik: These should go with all the other dev-env versions...

hack/dev-env.sh
```diff
@@ -271,6 +278,67 @@ install_managed_serviceaccount() {
     on "${HUB}" kubectl get managedclusteraddon -A
 }
 
+install_metallb() {
+    require_clusters "${CLUSTER1}" "${CLUSTER2}"
+    local metallb_version="${METALLB_VERSION:-v0.14.9}"
+
+    local kind_subnet
+    kind_subnet="$(docker network inspect kind -f '{{(index .IPAM.Config 0).Subnet}}' 2>/dev/null)" \
```
mkolesnik: `docker network inspect kind` is hardcoded, but we should also support Podman.
On a Podman-only host this fails.
Also note that Podman uses `.Subnets` not `.IPAM.Config` in the network inspect format.

test/e2e/multi_primary_test.go
```diff
@@ -0,0 +1,319 @@
...
+
+		Step("Waiting for curl pod to be ready on cluster1")
+		curlPod := util.WaitForPodReady(ctx, spokeClients["cluster1"], sampleNS, map[string]string{"app": "curl"}, 2*time.Minute)
+		GinkgoWriter.Printf("Curl pod ready: %s\n", curlPod)
+
+		Step("Verifying cross-cluster traffic")
+		// sawV1/sawV2 are intentionally outside the Eventually closure so that we can evaluate if the test-code
+		// has seen response from both the helloworld applications.
+		var sawV1, sawV2 bool
+		Eventually(func(g Gomega) {
```
mkolesnik: These should probably live inside the `Eventually`:

```suggestion
		Eventually(func(g Gomega) {
			var sawV1, sawV2 bool
```

test/e2e/multi_primary_test.go
```diff
@@ -0,0 +1,319 @@
...
+
+	BeforeAll(func(ctx SpecContext) {
+		meshNS = util.UniqueName("mp-test-ns")
+		util.CreateNamespace(ctx, hubClient, meshNS)
+
+		Step("Setting up cert-manager trust chain in %s", meshNS)
+		util.LoadAndApplyYAMLInNamespace(ctx, hubClient, filepath.Join(samplesDir, "cert-manager-issuer.yaml"), meshNS, nil)
+	})
+
+	AfterAll(func(ctx SpecContext) {
```
mkolesnik: This 45-line block manually reverses every `It`'s setup and must stay in
sync with all of them. 

You can simplify it:
- Cleanup verification ("should clean up on deletion") is already tested
  by the lifecycle suite. Don't retest it here.
- Record YAML paths as you apply them, then delete them all in AfterAll.
- Delete the mesh with `util.DeleteResource` (waits for completion).
- Delete the test namespace.

A simple recorder pattern:
```go
var appliedYAMLs []struct{ client client.Client; path string; vars map[string]string }
// ... in each apply call, append to appliedYAMLs ...
// AfterAll: reverse-iterate and delete each one
```

test/e2e/multi_primary_test.go
```diff
@@ -0,0 +1,319 @@
...
+	samplesDir  string
+)
+
+func init() {
+	_, thisFile, _, _ := runtime.Caller(0)
+	testdataDir = filepath.Join(filepath.Dir(thisFile), "testdata")
+	samplesDir = filepath.Join(filepath.Dir(thisFile), "..", "..", "samples")
+}
+
+var _ = Describe("Multi-primary multi-network mesh", Ordered, Serial, func() {
```
mkolesnik: Suggested structure:
```
Describe("Multi-primary data plane", Ordered, Serial)
  BeforeAll:
    - create mesh ns, create mesh CR, wait for Ready
    - verify istio-system exists with network labels (sanity)

  When("Istio multi-cluster is deployed")
    BeforeAll:
      - apply IstioCNI, Istio CRs, wait for istiod
      - apply east-west gateway, wait for LB IP

    It("should distribute remote secrets to spoke clusters")
    It("should have cross-cluster data plane traffic")

  AfterAll: clean up created resources
```
Operator MWs, cacerts MWs, MSA tokens are tested by the lifecycle suite.
Istio/gateway setup is infrastructure, not a test.

test/e2e/multi_primary_test.go
```diff
@@ -0,0 +1,319 @@
...
+						},
+					},
+				},
+			})
+
+		Step("Waiting for controller to reconcile the mesh")
+		Eventually(func(g Gomega) {
+			g.Expect(hubClient.Get(ctx, key.For(mesh), mesh)).To(Succeed())
+			g.Expect(meta.FindStatusCondition(mesh.Status.Conditions, meshv1alpha1.ConditionReady)).NotTo(BeNil())
+		}).WithTimeout(30 * time.Second).Should(Succeed())
```
mkolesnik: Move mesh creation and readiness wait to `BeforeAll`. This is setup
for the data plane tests, not a test concern. The lifecycle suite
already covers mesh deployment behavior.

In `BeforeAll`, after mesh is ready, add a sanity check that
`istio-system` exists with network labels (the controller creates it):
```go
for _, spokeClient := range spokeClients {
    ns := &corev1.Namespace{}
    Eventually(func(g Gomega) {
        g.Expect(spokeClient.Get(ctx, key.Of("istio-system"), ns)).To(Succeed())
        g.Expect(ns.Labels).To(HaveKey("topology.istio.io/network"))
    }).Should(Succeed())
}
```

test/e2e/multi_primary_test.go
```diff
@@ -0,0 +1,319 @@
...
+	var (
+		mesh     *meshv1alpha1.MultiClusterMesh
+		meshNS   string
+		networks = map[string]string{
+			"cluster1": "network1",
+			"cluster2": "network2",
+		}
+	)
+
+	BeforeAll(func(ctx SpecContext) {
```
mkolesnik, line 56: Unless there's a reason to set this up per test run, the cert-manager trust
chain feels like infrastructure setup, same category as MetalLB and Gateway
API CRDs (Makefile targets / dev-env.sh actions).

`setup_mesh` in dev-env.sh already does the same thing (L355-357).

Consider moving to a Makefile prerequisite of `test-e2e-multicluster` so the
test starts from "trust chain exists" and focuses on mesh behavior.

test/e2e/suite_test.go
```diff
@@ -28,9 +29,20 @@ import (
...
+)
+
 var (
+	clusters = []string{"cluster1", "cluster2"}
+
 	hubClient    client.Client
 	spokeClients map[string]client.Client
+	// spokeConfigs holds the raw REST configs needed for pod exec, which client.Client does not support.
+	spokeConfigs map[string]*rest.Config
 )
```
mkolesnik, line 42: `spokeClients` and `spokeConfigs` exist because some operations need `client.Client`
and others need the kubeconfig path. A thin wrapper solves this:
```go
type E2EClient struct {
    client.Client
    kubeconfig string
}

func NewE2EClient(kubeconfig string) *E2EClient {
    c := clientFrom(kubeconfig)
    return &E2EClient{Client: c, kubeconfig: kubeconfig}
}

func (c *E2EClient) Exec(ctx context.Context, ns, pod, container string, cmd []string) (string, error) {
    args := []string{"exec", "-n", ns, pod, "-c", container, "--kubeconfig", c.kubeconfig, "--"}
    args = append(args, cmd...)
    // exec.CommandContext(ctx, "kubectl", args...)
}

func (c *E2EClient) ApplyFile(ctx context.Context, path string, vars map[string]string) {
    rendered := renderYAML(path, vars)
    // kubectl apply --kubeconfig c.kubeconfig -f -
}

func (c *E2EClient) DeleteFile(ctx context.Context, path string, vars map[string]string) {
    // kubectl delete --kubeconfig c.kubeconfig -f -
}
```
Existing code (`spokeClient.Get`, `spokeClient.List`) works unchanged via embedding.
New code gets `spokeClient.Exec(...)`, `spokeClient.ApplyFile(...)`.
Eliminates `spokeConfigs`, the complex SPDY machinery, and the unstructured SSA path.

test/util/apply.go
```diff
@@ -0,0 +1,213 @@
...
+
+func EnsureNamespace(ctx context.Context, k8sClient client.Client, name string) {
+	ns := &corev1.Namespace{}
+	err := k8sClient.Get(ctx, types.NamespacedName{Name: name}, ns)
+	if err == nil {
+		return
+	}
+	Expect(client.IgnoreNotFound(err)).To(Succeed())
+	CreateNamespace(ctx, k8sClient, name)
+}
```
mkolesnik: SSA + unstructured + FieldOwner + ForceOwnership + nolint for a namespace is overkill.
Extend the existing `CreateNamespace` in `k8s.go` instead:
```go
func CreateNamespace(ctx context.Context, k8sClient client.Client, name string, labels ...map[string]string) {
    ns := &corev1.Namespace{
        ObjectMeta: metav1.ObjectMeta{Name: name},
    }
    if len(labels) > 0 && labels[0] != nil {
        ns.Labels = labels[0]
    }
    Expect(client.IgnoreAlreadyExists(k8sClient.Create(ctx, ns))).To(Succeed())
}
```
Existing callers don't change. New callers pass labels.
Alternatively, make `labels` a required `map[string]string` parameter and pass `nil` from existing callers.

`EnsureNamespace` is called once and becomes unnecessary. Remove it.

test/util/apply.go
```diff
@@ -0,0 +1,213 @@
...
+	var stdout, stderr bytes.Buffer
+	err = exec.StreamWithContext(ctx, remotecommand.StreamOptions{
+		Stdout: &stdout,
+		Stderr: &stderr,
+	})
+	if err != nil {
+		return "", fmt.Errorf("exec failed (stderr: %s): %w", strings.TrimSpace(stderr.String()), err)
+	}
+
+	return stdout.String(), nil
```
mkolesnik: This pulls in SPDY machinery + 3 new indirect deps (`gorilla/websocket`, `moby/spdystream`, `mxk/go-flowrate`) for something `kubectl exec` does out of the box.

For the E2E tests it will be just easier to use kubectl directly, eg:
```go
func ExecInPod(kubeconfig, namespace, podName, container string, command []string) (string, error) {
    args := []string{"exec", "-n", namespace, podName, "-c", container,
        "--kubeconfig", kubeconfig, "--"}
    args = append(args, command...)
    cmd := exec.Command("kubectl", args...)
    var stdout, stderr bytes.Buffer
    cmd.Stdout = &stdout
    cmd.Stderr = &stderr
    if err := cmd.Run(); err != nil {
        return "", fmt.Errorf("exec failed (stderr: %s): %w",
            strings.TrimSpace(stderr.String()), err)
    }
    return stdout.String(), nil
}
```
Also this should be in the proposed `E2EClient` class which will simplifyu all the calls even further.

test/e2e/multi_primary_test.go
```diff
@@ -0,0 +1,319 @@
...
+						meshcontroller.MeshNamespaceLabel: mesh.Namespace,
+					},
+				)).To(Succeed())
+				g.Expect(msaList.Items).To(HaveLen(1))
+				g.Expect(msaList.Items[0].Status.TokenSecretRef).NotTo(BeNil())
+			}).WithTimeout(2 * time.Minute).Should(Succeed())
+		}
+	}, SpecTimeout(5*time.Minute))
+
+	It("Istio control planes become ready on both clusters", func(ctx SpecContext) {
```
mkolesnik: This is infrastructure setup for the data plane tests, not test behavior.
Wrap in a `When("Istio multi-cluster is deployed")` context with this as
`BeforeAll`.

Also, there's no need to create `istio-system` (controller already did - #220).

test/e2e/multi_primary_test.go
```diff
@@ -0,0 +1,319 @@
...
+				"MeshID":      meshID,
+				"ClusterName": cluster,
+				"Network":     networks[cluster],
+			})
+		}
+
+		for cluster, spokeClient := range spokeClients {
+			Step("Waiting for istiod to be ready on %s", cluster)
+			util.WaitForDeploymentReady(ctx, spokeClient, "istiod", cpNamespace, 5*time.Minute)
+			GinkgoWriter.Printf("istiod is ready on %s\n", cluster)
```
mkolesnik: You can create a simple `Success` wrapper for cases when we want to report success, same as sail operator's:

```go
func Success(message string) {
	g.GinkgoWriter.Println("* " + message)
}
```

hack/kind/klusterlet-work-istio.yaml
```diff
@@ -0,0 +1,25 @@
...
+# On OpenShift, klusterlet-work-sa gets cluster-admin during import. However, on
+# Kind clusters joined via clusteradm, only minimal permissions are granted.
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: klusterlet-work-istio
+rules:
+  - apiGroups: [""]
+    resources: ["secrets"]
+    verbs: ["create", "get", "list", "update", "patch", "delete"]
```
mkolesnik: This grants cluster-wide Secret access, but cacerts only go to `istio-system`.
A namespaced Role + RoleBinding scoped to `istio-system` would be more precise.

If there's a concrete reason to keep it cluster-wide (other namespaces will need secrets), document it.
Otherwise, scope it down.

hack/kind/metallb-pool.yaml
mkolesnik, line 1: nit: this is a dev-env template (has `__ADDRESS_RANGE__` placeholder), not a user-facing sample.
Move to `hack/kind/metallb-pool.yaml` alongside `klusterlet-work-olm.yaml` and `klusterlet-work-istio.yaml`.

hack/dev-env.sh
```diff
@@ -271,6 +278,67 @@ install_managed_serviceaccount() {
...
+        || err "Failed to determine Kind Docker network subnet. Is Docker running?"
+
+    local base_prefix
+    base_prefix="$(echo "${kind_subnet}" | cut -d'.' -f1-2)"
+
+    local idx=0
+    for cluster in "${CLUSTER1}" "${CLUSTER2}"; do
```
mkolesnik: This loops both clusters sequentially, but installing MetalLB on cluster1 and cluster2 is fully independent.
These are network-heavy operations (downloading manifests, waiting for rollouts).

Should parallelize across clusters the same way the dev-env does it.
Same applies to `install_gateway_api` below.

test/e2e/multi_primary_test.go
mkolesnik, line 1: Every `Eventually` in this file overrides the timeout.
If 30s is too short for multicluster tests, let's change the default for this suite rather than overriding per-assertion.

If you want to affect only these tests you can set it at the `Describe` level, and it should be fine since they use a different build tag than the lifecycle e2e tests, but I think setting 2m on global suite level should be enough for most calls on local kind cluster.

If and when external cluster running will be supported (eg on OCP) then we can get the default from the env vars.

test/e2e/multi_primary_test.go
```diff
@@ -0,0 +1,319 @@
...
+				if sawV1 && sawV2 {
+					break
+				}
+			}
+			g.Expect(sawV1).To(BeTrue(), "never saw response from helloworld v1")
+			g.Expect(sawV2).To(BeTrue(), "never saw response from helloworld v2")
+		}).WithTimeout(3 * time.Minute).WithPolling(10 * time.Second).Should(Succeed())
+
+		GinkgoWriter.Println("Cross-cluster traffic verified: saw responses from both v1 and v2")
+	}, SpecTimeout(8*time.Minute))
```
mkolesnik: Extract the common template vars and group the deployments:
```go
sampleVars := map[string]string{"Namespace": sampleNS}
```
Then deploy all YAML resources together in one block instead of spreading
across individual steps.

If you want per-resource logging, use a local helper that wraps `Step` + `Apply`.

test/util/apply.go
```diff
@@ -0,0 +1,213 @@
...
+			if c.Type == appsv1.DeploymentAvailable {
+				g.Expect(c.Status).To(Equal(corev1.ConditionTrue),
+					"deployment %s/%s is not Available: %s", namespace, name, c.Message)
+				found = true
+				break
+			}
+		}
+		g.Expect(found).To(BeTrue(), "deployment %s/%s has no Available condition", namespace, name)
+	}).WithTimeout(timeout).WithPolling(2 * time.Second).Should(Succeed())
+}
```
mkolesnik: Hand-rolled condition loop. `mesh_lifecycle_test.go` does the same check
with Gomega matchers:
```go
Expect(deploy.Status.Conditions).To(ContainElement(And(
    HaveField("Type", appsv1.DeploymentAvailable),
    HaveField("Status", corev1.ConditionTrue),
)))
```

It's used 4 times so a helper is fine, but it belongs in `multi_primary_test.go`
(its only consumer), not `test/util/`.

test/util/apply.go
```diff
@@ -0,0 +1,213 @@
...
+					if c.Type == corev1.PodReady && c.Status == corev1.ConditionTrue {
+						podName = pod.Name
+						return
+					}
+				}
+			}
+		}
+		g.Expect(false).To(BeTrue(), "no ready pod found with labels %v in %s", labels, namespace)
+	}).WithTimeout(timeout).WithPolling(2 * time.Second).Should(Succeed())
+	return podName
```
mkolesnik: Called once (curl pod). Inline to the test.

test/util/apply.go
```diff
@@ -0,0 +1,213 @@
...
+			if pod.Status.Phase == corev1.PodRunning {
+				for _, c := range pod.Status.Conditions {
+					if c.Type == corev1.PodReady && c.Status == corev1.ConditionTrue {
+						podName = pod.Name
+						return
+					}
+				}
+			}
+		}
+		g.Expect(false).To(BeTrue(), "no ready pod found with labels %v in %s", labels, namespace)
```
mkolesnik: `g.Expect(false).To(BeTrue())` to force a failure is roundabout.
Use `g.Fail("no ready pod found...")` or restructure as a positive assertion.

test/util/apply.go
```diff
@@ -0,0 +1,213 @@
...
+		if obj.Object == nil {
+			continue
+		}
+		if defaultNS != "" && obj.GetNamespace() == "" {
+			obj.SetNamespace(defaultNS)
+		}
+		Expect(k8sClient.Patch(ctx, obj, client.Apply, client.FieldOwner("e2e-test"), client.ForceOwnership)). //nolint:staticcheck // client.Apply is the only way to SSA with unstructured objects
+															To(Succeed(), "failed to apply %s %s/%s", obj.GetKind(), obj.GetNamespace(), obj.GetName())
+	}
+}
```
mkolesnik: The unstructured decode + SSA loop is a lot of machinery for e2e tests. `kubectl` is already available in the dev-env. Keep the Go template rendering, pipe to kubectl:
```go
func ApplyYAML(ctx context.Context, kubeconfig, path string, vars map[string]string) {
    rendered := renderYAML(path, vars)
    cmd := exec.CommandContext(ctx, "kubectl", "apply",
        "--kubeconfig", kubeconfig, "-f", "-")
    cmd.Stdin = bytes.NewReader(rendered)
    var stderr bytes.Buffer
    cmd.Stderr = &stderr
    Expect(cmd.Run()).To(Succeed(),
        "failed to apply %s: %s", path, stderr.String())
}
```
Same pattern for delete (`kubectl delete --kubeconfig <path> -f -`).
The `InNamespace` variants become `kubectl apply -n <ns>`.

This eliminates the `unstructured` import, the EOF-loop decoder, and the SSA patch options.

Also keep in mind this can be part of the proposed `E2EClient` helper to further simplify the code.

test/e2e/testdata/helloworld-v2.yaml
mkolesnik, line 1: nit: near-identical to `helloworld-v1.yaml` (only `v1`/`v2` differs in name, labels, image)..

Since you're already templating `{{.Namespace}}`, merge into one `helloworld.yaml` with `{{.Version}}`:
```yaml
name: helloworld-{{.Version}}
image: quay.io/sail-dev/examples-helloworld-{{.Version}}
```

test/e2e/testdata/helloworld-v1.yaml
```diff
@@ -0,0 +1,29 @@
...
+      version: v1
+  template:
+    metadata:
+      labels:
+        app: helloworld
+        version: v1
+    spec:
+      containers:
+      - name: helloworld
+        image: quay.io/sail-dev/examples-helloworld-v1
```
mkolesnik: nit: no tag on the image, defaults to `:latest`.
Pin to a specific version for reproducibility.

test/e2e/multi_primary_test.go
```diff
@@ -0,0 +1,319 @@
...
+					continue
+				}
+				targetClient := spokeClients[target]
+				secret := &corev1.Secret{}
+				Expect(targetClient.Get(ctx, key.Of("istio-remote-secret-"+source, cpNamespace), secret)).
+					To(Succeed(), "remote secret for %s not found on %s", source, target)
+				Expect(secret.Labels).To(HaveKeyWithValue("istio/multiCluster", "true"))
+				Expect(secret.Annotations).To(HaveKeyWithValue("networking.istio.io/cluster", source))
+			}
+		}
```
mkolesnik: nit: iterate `spokeClients` directly for targets:
```go
for source := range spokeClients {
    for target, targetClient := range spokeClients {
        if source == target {
            continue
        }
        secret := &corev1.Secret{}
        Expect(targetClient.Get(ctx, key.Of("istio-remote-secret-"+source, cpNamespace), secret)).
            To(Succeed(), "remote secret for %s not found on %s", source, target)
        Expect(secret.Labels).To(HaveKeyWithValue("istio/multiCluster", "true"))
        Expect(secret.Annotations).To(HaveKeyWithValue("networking.istio.io/cluster", source))
    }
}
```

hack/dev-env.sh
```diff
@@ -271,6 +278,67 @@ install_managed_serviceaccount() {
     on "${HUB}" kubectl get managedclusteraddon -A
 }
 
+install_metallb() {
+    require_clusters "${CLUSTER1}" "${CLUSTER2}"
+    local metallb_version="${METALLB_VERSION:-v0.14.9}"
```
mkolesnik: nit: the Makefile already exports `METALLB_VERSION` and `GATEWAY_API_VERSION`.
This default duplicates the Makefile value and differs from how other functions
use their versions (e.g. `${OLM_VERSION}` directly, no local default).

Use `${METALLB_VERSION}` directly, same as the rest of the script.
Same applies to `gw_api_version` at L323.

test/util/apply.go
```diff
@@ -0,0 +1,213 @@
...
+	Eventually(func(g Gomega) {
+		svc := &corev1.Service{}
+		g.Expect(k8sClient.Get(ctx, key.Of(name, namespace), svc)).To(Succeed())
+		g.Expect(svc.Status.LoadBalancer.Ingress).NotTo(BeEmpty(),
+			"service %s/%s has no LoadBalancer ingress", namespace, name)
+		ip = svc.Status.LoadBalancer.Ingress[0].IP
+		if ip == "" {
+			ip = svc.Status.LoadBalancer.Ingress[0].Hostname
+		}
+		g.Expect(ip).NotTo(BeEmpty(), "service %s/%s has no IP or hostname", namespace, name)
```
mkolesnik: Use `address`:

```suggestion
		addess = svc.Status.LoadBalancer.Ingress[0].IP
		if addess == "" {
			addess = svc.Status.LoadBalancer.Ingress[0].Hostname
		}
		g.Expect(addess).NotTo(BeEmpty(), "service %s/%s has no IP or hostname", namespace, name)
```

mkolesnik (changes requested):

hack/dev-env.sh
sridhargaddam: Done, updated in the previous commit.

test/e2e/testdata/curl.yaml
sridhargaddam: Done

test/util/apply.go
sridhargaddam: Gomega (g) does not seem to have the Fail method, so let me change it to a positive assertion.

test/util/apply.go
sridhargaddam: > It's used 4 times so a helper is fine, but it belongs in multi_primary_test.go
(its only consumer), not test/util/

Addressed the first part, but I'm not convinced that we should move it to `multi_primary_test.go`. This API is a generic utility which future tests might likely need as well and also the /test/util package already has similar methods, so having it in the `test/util/` package seems to be more appropriate even thought its used only my the multi_primary tests.

test/util/apply.go
sridhargaddam: It's currently called once, but it sits alongside `WaitForDeploymentReady` and `WaitForLoadBalancerIP` as a set of helpers. Keeping them together in `test/util/` makes the test file cleaner and avoids moving it back when a second consumer appears.

hack/kind/klusterlet-work-istio.yaml
mkolesnik, line 1: This file isn't needed.
The work agent already gets the `admin` ClusterRole via `clusteradm join` (bound as `execution-admin`), which already includes full Secret CRUD.

test/e2e/suite_test.go
sridhargaddam: The E2EClient wrapper is a good idea but it would be a major refactor that touches all the test files. Let's track it as a follow-up to keep this PR focused on the test coverage. I reported the following issue for tracking purposes https://github.com/stolostron/multicluster-mesh-addon/issues/225

test/util/apply.go
sridhargaddam: Tracked in https://github.com/stolostron/multicluster-mesh-addon/issues/225

test/util/apply.go
sridhargaddam: https://github.com/stolostron/multicluster-mesh-addon/issues/225

hack/kind/klusterlet-work-istio.yaml
sridhargaddam: hmm...I remember running into issues when i initially tried validating the PR. Probably it was some timing issue. Let me remove it.

test/e2e/multi_primary_test.go
sridhargaddam: As explained in the comment above, "sawV1/sawV2 are intentionally outside the Eventually closure so that we can evaluate if the test-code has seen response from both the helloworld applications." If they were inside, the variables would reset on every retry.

test/e2e/multi_primary_test.go
sridhargaddam: The `meshNs` is unique for each test run. As you can see in L58, the test code generates a random namespace each time. The cert-manager `Issuer` and `Certificate` resources are created in this per-run namespace. If we moved this to a Makefile target, we would need to use a fixed namespace, which could cause repeated test runs to conflict with leftover resources and would make the test less self-contained and isolated.

Regarding `setup_mesh`, it uses a fixed `mesh-system` namespace, which is fine for the one-time setup in the dev environment.

hack/kind/klusterlet-work-istio.yaml
sridhargaddam: This file will be deleted.

test/e2e/multi_primary_test.go
sridhargaddam: IMO this is more readable (and it works). Please feel free to refactor this part in a follow on PR if you have strong opinions on the flow.

test/e2e/suite_test.go
mkolesnik: I'm glad you agree, but I don't think it's a separate issue.

In fact you're introducing a new mechanism to apply/exec in this PR already, I'm merely suggesting to use a client wrapper instead of the bare client.

Existing code will remain unchanged AFAICT, so I don't think it'll be a big refactor, but let me know WDYT.

sridhargaddam: Okay, let me give it a try.

test/util/apply.go
```diff
@@ -0,0 +1,98 @@
...
+
+	. "github.com/onsi/gomega"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/stolostron/multicluster-mesh-addon/pkg/key"
+)
+
+func renderYAML(path string, vars map[string]string) []byte {
```
mkolesnik: Please move to e2e_client.go file next to where it's actually used..

test/util/apply.go
```diff
@@ -0,0 +1,98 @@
...
+		return data
+	}
+	tmpl, err := template.New("manifest").Parse(string(data))
+	Expect(err).NotTo(HaveOccurred(), "failed to parse template %s", path)
+	var buf bytes.Buffer
+	Expect(tmpl.Execute(&buf, vars)).To(Succeed(), "failed to execute template %s", path)
+	return buf.Bytes()
+}
+
+func WaitForDeploymentReady(ctx context.Context, k8sClient client.Client, name, namespace string, timeout ...time.Duration) {
```
mkolesnik: All these `Wait*` funcs should live in the data plane e2e test file (as private).

Then you can delete this file (it's name doesn't make sense anyway)

test/util/e2e_client.go
```diff
@@ -0,0 +1,87 @@
...
+	"strings"
+
+	. "github.com/onsi/gomega"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// E2EClient wraps client.Client with a kubeconfig path, providing kubectl-based
+// operations (like pod exec, apply, delete) alongside the standard controller-runtime client.
+// ApplyFile calls are automatically recorded so Cleanup can reverse-delete them.
+type E2EClient struct {
```
mkolesnik, line 19: Thanks I really like this design, it's much simpler and more elegant that the previous way :muscle:

test/util/e2e_client.go
```diff
@@ -0,0 +1,87 @@
...
+	path      string
+	vars      map[string]string
+	namespace string
+}
+
+func NewE2EClient(c client.Client, kubeconfig string) *E2EClient {
+	return &E2EClient{Client: c, Kubeconfig: kubeconfig}
+}
+
+// ApplyFile renders a YAML template and applies it via kubectl. // The call is recorded so that Cleanup can reverse-delete all applied files.
```
mkolesnik, line 35: nit: Two comment sentences merged onto one line.
```suggestion
// ApplyFile renders a YAML template and applies it via kubectl.
// The call is recorded so that Cleanup can reverse-delete all applied files.
```

test/e2e/multi_primary_test.go
mkolesnik: Yes I know the mesh ns is unique (as it should be) and contains the resources either the test created or the controller..

I just think the cert-manager is not a test concern per-se..

Tell me WDYT of an approach of using a `ClusterIssuer` instead, same as I proposed in https://github.com/stolostron/multicluster-mesh-addon/pull/226/files#diff-cfd7d916f92e259933dd531eb3c0870a555274cd614ae5c2aa25baa1aa3cca5c

This way it's out of the test concern and we only create it once (also useful for the regular lifecycle tests, as you can see).

Appreciate your opinion :pray:

test/util/apply.go
```diff
@@ -0,0 +1,98 @@
...
+			}
+		}
+		g.Expect(found).To(BeTrue(), "no ready pod found with labels %v in %s", labels, namespace)
+	}).WithPolling(2 * time.Second)
+	if len(timeout) > 0 {
+		e = e.WithTimeout(timeout[0])
+	}
+	e.Should(Succeed())
+	return podName
+}
```
mkolesnik: The nested loop with `found` sentinel and double-break is hard to follow.
The `PodRunning` phase check is also redundant (a pod with `Ready=True` is necessarily Running).

Extract the search to a helper:
```suggestion
func WaitForPodReady(ctx context.Context, k8sClient client.Client, namespace string, labels map[string]string, timeout ...time.Duration) string {
	var podName string
	e := Eventually(func(g Gomega) {
		pods := &corev1.PodList{}
		g.Expect(k8sClient.List(ctx, pods, client.InNamespace(namespace), client.MatchingLabels(labels))).To(Succeed())
		g.Expect(pods.Items).NotTo(BeEmpty(), "no pods found with labels %v in %s", labels, namespace)
		podName = firstReadyPod(pods.Items)
		g.Expect(podName).NotTo(BeEmpty(), "no ready pod found with labels %v in %s", labels, namespace)
	}).WithPolling(2 * time.Second)
	if len(timeout) > 0 {
		e = e.WithTimeout(timeout[0])
	}
	e.Should(Succeed())
	return podName
}

func firstReadyPod(pods []corev1.Pod) string {
	for i := range pods {
		for _, c := range pods[i].Status.Conditions {
			if c.Type == corev1.PodReady && c.Status == corev1.ConditionTrue {
				return pods[i].Name
			}
		}
	}
	return ""
}
```

test/e2e/multi_primary_test.go
```diff
@@ -0,0 +1,259 @@
...
+		}).Should(Succeed())
+
+		Step("Verifying control plane namespace exists with network labels")
+		for _, spokeClient := range spokeClients {
+			ns := &corev1.Namespace{}
+			Eventually(func(g Gomega) {
+				g.Expect(spokeClient.Get(ctx, key.Of(cpNamespace), ns)).To(Succeed())
+				g.Expect(ns.Labels).To(HaveKey("topology.istio.io/network"))
+			}).Should(Succeed())
+		}
```
mkolesnik, line 100: The mesh is already `Ready` above, which means the CP namespace MW was applied and should be available.
This is re-verifying what the lifecycle tests already cover.

I'd drop it entirely..
```suggestion
```

If you want it as a sanity check, use a plain `Expect` since there's no race to wait out:
```suggestion
	Step("Verifying control plane namespace exists with network labels")
	for _, spokeClient := range spokeClients {
		ns := &corev1.Namespace{}
		Expect(spokeClient.Get(ctx, key.Of(cpNamespace), ns)).To(Succeed())
		Expect(ns.Labels).To(HaveKey("topology.istio.io/network"))
	}
```

test/e2e/multi_primary_test.go
```diff
@@ -0,0 +1,259 @@
...
+		BeforeAll(func(ctx SpecContext) {
+			for cluster, spokeClient := range spokeClients {
+				Step("Ensuring istio-cni namespace on %s", cluster)
+				util.CreateNamespace(ctx, spokeClient, "istio-cni")
+
+				Step("Applying IstioCNI CR on %s", cluster)
+				spokeClient.ApplyFile(ctx, filepath.Join(testdataDir, "istiocni-cr.yaml"), nil)
+
+				Step("Applying Istio CR on %s", cluster)
+				spokeClient.ApplyFile(ctx, filepath.Join(testdataDir, "istio-cr.yaml"), map[string]string{
```
mkolesnik, line 134: `CRName`, `TrustDomain`, and `MeshID` are constants defined above and never vary.

Hardcode them in the YAML:
```go
spokeClient.ApplyFile(ctx, filepath.Join(testdataDir, "istio-cr.yaml"), map[string]string{
    "CPNamespace": cpNamespace,
    "ClusterName": cluster,
    "Network":     networks[cluster],
})
```

Same for `helloworld-service.yaml` and `curl.yaml`: drop the `Namespace` template var and use `ApplyFile`'s namespace parameter instead (`spokeClient.ApplyFile(ctx, path, nil, sampleNS)`).

test/e2e/multi_primary_test.go
```diff
@@ -0,0 +1,259 @@
...
+			Step("Waiting for helloworld-v2 to be ready on cluster2")
+			util.WaitForDeploymentReady(ctx, spokeClients["cluster2"], "helloworld-v2", sampleNS)
+
+			Step("Deploying curl pod on cluster1")
+			spokeClients["cluster1"].ApplyFile(ctx, filepath.Join(testdataDir, "curl.yaml"), map[string]string{
+				"Namespace": sampleNS,
+			})
+
+			Step("Waiting for curl pod to be ready on cluster1")
+			curlPod := util.WaitForPodReady(ctx, spokeClients["cluster1"], sampleNS, map[string]string{"app": "curl"})
```
mkolesnik: The curl Deployment forces label-based pod discovery via `WaitForPodReady`. 

Two ways to avoid it:

**Option A**: Change curl.yaml to a bare `Pod`.
The name is fixed (`curl`), wait with a simpler wait, exec by name.
No label discovery needed.

**Option B**: Keep the Deployment but use `deployment/curl` as the exec target - `kubectl` resolves the pod for you:
```go
spokeClients["cluster1"].Exec(ctx, sampleNS, "deployment/curl", "curl", ...)
```

Then `WaitForPodReady` can be dropped and replaced with `WaitForDeploymentReady`.

test/e2e/multi_primary_test.go
```diff
@@ -0,0 +1,259 @@
...
+				"Version":   "v2",
+			})
+
+			Step("Waiting for helloworld-v1 to be ready on cluster1")
+			util.WaitForDeploymentReady(ctx, spokeClients["cluster1"], "helloworld-v1", sampleNS)
+
+			Step("Waiting for helloworld-v2 to be ready on cluster2")
+			util.WaitForDeploymentReady(ctx, spokeClients["cluster2"], "helloworld-v2", sampleNS)
+
+			Step("Deploying curl pod on cluster1")
```
mkolesnik: nit: Move the curl deployment up before the `WaitForDeploymentReady` calls so the image pull overlaps with helloworld startup and test is a bit faster.
The curl pod only needs the namespace, not helloworld running.

test/e2e/multi_primary_test.go
```diff
@@ -0,0 +1,259 @@
...
+					if strings.Contains(output, "v2") {
+						sawV2 = true
+					}
+					if sawV1 && sawV2 {
+						break
+					}
+				}
+				g.Expect(sawV1).To(BeTrue(), "never saw response from helloworld v1")
+				g.Expect(sawV2).To(BeTrue(), "never saw response from helloworld v2")
+			}).WithTimeout(3 * time.Minute).WithPolling(10 * time.Second).Should(Succeed())
```
mkolesnik: nit: Three simplifications:
1. **Drop inner loop**: the `for i := 0; i < 20` duplicates what Eventually already does (retry with timeout). 
   One curl per Eventually iteration with 1s polling is cleaner, and the magic `20` goes away.
2. **`curl -sS --fail`**: `-sS` stays silent but shows errors, `--fail` returns non-zero on HTTP errors (e.g. 503 during startup) so they're caught by `if err != nil` instead of silently producing a body that doesn't match v1/v2.
4. **`||` for flag tracking**: `sawV1 = sawV1 || strings.Contains(...)` replaces the 3-line if/assign block.

```suggestion
		Eventually(func(g Gomega) {
			output, err := spokeClients["cluster1"].Exec(ctx,
				sampleNS, curlPod, "curl",
				[]string{"curl", "-sS", "--fail", fmt.Sprintf("helloworld.%s:5000/hello", sampleNS)})
			if err != nil {
				return
			}
			sawV1 = sawV1 || strings.Contains(output, "v1")
			sawV2 = sawV2 || strings.Contains(output, "v2")
			g.Expect(sawV1).To(BeTrue(), "never saw response from helloworld v1")
			g.Expect(sawV2).To(BeTrue(), "never saw response from helloworld v2")
		}).WithTimeout(3 * time.Minute).WithPolling(1 * time.Second).Should(Succeed())
```

mkolesnik (commented):
Thanks Sridhar, a few comments and a question

test/e2e/multi_primary_test.go
sridhargaddam: As you know, the controller will support both `ClusterIssuer` and namespaced `Issuer`, so either approach works. One of my earlier PRs actually used `ClusterIssuer`, but it was later changed to a namespaced `Issuer` in PR#[142](https://github.com/stolostron/multicluster-mesh-addon/pull/142/). I don't have a strong preference. I just don't think it's worth blocking this PR over, since it could be a **follow-up change once this PR merges**. Hope that works for you.

sridhargaddam (commented):
Thanks for reviewing.

test/e2e/multi_primary_test.go
```diff
@@ -0,0 +1,270 @@
...
+					sampleNS, "deployment/curl", "curl",
+					[]string{"curl", "-sS", "--fail", fmt.Sprintf("helloworld.%s:5000/hello", sampleNS)})
+				if err != nil {
+					return
+				}
+				sawV1 = sawV1 || strings.Contains(output, "v1")
+				sawV2 = sawV2 || strings.Contains(output, "v2")
+				g.Expect(sawV1).To(BeTrue(), "never saw response from helloworld v1")
+				g.Expect(sawV2).To(BeTrue(), "never saw response from helloworld v2")
+			}).WithTimeout(3 * time.Minute).WithPolling(1 * time.Second).Should(Succeed())
```
mkolesnik: When `Exec` returns an error, `return` skips all assertions. Gomega treats a normal return as success, so Eventually stops retrying instead of polling for the full 3 minutes.

Move the assertions outside the error guard so they always run (and fail until exec succeeds and both versions are seen):
```suggestion
			Eventually(func(g Gomega) {
				output, err := spokeClients["cluster1"].Exec(ctx,
					sampleNS, "deployment/curl", "curl",
					[]string{"curl", "-sS", "--fail", fmt.Sprintf("helloworld.%s:5000/hello", sampleNS)})
				if err == nil {
					sawV1 = sawV1 || strings.Contains(output, "v1")
					sawV2 = sawV2 || strings.Contains(output, "v2")
				}
				g.Expect(sawV1).To(BeTrue(), "never saw response from helloworld v1")
				g.Expect(sawV2).To(BeTrue(), "never saw response from helloworld v2")
			}).WithTimeout(3 * time.Minute).WithPolling(1 * time.Second).Should(Succeed())
```

test/e2e/multi_primary_test.go
```diff
@@ -0,0 +1,270 @@
...
+					secret := &corev1.Secret{}
+					Expect(targetClient.Get(ctx, key.Of("istio-remote-secret-"+source, cpNamespace), secret)).
+						To(Succeed(), "remote secret for %s not found on %s", source, target)
+					Expect(secret.Labels).To(HaveKeyWithValue("istio/multiCluster", "true"))
+					Expect(secret.Annotations).To(HaveKeyWithValue("networking.istio.io/cluster", source))
+				}
+			}
+		}, SpecTimeout(2*time.Minute))
+
+		It("should have cross-cluster data plane traffic working", func(ctx SpecContext) {
```
mkolesnik, line 186: The test only curls from cluster1.
Asymmetric failures (cluster2->cluster1 broken) would be missed.

Sail's `multicluster_multiprimary_test.go` verifies from both clusters.

Either restructure this test to a table inside a `When` context with shared setup, or just loop over clusters for the verification:
```go
Step("Deploying curl on both clusters")
for _, spokeClient := range spokeClients {
	spokeClient.ApplyFile(ctx, filepath.Join(testdataDir, "curl.yaml"), nil, sampleNS)
}
// ... wait for curl on both clusters ...

Step("Verifying cross-cluster traffic from each cluster")
for cluster, spokeClient := range spokeClients {
	var sawV1, sawV2 bool
	Eventually(func(g Gomega) {
		output, err := spokeClient.Exec(ctx,
			sampleNS, "deployment/curl", "curl",
			[]string{"curl", "-sS", "--fail", fmt.Sprintf("helloworld.%s:5000/hello", sampleNS)})
		if err == nil {
			sawV1 = sawV1 || strings.Contains(output, "v1")
			sawV2 = sawV2 || strings.Contains(output, "v2")
		}
		g.Expect(sawV1).To(BeTrue(), "never saw v1 from %s", cluster)
		g.Expect(sawV2).To(BeTrue(), "never saw v2 from %s", cluster)
	}).WithTimeout(3 * time.Minute).WithPolling(1 * time.Second).Should(Succeed())
	Success("Cross-cluster traffic verified from %s", cluster)
}
```

hack/dev-env.sh
```diff
@@ -271,6 +271,80 @@ install_managed_serviceaccount() {
...
+                || err "Failed to inspect Kind network with Podman" ;;
+        docker)
+            kind_subnet="$(docker network inspect kind -f '{{(index .IPAM.Config 0).Subnet}}')" \
+                || err "Failed to inspect Kind network with Docker" ;;
+        *)
+            err "Unsupported container engine: ${engine}" ;;
+    esac
+
+    local base_prefix
+    base_prefix="$(echo "${kind_subnet}" | cut -d'.' -f1-2)"
```
mkolesnik: This extracts the first two octets and builds IPs in the `.255.x` range, assuming a `/16` subnet.
Podman's default Kind network is `10.89.0.0/24`, so `10.89.255.x` is unreachable.

I tested this manually: istiod discovers remote endpoints, but traffic through the east-west gateway times out because the MetalLB IPs aren't routable.

Use a range within the actual subnet, e.g. `10.89.0.200-210` / `10.89.0.211-220`.
Parse the subnet mask and compute a valid range, or use a fixed high-host offset within the subnet:
```bash
base_prefix="$(echo "${kind_subnet}" | cut -d'/' -f1 | cut -d'.' -f1-3)"
range_start="${base_prefix}.$((200 + idx * 10 + 1))"
range_end="${base_prefix}.$((200 + idx * 10 + 10))"
```

Makefile
```diff
@@ -166,6 +166,34 @@ test-e2e: ## Run e2e tests against dev-env clusters (requires make dev-env)
...
+	$(call log,Installing MetalLB: $*)
+	$(DEV_ENV_SCRIPT) install-metallb $*
+
+.PHONY: install-gateway-api
+install-gateway-api: $(addprefix install-gateway-api-,$(SPOKE_CLUSTERS)) ## Install Gateway API CRDs on spoke Kind clusters
+
+.PHONY: $(addprefix install-gateway-api-,$(SPOKE_CLUSTERS))
+$(addprefix install-gateway-api-,$(SPOKE_CLUSTERS)): install-gateway-api-%:
+	$(call log,Installing Gateway API: $*)
+	$(DEV_ENV_SCRIPT) install-gateway-api $*
```
mkolesnik: `SPOKE_CLUSTERS` is defined at L254 in the dev-env section, after these targets.
Make expands `$(addprefix install-metallb-,$(SPOKE_CLUSTERS))` at parse time, so it resolves to nothing.
`make test-e2e-multicluster-prereqs` reports "Nothing to be done" without installing anything.

I tested this on the dev-env: Gateway API CRDs and MetalLB were never installed, causing istiod to hang.

`test-e2e-multicluster` and prereqs can stay here (next to `test-e2e`), since its recipes use `$(DEV_KUBE_DIR)` which expands at execution time.
But `install-metallb` and `install-gateway-api` use `$(SPOKE_CLUSTERS)` in prerequisites, which expand at parse time. 

Move those these targets to the dev-env section after L254.

test/e2e/multi_primary_test.go
```diff
@@ -0,0 +1,274 @@
...
+			g.Expect(hubClient.Get(ctx, key.For(mesh), mesh)).To(Succeed())
+			g.Expect(meta.IsStatusConditionTrue(mesh.Status.Conditions, meshv1alpha1.ConditionReady)).To(BeTrue())
+		}).Should(Succeed())
+
+		Step("Verifying control plane namespace exists with network labels")
+		for _, spokeClient := range spokeClients {
+			ns := &corev1.Namespace{}
+			Expect(spokeClient.Get(ctx, key.Of(cpNamespace), ns)).To(Succeed())
+			Expect(ns.Labels).To(HaveKey("topology.istio.io/network"))
+		}
```
mkolesnik, line 100: The test can't verify cross-cluster traffic without remote secrets, and the controller doesn't distribute them yet.

Consider adding a temporary helper to create them in the test setup:

```go
// TODO(endpoint-discovery): Remove once the controller distributes remote secrets.
Step("Creating remote secrets for cross-cluster discovery")
createRemoteSecrets(ctx, spokeClients, cpNamespace)
```

The helper builds a kubeconfig from each spoke's kubeconfig (replacing the host-mapped port with `6443` for the internal Kind API) and applies it as an Istio remote secret on the other cluster.

I tested this locally and it works.
Happy to share the implementation if helpful.

sridhargaddam: I'm not fully convinced that we should add a temporary `createRemoteSecrets` method, but I'm ok to go ahead with it for now. We can remove it in a future PR once the appropriate functionality is added to the controller.

Makefile
sridhargaddam: Hmm.. I think this happened while addressing some of the review comments.

sridhargaddam:
The tests build and run successfully...

```
------------------------------
Multi-primary data plane when Istio multi-cluster is deployed should have cross-cluster data plane traffic working [Serial]
/home/sgaddam/goprojects/downstream/multicluster-mesh-addon/test/e2e/multi_primary_test.go:198
  STEP: Creating sample namespace with istio-injection on both clusters @ 08/03/26 21:30:42.721
  STEP: Deploying helloworld Service on both clusters @ 08/03/26 21:30:42.759
  STEP: Deploying helloworld-v1 on cluster1 @ 08/03/26 21:30:42.984
  STEP: Deploying helloworld-v2 on cluster2 @ 08/03/26 21:30:43.047
  STEP: Deploying curl on both clusters @ 08/03/26 21:30:43.119
  STEP: Waiting for helloworld-v1 to be ready on cluster1 @ 08/03/26 21:30:43.269
  STEP: Waiting for helloworld-v2 to be ready on cluster2 @ 08/03/26 21:31:05.316
  STEP: Waiting for curl to be ready on both clusters @ 08/03/26 21:31:05.319
  STEP: Verifying cross-cluster traffic from each cluster @ 08/03/26 21:31:05.328
  * Cross-cluster traffic verified from cluster1
  * Cross-cluster traffic verified from cluster2
  STEP: Cleaning up remote secrets @ 08/03/26 21:31:18.609
  STEP: Cleaning up spoke resources @ 08/03/26 21:31:18.667
  STEP: Deleting test mesh multi-primary-mesh @ 08/03/26 21:31:19.908
  STEP: Cleaning up hub resources @ 08/03/26 21:31:19.926
• [37.425 seconds]
------------------------------

Ran 2 of 2 Specs in 161.060 seconds
SUCCESS! -- 2 Passed | 0 Failed | 0 Pending | 0 Skipped
PASS

Ginkgo ran 1 suite in 2m43.020190752s
Test Suite Passed
```

test/e2e/multi_primary_test.go
```diff
@@ -0,0 +1,290 @@
...
+	)
+
+	BeforeAll(func(ctx SpecContext) {
+		meshNS = util.UniqueName("mp-test-ns")
+		util.CreateNamespace(ctx, hubClient, meshNS)
+
+		Step("Setting up cert-manager trust chain in %s", meshNS)
+		hubClient.ApplyFile(ctx, filepath.Join(samplesDir, "cert-manager-issuer.yaml"), nil, meshNS)
+
```
yxun, line 62: We need a step of enabling ManifestWorkReplicaSet feature. I can work on that in a separate PR after this e2e one. It needs to update a `ClusterManager` CR.

test/e2e/multi_primary_test.go
```diff
@@ -0,0 +1,290 @@
...
+		// REVISIT: Currently, the controller does not support secret distribution.
+		// See PIt specs in mesh_lifecycle_test.go for what the controller will eventually do.
+		It("should distribute remote secrets to spoke clusters", func(ctx SpecContext) {
+			for source := range spokeClients {
+				for target, targetClient := range spokeClients {
+					if source == target {
+						continue
+					}
+					secret := &corev1.Secret{}
+					Expect(targetClient.Get(ctx, key.Of("istio-remote-secret-"+source, cpNamespace), secret)).
```
yxun, line 178: Because multiple meshes can target the same cluster, I am using this remote secret naming format:

`fmt.Sprintf("%s-%s-%s-%s", mesh.Namespace, mesh.Name, "istio-remote-secret", cluster.Name)`

I can update this name in separate PR later after the controller change PR.

test/e2e/multi_primary_test.go
sridhargaddam: Sure, as part of your controller changes, we can update this part of the code as well. Whether we should automatically update the feature gate in the code/makefile or make it a prerequisite is something we can discuss during your PR review.

test/e2e/multi_primary_test.go
sridhargaddam: `remote_secret.go` is a temporary implementation which will be deleted once the controller changes are implemented. So, feel free to delete it and modify this part of code as part of your changes.

test/e2e/multi_primary_test.go
```diff
@@ -0,0 +1,290 @@
...
+	}, NodeTimeout(5*time.Minute))
+
+	AfterAll(func(ctx SpecContext) {
+		// TODO(endpoint-discovery): Remove once the controller distributes remote secrets.
+		Step("Cleaning up remote secrets")
+		spokeClientMap := make(map[string]client.Client, len(spokeClients))
+		for name, sc := range spokeClients {
+			spokeClientMap[name] = sc
+		}
+		util.CleanupRemoteSecrets(ctx, spokeClientMap, clusters, cpNamespace)
```
mkolesnik: nit: These remote secrets live in `cpNamespace` (istio-system).
When the mesh is deleted below, the controller deletes the namespace MW, which cascades to delete everything in the namespace on spoke.

This cleanup block (and `CleanupRemoteSecrets`) is redundant.

```suggestion
```

I'm not holding on this, just wanted to bring it to your attention

test/e2e/multi_primary_test.go
```diff
@@ -0,0 +1,290 @@
...
+				waitForDeploymentReady(ctx, spokeClient, "istiod", cpNamespace, 5*time.Minute)
+				Success("istiod is ready on %s", cluster)
+			}
+
+			// TODO(endpoint-discovery): Remove once the controller distributes remote secrets.
+			Step("Creating remote secrets for cross-cluster discovery")
+			spokeClientMap := make(map[string]client.Client, len(spokeClients))
+			for name, sc := range spokeClients {
+				spokeClientMap[name] = sc
+			}
```
mkolesnik: nit: FYI The map conversion is forced by `CreateAndDistributeRemoteSecrets` accepting `map[string]client.Client`.
Since this is test code, the function could just accept `map[string]*util.E2EClient` directly (same package) or use the global `spokeClients`.

I'm not holding on this, just wanted to bring it to your attention

mkolesnik (commented):
Thanks Sridhar,

I have some nits on the temp code but won't block on them, so FYI and if you want to fix them fix them.
I'm approving this and it's still on hold, so remove the hold after the other reviews are done.

/lgtm

test/e2e/multi_primary_test.go
sridhargaddam: Agree @mkolesnik that the namespace will be deleted. This whole section was a temporary code. Let me update the PR.

mkolesnik (approved):

PR #220, opened 2026-07-15
Create control plane namespace on spoke clusters

### LINKED ISSUE #219: Create control plane namespace early when mesh is created
When a `MultiClusterMesh` is created, the controller distributes `cacerts` to spoke clusters via ManifestWork.
The ManifestWork targets the control plane namespace (default `istio-system`), which does not exist yet at that point.
The work agent fails with "namespaces <namespace> not found" and enters exponential backoff.
It recovers once the namespace is created (when Istio is installed), but the default backoff caps at ~16 minutes, so the delay can be significant.

The controller should create the control plane namespace on spoke clusters as soon as the mesh is created, before distributing cacerts or any other resources that target it.

The namespace also needs the `topology.istio.io/network` label for multi-network setups.
Istio uses this label as a fallback to determine what network non-injected endpoints belong to.
The sidecar injector also sets this label on injected pods (from `values.global.network`), so for injected workloads the namespace label is redundant, but it is still part of Istio documented prerequisites and used in ambient mode.

### Approaches

**Option A: `ManifestWorkReplicaSet` with `CreateOnly`**

Create a bare namespace (no network label) via MWRS.

Pros:
- Simple, single MWRS fans out to all clusters
- Scales automatically with Placement, no per-cluster management
- `CreateOnly` preserves any labels added later by the user

Cons:
- Namespace is created without the network label, user must set it manually on each cluster
- Requires `ManifestWorkReplicaSet` feature gate on ClusterManager

**Option B: Per-cluster `ManifestWork` with network label**

Create the namespace via per-cluster ManifestWork, setting `topology.istio.io/network` to a derived value (e.g. the cluster name, or `<clustername>-network`).

Pros:
- Namespace is fully configured in one step
- No feature gate dependency
- Consistent with how we already create per-cluster ManifestWorks (cacerts, operator)

Cons:
- One ManifestWork per cluster instead of one MWRS, manual scaling and tracking
- Controller must derive the network name, which couples us to a naming convention
- The network name must match what is later set in the Istio CR `values.global.network`, adding a coordination requirement

**Option C: `Policy`**

Use a configuration policy to ensure the namespace exists on all clusters matching a Placement.

Pros:
- Scales automatically with Placement like MWRS
- Can enforce (re-create if deleted) or just inform
- Could template the network label per cluster using hub templates

Cons:
- Dependency on the governance-policy-framework addon being installed
- More complex resource model (Policy + PlacementBinding + Placement)
- Heavier mechanism for creating a single namespace

mkolesnik:
The cacerts ManifestWork targets the control plane namespace (default istio-system), which does not exist until the Istio operator creates it.
This causes the work agent to fail and enter exponential backoff.

Create the namespace via a per-cluster ManifestWork before applying operator or cacerts resources.
Make `spec.controlPlane.namespace` immutable and validate it does not collide with the operator namespace.

Closes #219

### DIFF
--- a/pkg/hub/mesh/controller.go
+++ b/pkg/hub/mesh/controller.go
@@ -42,8 +42,9 @@ import (
 )
 
 const (
-	OperatorManifestWorkName = "multicluster-mesh-operator"
-	ManifestWorkNameCacerts  = "multicluster-mesh-cacerts"
+	OperatorManifestWorkName   = "multicluster-mesh-operator"
+	ManifestWorkNameCacerts    = "multicluster-mesh-cacerts"
+	ManifestWorkNameCPNSPrefix = "multicluster-mesh-cp-ns-"
 
 	FeedbackInstalledCSV = "installedCSV"
 
@@ -57,7 +58,8 @@ const (
 	MeshNameLabel      = "mesh.open-cluster-management.io/mesh-name"
 	MeshNamespaceLabel = "mesh.open-cluster-management.io/mesh-namespace"
 
-	ClusterSetLabel = "cluster.open-cluster-management.io/clusterset"
+	ClusterSetLabel   = "cluster.open-cluster-management.io/clusterset"
+	IstioNetworkLabel = "topology.istio.io/network"
 
 	Day = 24 * time.Hour
 )
@@ -211,6 +213,14 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 // validate checks for conflicts that prevent reconciliation.
 // Sets a condition on the mesh and returns true if a conflict is found.
 func (r *Reconciler) validate(ctx context.Context, mesh *meshv1alpha1.MultiClusterMesh) (conflict bool, err error) {
+	// CEL cross-field rule on the spec struct exceeds the estimated cost budget,
+	// so this is validated here instead of via kubebuilder markers.
+	if mesh.GetControlPlaneNamespace() == mesh.Spec.Operator.Namespace {
+		mesh.SetReadyCondition(metav1.ConditionFalse, meshv1alpha1.ReasonNamespaceConflict,
+			"controlPlane.namespace %q must not equal operator.namespace", mesh.GetControlPlaneNamespace())
+		return true, nil
+	}
+
 	if err = r.forEachMeshInClusterSet(ctx, mesh.Spec.ClusterSet, func(other *meshv1alpha1.MultiClusterMesh) {
 		if other.UID == mesh.UID || conflict {
 			return
@@ -249,6 +259,12 @@ func (r *Reconciler) doReconcile(ctx context.Context, mesh *meshv1alpha1.MultiCl
 	for _, cluster := range clusters {
 		klog.V(4).Infof("Reconciling cluster %s", cluster.Name)
 
+		cpNsWork, err := r.workApplier.Apply(ctx, r.buildControlPlaneNamespaceManifestWork(mesh, &cluster))
+		if err != nil {
+			return fmt.Errorf("failed to apply control plane namespace ManifestWork on cluster %s: %w", cluster.Name, err)
+		}
+		klog.V(4).Infof("Applied control plane namespace ManifestWork %s/%s", cpNsWork.Namespace, cpNsWork.Name)
+
 		work, err := r.workApplier.Apply(ctx, r.buildOperatorManifestWork(mesh, &cluster))
 		if err != nil {
 			return fmt.Errorf("failed to apply operator ManifestWork on cluster %s: %w", cluster.Name, err)
@@ -272,6 +288,10 @@ func (r *Reconciler) doReconcile(ctx context.Context, mesh *meshv1alpha1.MultiCl
 	forceCleanupAll := mesh.Spec.Security.Trust.CertManager.IssuerRef.Name == ""
 	if err := r.cleanupCertificates(ctx, mesh, clusters, forceCleanupAll); err != nil {
 		return fmt.Errorf("failed to cleanup Certificates: %w", err)
+	}
+
+	if err := r.cleanupMeshOwnedManifestWorks(ctx, mesh, clusters); err != nil {
+		return fmt.Errorf("failed to cleanup mesh-owned ManifestWorks: %w", err)
 	}
 
 	if err := r.cleanupManifestWorks(ctx, mesh.Spec.ClusterSet); err != nil {
@@ -353,6 +373,11 @@ func (r *Reconciler) handleDeletion(ctx context.Context, mesh *meshv1alpha1.Mult
 	}
 
 	klog.Infof("Handling deletion for MultiClusterMesh %s/%s", mesh.Namespace, mesh.Name)
+
+	if err := r.cleanupMeshOwnedManifestWorks(ctx, mesh, nil); err != nil {
+		return fmt.Errorf("failed to cleanup mesh-owned ManifestWorks: %w", err)
+	}
+
 	if err := r.cleanupManifestWorks(ctx, mesh.Spec.ClusterSet); err != nil {
 		return fmt.Errorf("failed to cleanup ManifestWorks: %w", err)
 	}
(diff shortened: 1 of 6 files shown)


pkg/hub/mesh/controller.go
```diff
@@ -734,6 +783,23 @@ func (r *Reconciler) ensureCacertsManifestWork(ctx context.Context, mesh *meshv1
...
+
+	return buildMeshOwnedManifestWork(mesh, cluster.Name, ManifestWorkNameCPNSPrefix+cpNamespace, &corev1.Namespace{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "v1",
+			Kind:       "Namespace",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name: cpNamespace,
+			Labels: map[string]string{
+				IstioNetworkLabel: cluster.Name,
```
yxun: Our design doc says

```
Network Partitioning: The Controller analyzes the ManagedCluster labels (e.g., 
topology.kubernetes.io/network) and automatically populates the values.global.network field in the 
generated Istio CR. 
```

We may get this IstioNetworkLabel value from ManagedCluster labels (e.g., 
topology.kubernetes.io/network)

pkg/hub/mesh/controller.go
```diff
@@ -749,16 +815,20 @@ func (r *Reconciler) buildCacertsManifestWork(mesh *meshv1alpha1.MultiClusterMes
...
 			Namespace: clusterName,
 			Labels:    meshOwnedLabels(mesh, clusterName),
 		},
 		Spec: workv1.ManifestWorkSpec{
 			Workload: workv1.ManifestsTemplate{
 				Manifests: []workv1.Manifest{{
-					RawExtension: runtime.RawExtension{Object: cacertsSecret},
+					RawExtension: runtime.RawExtension{Object: obj},
 				}},
 			},
```
yxun, line 838:
```suggestion
			},
			ManifestConfigs: []workv1.ManifestConfigOption{
				workv1.ManifestConfigOption{
					UpdateStrategy: &workv1.UpdateStrategy{
						Type: workv1.UpdateStrategyTypeCreateOnly},
				},
			},
```

We may add `CreateOnly` update strategy type here.

mkolesnik: What's the rationale?

I checked and: 
* Sail watches namespaces but never modifies them.
* Istio only reads the labels from the namespace.

I'd prefer to keep the default `Update` strategy here to keep it simple for now.

pkg/hub/mesh/controller.go
mkolesnik: I'll update this PR to read `topology.kubernetes.io/network` from ManagedCluster labels with fallback to `cluster.Name`.

sridhargaddam: Please also update the design doc.

sridhargaddam: What are your thoughts about user configuring Istio CR with a different network label via values.global.network? Should we document that users should configure the global.network with the clusterName (or the network specified as part of topology.kubernetes.io/network field) itself in our design.doc (or at an appropriate place in docs)?

pkg/hub/mesh/controller.go
```diff
@@ -249,6 +259,12 @@ func (r *Reconciler) doReconcile(ctx context.Context, mesh *meshv1alpha1.MultiCl
 	for _, cluster := range clusters {
 		klog.V(4).Infof("Reconciling cluster %s", cluster.Name)
 
+		cpNsWork, err := r.workApplier.Apply(ctx, r.buildControlPlaneNamespaceManifestWork(mesh, &cluster))
```
sridhargaddam, line 262: Please update the design doc that Addon will now create the ControlPlaneNamespace.

pkg/hub/mesh/controller.go
```diff
@@ -429,6 +454,30 @@ func (r *Reconciler) cleanupCertificates(ctx context.Context, mesh *meshv1alpha1
...
+		return fmt.Errorf("failed to list mesh-owned ManifestWorks: %w", err)
+	}
+
+	for _, work := range workList.Items {
+		if clusterNames[work.Namespace] {
+			continue
+		}
+
+		klog.Infof("Deleting mesh-owned ManifestWork %s/%s", work.Namespace, work.Name)
+		if err := r.workApplier.Delete(ctx, work.Namespace, work.Name); err != nil {
```
sridhargaddam, line 473: Currently, in phase1 we expect users to install Istio control plane using GitOps/ArgoCD etc. If the addon deletes the controlPlane namespace we will be deleting any artifacts present in the namespace. This should be explicitly documented in our docs.

pkg/hub/mesh/controller.go
```diff
@@ -429,6 +454,30 @@ func (r *Reconciler) cleanupCertificates(ctx context.Context, mesh *meshv1alpha1
...
+			return fmt.Errorf("failed to delete ManifestWork %s/%s: %w", work.Namespace, work.Name, err)
```
sridhargaddam, line 474: What would happen if there is a pre-existing controlPlane namespace on the ManagedClusters? I guess the addon will now own it and deletes the namespace when the mesh is deleted. This violates the do no harm to the existing things on the cluster.

test/integration/controller_test.go
```diff
@@ -427,6 +454,22 @@ var _ = Describe("MultiClusterMesh Controller", func() {
...
+			expectMeshConditionReason(conflictMesh, testNs, meshv1alpha1.ConditionReady, meshv1alpha1.ReasonNamespaceConflict)
+		})
+
+		It("should block mesh when operator namespace equals default control plane namespace", func() {
+			conflictMesh := meshName + "-opns"
+			util.CreateMultiClusterMesh(ctx, k8sClient, conflictMesh, testNs, testClusterSet, meshv1alpha1.MultiClusterMeshSpec{
+				Operator: meshv1alpha1.OperatorConfig{Namespace: "istio-system"},
+			})
+			expectMeshConditionReason(conflictMesh, testNs, meshv1alpha1.ConditionReady, meshv1alpha1.ReasonNamespaceConflict)
+		})
```
sridhargaddam, line 502: Do we really need this test-case as its kind of validated already in the previous test?

pkg/hub/mesh/controller.go
yxun: If there is a running Istio control plane or other resources not owned by the ManifestWork, we don't want to accidentally delete them when their namespace is deleted on the spoke with the removed of ManifestWork.

Ok, I agree we can keep it simple with the default strategy for now. And revisit resources deletion or garbage collection after integrating Istio installation using this controller.

Ref: https://open-cluster-management.io/docs/concepts/work-distribution/manifestwork/#garbage-collection
```
By default, deleteOption is set as Foreground which means the applied resources on the spoke will be deleted with the removal of ManifestWork. 
```

pkg/hub/mesh/controller.go
mkolesnik: As I said to @yxun let's keep it simple for now.

If we want to support brownfield in the future this is one of many places we'll have to fix, and I suspect an OCM policy would be a better fit for this.

pkg/hub/mesh/controller.go
mkolesnik: Done in the latest push. Design doc updated, and added `values.global.network` to the Phase 1 user responsibilities.

pkg/hub/mesh/controller.go
mkolesnik: Added a "Control Plane Namespace" section to the design doc.

pkg/hub/mesh/controller.go
mkolesnik: Documented. The new section notes that deleting the mesh deletes the namespace and everything in it.

test/integration/controller_test.go
mkolesnik: They test opposite directions of the same conflict. One sets CP namespace to the operator namespace, the other sets operator namespace to the default CP namespace. Different code paths.

yxun:
lgtm

PR #221, opened 2026-07-16
Add ManagedClusterSetBinding and Placement in controller

yxun:
This PR adds `Placement` API from `open-cluster-management.io/api/cluster/v1beta1` and `ManagedClusterSetBinding` API from `open-cluster-management.io/api/cluster/v1beta2` . 

This PR adds controller permission for those two resources and integration tests. 

TODO: Implement resource distribution via `ManifestWorkReplicaSet`.  Add integration tests and e2e tests. PR #215

### DIFF
--- a/main.go
+++ b/main.go
@@ -20,8 +20,10 @@ import (
 	"k8s.io/klog/v2"
 	"k8s.io/utils/clock"
 	clusterv1 "open-cluster-management.io/api/cluster/v1"
+	clusterv1beta1 "open-cluster-management.io/api/cluster/v1beta1"
 	clusterv1beta2 "open-cluster-management.io/api/cluster/v1beta2"
 	workv1 "open-cluster-management.io/api/work/v1"
+	workv1alpha1 "open-cluster-management.io/api/work/v1alpha1"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
@@ -42,8 +44,10 @@ func init() {
 	utilruntime.Must(scheme.AddToScheme(runtimeScheme))
 	utilruntime.Must(meshv1alpha1.Install(runtimeScheme))
 	utilruntime.Must(clusterv1.Install(runtimeScheme))
+	utilruntime.Must(clusterv1beta1.Install(runtimeScheme))
 	utilruntime.Must(clusterv1beta2.Install(runtimeScheme))
 	utilruntime.Must(workv1.Install(runtimeScheme))
+	utilruntime.Must(workv1alpha1.Install(runtimeScheme))
 	utilruntime.Must(operatorsv1.AddToScheme(runtimeScheme))
 	utilruntime.Must(operatorsv1alpha1.AddToScheme(runtimeScheme))
 	utilruntime.Must(certmanagerv1.AddToScheme(runtimeScheme))
--- a/main.go
+++ b/main.go
@@ -22,7 +22,6 @@ import (
 	clusterv1 "open-cluster-management.io/api/cluster/v1"
 	clusterv1beta1 "open-cluster-management.io/api/cluster/v1beta1"
 	clusterv1beta2 "open-cluster-management.io/api/cluster/v1beta2"
-	ocmoperatorv1 "open-cluster-management.io/api/operator/v1"
 	workv1 "open-cluster-management.io/api/work/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
@@ -47,7 +46,6 @@ func init() {
 	utilruntime.Must(clusterv1beta1.Install(runtimeScheme))
 	utilruntime.Must(clusterv1beta2.Install(runtimeScheme))
 	utilruntime.Must(workv1.Install(runtimeScheme))
-	utilruntime.Must(ocmoperatorv1.Install(runtimeScheme))
 	utilruntime.Must(operatorsv1.AddToScheme(runtimeScheme))
 	utilruntime.Must(operatorsv1alpha1.AddToScheme(runtimeScheme))
 	utilruntime.Must(certmanagerv1.AddToScheme(runtimeScheme))
@@ -169,12 +167,6 @@ func runController(ctx context.Context, controllerContext *controllercmd.Control
 	klog.Info("Starting manager...")
 	if err := mgr.Start(ctx); err != nil {
 		klog.Errorf("Problem running manager: %v", err)
-		return err
-	}
-
-	// Enable ManifestWorkReplicaSet feature
-	if err := meshcontroller.EnableFeatureManifestWorkReplicaSet(context.Background(), mgr.GetClient()); err != nil {
-		klog.Errorf("failed to enable ManifestWorkReplicaSet feature for hub ClusterManager: %v", err)
 		return err
 	}
 
--- a/main.go
+++ b/main.go
@@ -23,7 +23,6 @@ import (
 	clusterv1beta1 "open-cluster-management.io/api/cluster/v1beta1"
 	clusterv1beta2 "open-cluster-management.io/api/cluster/v1beta2"
 	workv1 "open-cluster-management.io/api/work/v1"
-	workv1alpha1 "open-cluster-management.io/api/work/v1alpha1"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
@@ -47,7 +46,6 @@ func init() {
 	utilruntime.Must(clusterv1beta1.Install(runtimeScheme))
 	utilruntime.Must(clusterv1beta2.Install(runtimeScheme))
 	utilruntime.Must(workv1.Install(runtimeScheme))
-	utilruntime.Must(workv1alpha1.Install(runtimeScheme))
 	utilruntime.Must(operatorsv1.AddToScheme(runtimeScheme))
 	utilruntime.Must(operatorsv1alpha1.AddToScheme(runtimeScheme))
 	utilruntime.Must(certmanagerv1.AddToScheme(runtimeScheme))
--- a/main.go
+++ b/main.go
@@ -22,6 +22,7 @@ import (
 	clusterv1 "open-cluster-management.io/api/cluster/v1"
 	clusterv1beta1 "open-cluster-management.io/api/cluster/v1beta1"
 	clusterv1beta2 "open-cluster-management.io/api/cluster/v1beta2"
+	ocmoperatorv1 "open-cluster-management.io/api/operator/v1"
 	workv1 "open-cluster-management.io/api/work/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
@@ -46,6 +47,7 @@ func init() {
 	utilruntime.Must(clusterv1beta1.Install(runtimeScheme))
 	utilruntime.Must(clusterv1beta2.Install(runtimeScheme))
 	utilruntime.Must(workv1.Install(runtimeScheme))
+	utilruntime.Must(ocmoperatorv1.Install(runtimeScheme))
 	utilruntime.Must(operatorsv1.AddToScheme(runtimeScheme))
 	utilruntime.Must(operatorsv1alpha1.AddToScheme(runtimeScheme))
 	utilruntime.Must(certmanagerv1.AddToScheme(runtimeScheme))
@@ -167,6 +169,12 @@ func runController(ctx context.Context, controllerContext *controllercmd.Control
 	klog.Info("Starting manager...")
 	if err := mgr.Start(ctx); err != nil {
 		klog.Errorf("Problem running manager: %v", err)
+		return err
+	}
+
+	// Enable ManifestWorkReplicaSet feature
+	if err := meshcontroller.EnableFeatureManifestWorkReplicaSet(context.Background(), mgr.GetClient()); err != nil {
+		klog.Errorf("failed to enable ManifestWorkReplicaSet feature for hub ClusterManager: %v", err)
 		return err
 	}
 
--- a/pkg/hub/mesh/controller.go
+++ b/pkg/hub/mesh/controller.go
@@ -561,9 +561,9 @@ func (r *Reconciler) getClustersFromSet(ctx context.Context, clusterSetName stri
 	return clusterList.Items, nil
 }
 
-func (r *Reconciler) getManifestWorkReplicaSet(ctx context.Context, mwrsName string) (*workv1alpha1.ManifestWorkReplicaSet, error) {
+func (r *Reconciler) getManifestWorkReplicaSet(ctx context.Context, mwrsNamespace, mwrsName string) (*workv1alpha1.ManifestWorkReplicaSet, error) {
 	mwrs := &workv1alpha1.ManifestWorkReplicaSet{}
-	if err := r.Get(ctx, key.Of(mwrsName), mwrs); err != nil {
+	if err := r.Get(ctx, key.Of(mwrsName, mwrsNamespace), mwrs); err != nil {
 		if apierrors.IsNotFound(err) {
 			klog.V(4).Infof("ManifestWorkReplicaSet %s not found", mwrsName)
 			return nil, nil
@@ -732,6 +732,7 @@ func (r *Reconciler) ensureCacertsManifestWork(ctx context.Context, mesh *meshv1
 	secretName := getCacertsName(cluster.Name)
 	secret := &corev1.Secret{}
 	err := r.Get(ctx, key.Of(secretName, mesh.Namespace), secret)
+
 	if err != nil {
 		if apierrors.IsNotFound(err) {
 			klog.V(4).Infof("Secret %s/%s not found yet, waiting for cert-manager to create it", mesh.Namespace, secretName)
--- a/pkg/hub/mesh/controller.go
+++ b/pkg/hub/mesh/controller.go
@@ -46,6 +46,7 @@ const (
 	OperatorManifestWorkName   = "multicluster-mesh-operator"
 	ManifestWorkNameCacerts    = "multicluster-mesh-cacerts"
 	ManifestWorkReplicaSetName = "multicluster-mesh-mwrset"
+	ManifestWorkNameCPNSPrefix = "multicluster-mesh-cp-ns-"
 
 	FeedbackInstalledCSV = "installedCSV"
 
@@ -59,7 +60,8 @@ const (
 	MeshNameLabel      = "mesh.open-cluster-management.io/mesh-name"
 	MeshNamespaceLabel = "mesh.open-cluster-management.io/mesh-namespace"
 
-	ClusterSetLabel = "cluster.open-cluster-management.io/clusterset"
+	ClusterSetLabel   = "cluster.open-cluster-management.io/clusterset"
+	IstioNetworkLabel = "topology.istio.io/network"
 
 	Day = 24 * time.Hour
 )
@@ -215,6 +217,14 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 // validate checks for conflicts that prevent reconciliation.
 // Sets a condition on the mesh and returns true if a conflict is found.
 func (r *Reconciler) validate(ctx context.Context, mesh *meshv1alpha1.MultiClusterMesh) (conflict bool, err error) {
+	// CEL cross-field rule on the spec struct exceeds the estimated cost budget,
+	// so this is validated here instead of via kubebuilder markers.
+	if mesh.GetControlPlaneNamespace() == mesh.Spec.Operator.Namespace {
+		mesh.SetReadyCondition(metav1.ConditionFalse, meshv1alpha1.ReasonNamespaceConflict,
+			"controlPlane.namespace %q must not equal operator.namespace", mesh.GetControlPlaneNamespace())
+		return true, nil
+	}
+
 	if err = r.forEachMeshInClusterSet(ctx, mesh.Spec.ClusterSet, func(other *meshv1alpha1.MultiClusterMesh) {
 		if other.UID == mesh.UID || conflict {
 			return
@@ -253,6 +263,12 @@ func (r *Reconciler) doReconcile(ctx context.Context, mesh *meshv1alpha1.MultiCl
 	for _, cluster := range clusters {
 		klog.V(4).Infof("Reconciling cluster %s", cluster.Name)
 
+		cpNsWork, err := r.workApplier.Apply(ctx, r.buildControlPlaneNamespaceManifestWork(mesh, &cluster))
+		if err != nil {
+			return fmt.Errorf("failed to apply control plane namespace ManifestWork on cluster %s: %w", cluster.Name, err)
+		}
+		klog.V(4).Infof("Applied control plane namespace ManifestWork %s/%s", cpNsWork.Namespace, cpNsWork.Name)
+
 		work, err := r.workApplier.Apply(ctx, r.buildOperatorManifestWork(mesh, &cluster))
 		if err != nil {
 			return fmt.Errorf("failed to apply operator ManifestWork on cluster %s: %w", cluster.Name, err)
@@ -276,6 +292,10 @@ func (r *Reconciler) doReconcile(ctx context.Context, mesh *meshv1alpha1.MultiCl
 	forceCleanupAll := mesh.Spec.Security.Trust.CertManager.IssuerRef.Name == ""
 	if err := r.cleanupCertificates(ctx, mesh, clusters, forceCleanupAll); err != nil {
 		return fmt.Errorf("failed to cleanup Certificates: %w", err)
+	}
+
+	if err := r.cleanupMeshOwnedManifestWorks(ctx, mesh, clusters); err != nil {
+		return fmt.Errorf("failed to cleanup mesh-owned ManifestWorks: %w", err)
 	}
 
 	if err := r.cleanupManifestWorks(ctx, mesh.Spec.ClusterSet); err != nil {
@@ -357,6 +377,11 @@ func (r *Reconciler) handleDeletion(ctx context.Context, mesh *meshv1alpha1.Mult
 	}
 
 	klog.Infof("Handling deletion for MultiClusterMesh %s/%s", mesh.Namespace, mesh.Name)
+
+	if err := r.cleanupMeshOwnedManifestWorks(ctx, mesh, nil); err != nil {
+		return fmt.Errorf("failed to cleanup mesh-owned ManifestWorks: %w", err)
+	}
+
 	if err := r.cleanupManifestWorks(ctx, mesh.Spec.ClusterSet); err != nil {
 		return fmt.Errorf("failed to cleanup ManifestWorks: %w", err)
 	}
@@ -427,6 +452,30 @@ func (r *Reconciler) cleanupCertificates(ctx context.Context, mesh *meshv1alpha1
 		klog.Infof("Deleting Certificate %s/%s (cluster %s no longer in ClusterSet %s)", cert.Namespace, cert.Name, clusterName, mesh.Spec.ClusterSet)
 		if err := client.IgnoreNotFound(r.Delete(ctx, &cert)); err != nil {
 			return fmt.Errorf("failed to delete Certificate %s/%s: %w", cert.Namespace, cert.Name, err)
+		}
+	}
+
+	return nil
+}
+
+func (r *Reconciler) cleanupMeshOwnedManifestWorks(ctx context.Context, mesh *meshv1alpha1.MultiClusterMesh, clusters []clusterv1.ManagedCluster) error {
+	clusterNames := clusterNameSet(clusters)
+
+	workList := &workv1.ManifestWorkList{}
+	if err := r.List(ctx, workList,
+		client.MatchingLabels{MeshNameLabel: mesh.Name, MeshNamespaceLabel: mesh.Namespace},
+	); err != nil {
+		return fmt.Errorf("failed to list mesh-owned ManifestWorks: %w", err)
+	}
+
+	for _, work := range workList.Items {
+		if clusterNames[work.Namespace] {
+			continue
+		}
+
+		klog.Infof("Deleting mesh-owned ManifestWork %s/%s", work.Namespace, work.Name)
+		if err := r.workApplier.Delete(ctx, work.Namespace, work.Name); err != nil {
+			return fmt.Errorf("failed to delete ManifestWork %s/%s: %w", work.Namespace, work.Name, err)
 		}
 	}
 
@@ -750,6 +799,28 @@ func (r *Reconciler) ensureCacertsManifestWork(ctx context.Context, mesh *meshv1
 	return nil
 }
 
+func (r *Reconciler) buildControlPlaneNamespaceManifestWork(mesh *meshv1alpha1.MultiClusterMesh, cluster *clusterv1.ManagedCluster) *workv1.ManifestWork {
+	cpNamespace := mesh.GetControlPlaneNamespace()
+
+	network := cluster.Name
+	if v, ok := cluster.Labels[IstioNetworkLabel]; ok && v != "" {
+		network = v
+	}
+
+	return buildMeshOwnedManifestWork(mesh, cluster.Name, ManifestWorkNameCPNSPrefix+cpNamespace, &corev1.Namespace{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "v1",
+			Kind:       "Namespace",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name: cpNamespace,
+			Labels: map[string]string{
+				IstioNetworkLabel: network,
+			},
+		},
+	})
+}
+
 // buildCacertsManifestWork builds a ManifestWork for distributing the cacerts secret
 func (r *Reconciler) buildCacertsManifestWork(mesh *meshv1alpha1.MultiClusterMesh, clusterName string, secret *corev1.Secret) *workv1.ManifestWork {
 	cacertsSecret := &corev1.Secret{
@@ -765,16 +836,20 @@ func (r *Reconciler) buildCacertsManifestWork(mesh *meshv1alpha1.MultiClusterMes
 		Data: secret.Data,
 	}
 
+	return buildMeshOwnedManifestWork(mesh, clusterName, ManifestWorkNameCacerts, cacertsSecret)
+}
+
+func buildMeshOwnedManifestWork(mesh *meshv1alpha1.MultiClusterMesh, clusterName, name string, obj runtime.Object) *workv1.ManifestWork {
 	return &workv1.ManifestWork{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      ManifestWorkNameCacerts,
+			Name:      name,
 			Namespace: clusterName,
 			Labels:    meshOwnedLabels(mesh, clusterName),
 		},
 		Spec: workv1.ManifestWorkSpec{
 			Workload: workv1.ManifestsTemplate{
 				Manifests: []workv1.Manifest{{
-					RawExtension: runtime.RawExtension{Object: cacertsSecret},
+					RawExtension: runtime.RawExtension{Object: obj},
 				}},
 			},
 		},
--- a/pkg/hub/mesh/controller.go
+++ b/pkg/hub/mesh/controller.go
@@ -27,6 +27,7 @@ import (
 	clusterv1 "open-cluster-management.io/api/cluster/v1"
 	clusterv1beta2 "open-cluster-management.io/api/cluster/v1beta2"
 	workv1 "open-cluster-management.io/api/work/v1"
+	workv1alpha1 "open-cluster-management.io/api/work/v1alpha1"
 	"open-cluster-management.io/sdk-go/pkg/apis/work/v1/applier"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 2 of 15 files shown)


pkg/hub/mesh/controller.go
```diff
@@ -606,6 +610,18 @@ func (r *Reconciler) getClustersFromSet(ctx context.Context, clusterSetName stri
 	return clusterList.Items, nil
 }
 
+func (r *Reconciler) getManifestWorkReplicaSet(ctx context.Context, mwrsNamespace, mwrsName string) (*workv1alpha1.ManifestWorkReplicaSet, error) {
```
mkolesnik: These helpers are never called from the reconciler. For MWRS to work, the reconciler needs to:

1. Create a `ManagedClusterSetBinding` in the mesh namespace binding the mesh's ClusterSet.
   This should be singleton per ClusterSet (like operator ManifestWorks are addon-owned, not mesh-owned), since multiple meshes targeting the same ClusterSet share it.
3. Create a `Placement` referencing the ClusterSet.
4. Create the `ManifestWorkReplicaSet` referencing the Placement.
5. Clean up all three on mesh deletion (respecting the singleton pattern for the binding).

Without the reconciler integration, the helpers and RBAC are dead code.

pkg/hub/mesh/controller_test.go
```diff
@@ -110,6 +113,49 @@ func TestIsOlderMesh(t *testing.T) {
 	}
 }
 
+func TestBuildManifestWorkReplicaSet(t *testing.T) {
```
mkolesnik: This test verifies a fake client round-trip (build, store, fetch), not controller behavior.
Once the helpers are wired into the reconciler, replace this with an integration test:
create a mesh,
verify the reconciler produces the correct Placement, ManagedClusterSetBinding, and MWRS with expected labels and specs,
then verify cleanup on mesh deletion.

mkolesnik (changes requested):

pkg/hub/mesh/controller.go
```diff
@@ -139,6 +141,8 @@ func RegisterController(mgr manager.Manager) error {
 //+kubebuilder:rbac:groups=mesh.open-cluster-management.io,resources=multiclustermeshes/finalizers,verbs=update
 //+kubebuilder:rbac:groups=cluster.open-cluster-management.io,resources=managedclusters,verbs=get;list;watch
 //+kubebuilder:rbac:groups=cluster.open-cluster-management.io,resources=managedclustersets,verbs=get;list;watch
+//+kubebuilder:rbac:groups=cluster.open-cluster-management.io,resources=managedclustersetbindings,verbs=get;list;create;delete
```
mkolesnik: Missing RBAC for `managedclustersets/bind`.
OCM's admission webhook requires this to create MCSBs.

Add:
```suggestion
//+kubebuilder:rbac:groups=cluster.open-cluster-management.io,resources=managedclustersetbindings,verbs=get;list;create;delete
//+kubebuilder:rbac:groups=cluster.open-cluster-management.io,resources=managedclustersets/bind,verbs=create
```
Without this, MCSB creation passes in envtest but fails on real clusters.

pkg/hub/mesh/controller.go
```diff
@@ -872,3 +899,135 @@ func meshOwnedLabels(mesh *meshv1alpha1.MultiClusterMesh, clusterName string) ma
...
 }
+
+// ensureManagedClusterSetBinding creates a ManagedClusterSetBinding. It binds the mesh's ClusterSet in the mesh namespace
+func (r *Reconciler) ensureManagedClusterSetBinding(ctx context.Context, mesh *meshv1alpha1.MultiClusterMesh) error {
+	bindingList := &clusterv1beta2.ManagedClusterSetBindingList{}
+	if err := r.List(ctx, bindingList, client.MatchingLabels{
+		ManagedByLabel: ManagedByValue, MeshNamespaceLabel: mesh.Namespace, ClusterSetLabel: mesh.Spec.ClusterSet,
+	}); err != nil {
+		return fmt.Errorf("failed to list ManagedClusterSetBinding in namespace %s: %w", mesh.Namespace, err)
+	}
```
mkolesnik: The MCSB name is deterministic (`mesh.Spec.ClusterSet`), so Get by name is simpler and more robust than List by labels:
```go
binding := &clusterv1beta2.ManagedClusterSetBinding{}
err := r.Get(ctx, key.Of(mesh.Spec.ClusterSet, mesh.Namespace), binding)
if err == nil {
    return nil
}
if !apierrors.IsNotFound(err) {
    return fmt.Errorf("failed to get ManagedClusterSetBinding: %w", err)
}
```
This also makes the `MeshNamespaceLabel` and `ClusterSetLabel` on the MCSB unnecessary, only `ManagedByLabel` is needed for cleanup scoping.

pkg/hub/mesh/controller.go
```diff
@@ -872,3 +899,135 @@ func meshOwnedLabels(mesh *meshv1alpha1.MultiClusterMesh, clusterName string) ma
...
+		klog.Infof("Deleting ManagedClusterSetBinding %s/%s (no mesh targets this clusterSet)", binding.Namespace, binding.Name)
+		if err := client.IgnoreNotFound(r.Delete(ctx, &binding)); err != nil {
+			return fmt.Errorf("failed to delete ManagedClusterSetBinding %s/%s: %w", binding.Namespace, binding.Name, err)
+		}
+	}
+	return nil
+}
+
+// ensurePlacement creates a Placement referencing the mesh's ClusterSet
+func (r *Reconciler) ensurePlacement(ctx context.Context, mesh *meshv1alpha1.MultiClusterMesh) error {
```
mkolesnik: Please make the Placement per-mesh with an owner reference instead of shared per ClusterSet:
- Name: `mesh.Name` (not `mesh.Spec.ClusterSet`)
- Add `controllerutil.SetControllerReference(mesh, placement, r.Scheme)`
- Labels: add `MeshNameLabel` for identification
- Drop `cleanupPlacement` entirely, GC handles it

Per-mesh Placements with owner refs are simpler (no manual cleanup logic), and each mesh can independently select clusters in the future. 

The shared Placement adds complexity (the `cleanupPlacement` + `forEachMeshInClusterSet` preamble) for no benefit since all meshes targeting the same ClusterSet get the same cluster selection anyway.

pkg/hub/mesh/controller.go
```diff
@@ -872,3 +899,135 @@ func meshOwnedLabels(mesh *meshv1alpha1.MultiClusterMesh, clusterName string) ma
...
+		klog.V(4).Infof("ManagedClusterSetBinding for ClusterSet %s exists", mesh.Spec.ClusterSet)
+		return nil
+	}
+
+	if err := r.Get(ctx, key.Of(mesh.Spec.ClusterSet), &clusterv1beta2.ManagedClusterSet{}); err != nil {
+		if apierrors.IsNotFound(err) {
+			klog.V(4).Infof("ManagedClusterSet %s not found", mesh.Spec.ClusterSet)
+			return nil
+		}
+		return fmt.Errorf("failed to list ManagedClusterSet: %w", err)
```
mkolesnik: nit: This is `r.Get()`, not a List. Error message should be `"failed to get ManagedClusterSet: %w"`.
Same issue in `ensurePlacement` (L988).

pkg/hub/mesh/controller.go
```diff
@@ -44,6 +45,7 @@ import (
 const (
 	OperatorManifestWorkName   = "multicluster-mesh-operator"
 	ManifestWorkNameCacerts    = "multicluster-mesh-cacerts"
+	ManifestWorkReplicaSetName = "multicluster-mesh-mwrset"
```
mkolesnik: nit: `ManifestWorkReplicaSetName` is defined but the MWRS code isn't part of this change.

pkg/hub/mesh/controller.go
yxun: I see. I was not sure if we want to keep this naming decision in the future so I listed it by labels.
I will update this with Get by name. 
~However, I feel adding those two `MeshNamespaceLabel` and `ClusterSetLabel` labels is still good. We may have same name MCSB in different mesh namespaces~

yxun: ~For cleanup MCSB, we still need both `ManagedByLabel` and `ClusterSetLabel` labels. The reason is for the following scenario:
When a MCS has no active mesh referring it (those meshes had been deleted), we need to clean same name MCSBs in different mesh namespaces. 
Because we may not know those deleted mesh namespace's names, I used same check active mesh logic from the ManifestWork cleanup part and then listed MCSBs by those two labels.~

pkg/hub/mesh/controller.go
```diff
@@ -134,12 +137,38 @@ func RegisterController(mgr manager.Manager) error {
...
+
+	if len(cmList.Items) == 0 {
+		return fmt.Errorf("ClusterManager Object not found")
+	}
+
+	for _, cm := range cmList.Items {
+		cm.Spec.WorkConfiguration.FeatureGates = append(cm.Spec.WorkConfiguration.FeatureGates,
+			ocmoperatorv1.FeatureGate{Feature: "ManifestWorkReplicaSet", Mode: "Enable"})
+
+		if err := client.Update(ctx, &cm); err != nil {
```
sridhargaddam: I think we should first check if the featureGate is already enabled and only try enabling it if its not done already.

main.go
```diff
@@ -168,5 +172,11 @@ func runController(ctx context.Context, controllerContext *controllercmd.Control
 		return err
 	}
 
+	// Enable ManifestWorkReplicaSet feature
+	if err := meshcontroller.EnableFeatureManifestWorkReplicaSet(context.Background(), mgr.GetClient()); err != nil {
+		klog.Errorf("failed to enable ManifestWorkReplicaSet feature for hub ClusterManager: %v", err)
+		return err
+	}
+
```
sridhargaddam: This becomes a dead code as its after mgr.Start (at L170), it should be placed before starting the manager.

pkg/hub/mesh/controller.go
```diff
@@ -872,3 +916,99 @@ func meshOwnedLabels(mesh *meshv1alpha1.MultiClusterMesh, clusterName string) ma
...
+}
+
+// cleanupManagedClusterSetBinding deletes ManagedClusterSetBinding when the given ClusterSet has been deleted
+func (r *Reconciler) cleanupManagedClusterSetBinding(ctx context.Context, clusterSet string) error {
+	if err := r.Get(ctx, key.Of(clusterSet), &clusterv1beta2.ManagedClusterSet{}); err != nil {
+		if !apierrors.IsNotFound(err) {
+			return fmt.Errorf("failed to get ManagedClusterSet: %w", err)
+		}
+		bindingList := &clusterv1beta2.ManagedClusterSetBindingList{}
+		if err := r.List(ctx, bindingList, client.MatchingLabels{"metadata.name": clusterSet, ManagedByLabel: ManagedByValue}); err != nil {
```
sridhargaddam: The MCSB is created with only `ManagedByLabel: ManagedByValue`. This label query will never match any MCSB, making the entire cleanup function a no-op. You should list by ManagedByLabel and filter by `binding.Name == clusterSet`

test/integration/controller_test.go
```diff
@@ -1062,6 +1116,63 @@ func expectNoManagedServiceAccount(meshNamespace, meshName, clusterName string)
...
+		return placementList.Items
+	}).Should(BeEmpty())
+}
+
+func expectManagedClusterSetBindingDeleted(clusterSet string) {
+	Eventually(func() []clusterv1beta2.ManagedClusterSetBinding {
+		bindingList := &clusterv1beta2.ManagedClusterSetBindingList{}
+		Expect(k8sClient.List(ctx, bindingList, client.MatchingLabels{
+			meshcontroller.ManagedByLabel:  meshcontroller.ManagedByValue,
+			meshcontroller.ClusterSetLabel: clusterSet,
```
sridhargaddam: We don't use this label (i.e., ClusterSetLabel) while creating the MCSB.

test/integration/controller_test.go
```diff
@@ -1062,6 +1116,63 @@ func expectNoManagedServiceAccount(meshNamespace, meshName, clusterName string)
 	}).Should(BeTrue())
 }
 
+func expectNoManagedClusterSetBinding(clusterSet string) {
+	Consistently(func() []clusterv1beta2.ManagedClusterSetBinding {
+		bindingList := &clusterv1beta2.ManagedClusterSetBindingList{}
+		Expect(k8sClient.List(ctx, bindingList, client.MatchingLabels{"metadata.name": clusterSet})).To(Succeed())
```
sridhargaddam: Please update this as well - we are searching on "metadata.name" here.

pkg/hub/mesh/controller.go
```diff
@@ -134,12 +137,38 @@ func RegisterController(mgr manager.Manager) error {
...
+}
+
 //+kubebuilder:rbac:groups=mesh.open-cluster-management.io,resources=multiclustermeshes,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=mesh.open-cluster-management.io,resources=multiclustermeshes/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=mesh.open-cluster-management.io,resources=multiclustermeshes/finalizers,verbs=update
 //+kubebuilder:rbac:groups=cluster.open-cluster-management.io,resources=managedclusters,verbs=get;list;watch
 //+kubebuilder:rbac:groups=cluster.open-cluster-management.io,resources=managedclustersets,verbs=get;list;watch
+//+kubebuilder:rbac:groups=cluster.open-cluster-management.io,resources=managedclustersetbindings,verbs=get;list;create;delete
+//+kubebuilder:rbac:groups=cluster.open-cluster-management.io,resources=managedclustersets/bind,verbs=create
+//+kubebuilder:rbac:groups=cluster.open-cluster-management.io,resources=placements,verbs=get;list;create;update;patch;delete
```
sridhargaddam:
```suggestion
//+kubebuilder:rbac:groups=cluster.open-cluster-management.io,resources=placements,verbs=get;list;watch;create;update;patch;delete
```

pkg/hub/mesh/controller.go
```diff
@@ -134,12 +137,38 @@ func RegisterController(mgr manager.Manager) error {
...
 //+kubebuilder:rbac:groups=work.open-cluster-management.io,resources=manifestworks,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=work.open-cluster-management.io,resources=manifestworkreplicasets,verbs=get;list;create;update;patch;delete
```
sridhargaddam:
```suggestion
//+kubebuilder:rbac:groups=work.open-cluster-management.io,resources=manifestworkreplicasets,verbs=get;list;watch;create;update;patch;delete
```

pkg/hub/mesh/controller.go
```diff
@@ -872,3 +916,99 @@ func meshOwnedLabels(mesh *meshv1alpha1.MultiClusterMesh, clusterName string) ma
...
+		return fmt.Errorf("failed to get ManagedClusterSetBinding: %w", err)
+	}
+
+	if err := r.Get(ctx, key.Of(mesh.Spec.ClusterSet), &clusterv1beta2.ManagedClusterSet{}); err != nil {
+		if apierrors.IsNotFound(err) {
+			klog.V(4).Infof("ManagedClusterSet %s not found", mesh.Spec.ClusterSet)
+			return nil
+		}
+		return fmt.Errorf("failed to get ManagedClusterSet: %w", err)
+	}
```
sridhargaddam: I think this is redundant as the caller (via getClustersFromSet) already validates that the clusterSet exists.

pkg/hub/mesh/controller.go
```diff
@@ -872,3 +916,99 @@ func meshOwnedLabels(mesh *meshv1alpha1.MultiClusterMesh, clusterName string) ma
...
+
+// ensurePlacement creates a Placement referencing the mesh's ClusterSet
+func (r *Reconciler) ensurePlacement(ctx context.Context, mesh *meshv1alpha1.MultiClusterMesh) error {
+	if err := r.Get(ctx, key.Of(mesh.Spec.ClusterSet), &clusterv1beta2.ManagedClusterSet{}); err != nil {
+		if apierrors.IsNotFound(err) {
+			klog.V(4).Infof("ManagedClusterSet %s not found", mesh.Spec.ClusterSet)
+			return nil
+		}
+		return fmt.Errorf("failed to get ManagedClusterSet: %w", err)
+	}
```
sridhargaddam: Same here (i.e., redundant)

pkg/hub/mesh/endpoint_discovery.go
```diff
@@ -106,3 +111,48 @@ func (r *Reconciler) ensureManagedServiceAccountUpdated(ctx context.Context, mes
...
+
+	_, err := controllerutil.CreateOrUpdate(ctx, r.Client, mwrc, func() error {
+		mwrc.Spec.PlacementRefs = []workv1alpha1.LocalPlacementReference{{Name: placement.Name}}
+		mwrc.Spec.ManifestWorkTemplate = workTemplate
+		return nil
+	})
+
+	if err != nil {
+		return fmt.Errorf("failed to ensure ManifestWorkReplicaSet %s/%s: %w", mesh.Namespace, mwrc.Name, err)
+	}
```
sridhargaddam: I think this whole section should be re-written as follows.

```
	mwrs := &workv1alpha1.ManifestWorkReplicaSet{
		ObjectMeta: metav1.ObjectMeta{
			Name:      mesh.Name,
			Namespace: mesh.Namespace,
		},
	}

	if err := controllerutil.SetControllerReference(mesh, mwrs, r.Scheme); err != nil {
		return fmt.Errorf("failed to set owner reference for ManifestWorkReplicaSet %s: %w", mwrs.Name, err)
	}

	_, err := controllerutil.CreateOrUpdate(ctx, r.Client, mwrs, func() error {
		if mwrs.Labels == nil {
			mwrs.Labels = make(map[string]string)
		}
		mwrs.Labels[ManagedByLabel] = ManagedByValue
		mwrs.Labels[MeshNameLabel] = mesh.Name
		mwrs.Labels[MeshNamespaceLabel] = mesh.Namespace
		mwrs.Spec.PlacementRefs = []workv1alpha1.LocalPlacementReference{{Name: placement.Name}}
		mwrs.Spec.ManifestWorkTemplate = workTemplate
		return nil
	})
	if err != nil {
		return fmt.Errorf("failed to ensure ManifestWorkReplicaSet %s/%s: %w", mesh.Namespace, mwrs.Name, err)
	}

```

pkg/hub/mesh/controller.go
```diff
@@ -134,12 +137,38 @@ func RegisterController(mgr manager.Manager) error {
 		Complete(reconciler)
 }
 
+// EnableFeatureManifestWorkReplicaSet configures the cluster-manager object on the hub cluster and enables its ManifestWorkReplicaSet feature
+func EnableFeatureManifestWorkReplicaSet(ctx context.Context, client client.Client) error {
```
sridhargaddam: As discussed on slack, lets remove this API and add it as a pre-requisite.

main.go
yxun: I got client List cache error when I put this before starting the manager initially. 
I think it's related to which client we will use for updating ClusterManager.

I will resolve this discussion here since we decided not to enable this in the addon an put it as a prereq.

pkg/hub/mesh/controller.go
yxun: Hi @sridhargaddam , I checked the getClustersFromSet . It does not return error when the ClusterSet is not found. When the ClusterSet is not found , controller will doReconcile and create MCSB with non-existing ClusterSet.

After removing this , the tests (When("referencing a non-existing ClusterSet")) fails.  We may need this check there. wdyt ?

ref: https://github.com/stolostron/multicluster-mesh-addon/blob/main/pkg/hub/mesh/controller.go#L603
ref: https://github.com/stolostron/multicluster-mesh-addon/blob/main/pkg/hub/mesh/controller.go#L183

sridhargaddam: You will have to do something on the following lines ...
```
func (r *Reconciler) doReconcile(ctx context.Context, mesh *meshv1alpha1.MultiClusterMesh, clusters []clusterv1.ManagedCluster) error {
    // ... per-cluster operations ...

    clusterSetExists, err := r.clusterSetExists(ctx, mesh.Spec.ClusterSet)
    if err != nil {
        return err
    }

    if clusterSetExists {
        if err := r.ensureManagedClusterSetBinding(ctx, mesh); err != nil { ... }
        if err := r.ensurePlacement(ctx, mesh); err != nil { ... }
    }

    // ... cleanup ...
}
```

pkg/hub/mesh/controller.go
```diff
@@ -872,3 +897,106 @@ func meshOwnedLabels(mesh *meshv1alpha1.MultiClusterMesh, clusterName string) ma
...
+	for _, binding := range bindingList.Items {
+		if binding.Name == clusterSet {
+			klog.Infof("Deleting ManagedClusterSetBinding %s/%s", binding.Namespace, binding.Name)
+			if err := client.IgnoreNotFound(r.Delete(ctx, &binding)); err != nil {
+				return fmt.Errorf("failed to delete ManagedClusterSetBinding %s/%s: %w", binding.Namespace, binding.Name, err)
+			}
+		}
+	}
+	return nil
+}
```
mkolesnik: The List has no namespace scope, so it returns MCSBs from ALL namespaces. When mesh-A in ns-A is deleted, this deletes the binding in ns-B too (if it has the same ClusterSet name).

Scope to the mesh's namespace and check for other same-namespace meshes:
```suggestion
func (r *Reconciler) cleanupManagedClusterSetBinding(ctx context.Context, mesh *meshv1alpha1.MultiClusterMesh) error {
    meshList := &meshv1alpha1.MultiClusterMeshList{}
    if err := r.List(ctx, meshList, client.InNamespace(mesh.Namespace), client.MatchingFields{"spec.clusterSet": mesh.Spec.ClusterSet}); err != nil {
        return fmt.Errorf("failed to list meshes for ClusterSet %s in namespace %s: %w", mesh.Spec.ClusterSet, mesh.Namespace, err)
    }

    for i := range meshList.Items {
        other := &meshList.Items[i]
        if other.UID != mesh.UID && other.DeletionTimestamp.IsZero() {
			klog.V(4).Infof("ManagedClusterSetBinding %s/%s still needed by mesh %s/%s", mesh.Namespace, mesh.Spec.ClusterSet, other.Namespace, other.Name)
            return nil
        }
    }

    binding := &clusterv1beta2.ManagedClusterSetBinding{
        ObjectMeta: metav1.ObjectMeta{
            Name:      mesh.Spec.ClusterSet,
            Namespace: mesh.Namespace,
        },
    }

	klog.Infof("Deleting ManagedClusterSetBinding %s/%s", mesh.Namespace, mesh.Spec.ClusterSet)
    return client.IgnoreNotFound(r.Delete(ctx, binding))
}
```
This also changes the signature from `clusterSet string` to `mesh *meshv1alpha1.MultiClusterMesh`.

test/integration/controller_test.go
```diff
@@ -1062,6 +1120,63 @@ func expectNoManagedServiceAccount(meshNamespace, meshName, clusterName string)
...
+func expectNoPlacement(clusterSet string) {
+	Consistently(func() []clusterv1beta1.Placement {
+		placementList := &clusterv1beta1.PlacementList{}
+		Expect(k8sClient.List(ctx, placementList, client.MatchingLabels{
+			meshcontroller.ManagedByLabel:  meshcontroller.ManagedByValue,
+			meshcontroller.ClusterSetLabel: clusterSet,
+		})).To(Succeed())
+		return placementList.Items
+	}).Should(BeEmpty())
+}
```
mkolesnik: Since each test uses a unique namespace, drop all label filters and just list in the namespace:
```go
func expectNoPlacement() {
    Consistently(func() []clusterv1beta1.Placement {
        placementList := &clusterv1beta1.PlacementList{}
        Expect(k8sClient.List(ctx, placementList, client.InNamespace(testNs))).To(Succeed())
        return placementList.Items
    }).Should(BeEmpty())
}
```
Same for `expectNoManagedClusterSetBinding`, `expectPlacementDeleted`, and `expectManagedClusterSetBindingDeleted`. No label filters needed when the namespace is unique per test.

pkg/hub/mesh/controller.go
```diff
@@ -139,6 +140,9 @@ func RegisterController(mgr manager.Manager) error {
 //+kubebuilder:rbac:groups=mesh.open-cluster-management.io,resources=multiclustermeshes/finalizers,verbs=update
 //+kubebuilder:rbac:groups=cluster.open-cluster-management.io,resources=managedclusters,verbs=get;list;watch
 //+kubebuilder:rbac:groups=cluster.open-cluster-management.io,resources=managedclustersets,verbs=get;list;watch
+//+kubebuilder:rbac:groups=cluster.open-cluster-management.io,resources=managedclustersetbindings,verbs=get;list;create;delete
+//+kubebuilder:rbac:groups=cluster.open-cluster-management.io,resources=managedclustersets/bind,verbs=create
+//+kubebuilder:rbac:groups=cluster.open-cluster-management.io,resources=placements,verbs=get;list;watch;create;update;patch;delete
```
mkolesnik: nit: `watch` and `patch` are unused for placements. If keeping `CreateOrUpdate`, the minimum is `get;list;create;update;delete`. If simplifying to Get-then-Create, `get;list;create;delete` suffices.

pkg/hub/mesh/controller.go
```diff
@@ -872,3 +897,106 @@ func meshOwnedLabels(mesh *meshv1alpha1.MultiClusterMesh, clusterName string) ma
...
+
+	_, err := controllerutil.CreateOrUpdate(ctx, r.Client, placement, func() error {
+		placement.Labels[ClusterSetLabel] = mesh.Spec.ClusterSet
+		placement.Spec.ClusterSets = []string{mesh.Spec.ClusterSet}
+		return nil
+	})
+	if err != nil {
+		return fmt.Errorf("failed to ensure placement %s: %w", placement.Name, err)
+	}
+	return nil
```
```suggestion
func (r *Reconciler) ensurePlacement(ctx context.Context, mesh *meshv1alpha1.MultiClusterMesh) error {
	placement := &clusterv1beta1.Placement{
		ObjectMeta: metav1.ObjectMeta{Name: mesh.Name, Namespace: mesh.Namespace},
	}
	_, err := controllerutil.CreateOrUpdate(ctx, r.Client, placement, func() error {
		if placement.Labels == nil {
			placement.Labels = make(map[string]string)
		}
		placement.Labels[ManagedByLabel] = ManagedByValue
		placement.Labels[MeshNameLabel] = mesh.Name
		placement.Labels[MeshNamespaceLabel] = mesh.Namespace
		placement.Spec.ClusterSets = []string{mesh.Spec.ClusterSet}
		return controllerutil.SetControllerReference(mesh, placement, r.Scheme)
	})
	if err != nil {
		return fmt.Errorf("failed to ensure placement %s: %w", placement.Name, err)
	}
	return nil
}
```
`CreateOrUpdate` needs Name/Namespace to do the `Get`, but all desired state (labels, spec, owner ref) belongs in the mutate func. The mutator always runs (both create and update paths), so this restores drift on every reconcile.

test/integration/controller_test.go
```diff
@@ -842,34 +890,38 @@ var _ = Describe("MultiClusterMesh Controller", func() {
...
 				util.ExpectResourceDeleted(ctx, k8sClient, &msav1beta1.ManagedServiceAccount{},
 					expectedManagedServiceAccountName(testNs, meshName), clusterName)
 
 				Consistently(func(g Gomega) {
-					getManagedServiceAccount(g, otherNs, otherMesh, clusterName)
+					getManagedServiceAccount(g, testNs, otherMesh, clusterName)
 				}).Should(Succeed())
 			})
+
+			It("should keep the ManagedClusterSetBinding when only one mesh is deleted", func() {
```
mkolesnik, line 917: `expectManagedClusterSetBinding` uses `Eventually`, which only proves the binding exists at some point after deletion. It doesn't catch a delete-then-recreate race. Use `Consistently` to verify the binding persists:
```suggestion
It("should keep the ManagedClusterSetBinding when only one mesh is deleted", func() {
    binding := expectManagedClusterSetBinding(testNs, testClusterSet)
    rv := binding.ResourceVersion
    util.DeleteResource(ctx, k8sClient, &meshv1alpha1.MultiClusterMesh{}, meshName, testNs)
    Consistently(func(g Gomega) {
        b := &clusterv1beta2.ManagedClusterSetBinding{}
        g.Expect(k8sClient.Get(ctx, key.Of(testClusterSet, testNs), b)).To(Succeed())
        g.Expect(b.ResourceVersion).To(Equal(rv))
    }).Should(Succeed())
})
```

pkg/hub/mesh/controller.go
```diff
@@ -872,3 +897,106 @@ func meshOwnedLabels(mesh *meshv1alpha1.MultiClusterMesh, clusterName string) ma
...
+				ClusterSetLabel:    mesh.Spec.ClusterSet,
+			},
+		},
+		Spec: clusterv1beta1.PlacementSpec{
+			ClusterSets: []string{mesh.Spec.ClusterSet},
+		},
+	}
+
+	if err := controllerutil.SetControllerReference(mesh, placement, r.Scheme); err != nil {
+		return fmt.Errorf("failed to set owner reference for placement %s:%w", placement.Name, err)
```
mkolesnik: nit:
```suggestion
		return fmt.Errorf("failed to set owner reference for placement %s: %w", placement.Name, err)
```

pkg/hub/mesh/controller.go
```diff
@@ -872,3 +897,106 @@ func meshOwnedLabels(mesh *meshv1alpha1.MultiClusterMesh, clusterName string) ma
...
+func (r *Reconciler) ensurePlacement(ctx context.Context, mesh *meshv1alpha1.MultiClusterMesh) error {
+	placement := &clusterv1beta1.Placement{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      mesh.Name,
+			Namespace: mesh.Namespace,
+			Labels: map[string]string{
+				ManagedByLabel:     ManagedByValue,
+				MeshNameLabel:      mesh.Name,
+				MeshNamespaceLabel: mesh.Namespace,
+				ClusterSetLabel:    mesh.Spec.ClusterSet,
```
```suggestion
```
`ClusterSetLabel` is redundant on the Placement. It's per-mesh, named `mesh.Name`, owned by the mesh (GC handles deletion), and the ClusterSet is already in `Spec.ClusterSets`.
The MCSB had this label removed per earlier review.

Drop it here too for consistency, and simplify the test helpers that filter by it.

pkg/hub/mesh/controller.go
```diff
@@ -303,11 +307,28 @@ func (r *Reconciler) doReconcile(ctx context.Context, mesh *meshv1alpha1.MultiCl
...
+			return fmt.Errorf("failed to ensure ManagedClusterSetBinding for mesh %s binding %s: %w", mesh.Name, mesh.Spec.ClusterSet, err)
+		}
+		if err := r.ensurePlacement(ctx, mesh); err != nil {
+			return fmt.Errorf("failed to ensure Placement for mesh %s binding %s: %w", mesh.Name, mesh.Spec.ClusterSet, err)
+		}
+	} else {
+		if err := r.cleanupManagedClusterSetBinding(ctx, mesh); err != nil {
+			return fmt.Errorf("failed to cleanup ManagedClusterSetBinding: %w", err)
+		}
+	}
```
mkolesnik: Consider removing this `else` branch. A binding pointing to a non-existent ClusterSet is harmless and gets cleaned up when the last mesh is deleted (`handleDeletion`). This branch also behaves inconsistently: it deletes for a single mesh but skips for two meshes (cleanup finds the other and returns early).

The corresponding test ("should cleanup ManagedClusterSetBinding and Placement when the ClusterSet is deleted") can be removed too. The `expectPlacementDeleted` assertion is wrong anyway since the Placement is owned by the mesh, not the ClusterSet.

pkg/hub/mesh/controller.go
```diff
@@ -139,6 +140,9 @@ func RegisterController(mgr manager.Manager) error {
 //+kubebuilder:rbac:groups=mesh.open-cluster-management.io,resources=multiclustermeshes/finalizers,verbs=update
 //+kubebuilder:rbac:groups=cluster.open-cluster-management.io,resources=managedclusters,verbs=get;list;watch
 //+kubebuilder:rbac:groups=cluster.open-cluster-management.io,resources=managedclustersets,verbs=get;list;watch
+//+kubebuilder:rbac:groups=cluster.open-cluster-management.io,resources=managedclustersetbindings,verbs=get;list;create;delete
+//+kubebuilder:rbac:groups=cluster.open-cluster-management.io,resources=managedclustersets/bind,verbs=create
+//+kubebuilder:rbac:groups=cluster.open-cluster-management.io,resources=placements,verbs=get;list;create;update;delete
```
mkolesnik: I tested this on the dev-env.
After creating a mesh, the controller logs show continuous errors that never stop:
```
E0803 "Failed to watch" err="managedclustersetbindings ... cannot watch resource"
E0803 "Failed to watch" err="placements ... cannot watch resource"
```

Also, the controller never deletes Placements directly (relies on GC via owner reference). `delete` is unused here. 

Add `watch` to both and remove `delete` from placement:
```suggestion
//+kubebuilder:rbac:groups=cluster.open-cluster-management.io,resources=managedclustersetbindings,verbs=get;list;watch;create;delete
//+kubebuilder:rbac:groups=cluster.open-cluster-management.io,resources=managedclustersets/bind,verbs=create
//+kubebuilder:rbac:groups=cluster.open-cluster-management.io,resources=placements,verbs=get;list;watch;create;update
```

test/integration/controller_test.go
```diff
@@ -842,32 +885,41 @@ var _ = Describe("MultiClusterMesh Controller", func() {
...
 				util.ExpectResourceDeleted(ctx, k8sClient, &msav1beta1.ManagedServiceAccount{},
 					expectedManagedServiceAccountName(testNs, meshName), clusterName)
 
 				Consistently(func(g Gomega) {
-					getManagedServiceAccount(g, otherNs, otherMesh, clusterName)
+					getManagedServiceAccount(g, testNs, otherMesh, clusterName)
+				}).Should(Succeed())
+			})
+
+			It("should keep the ManagedClusterSetBinding when only one mesh is deleted", func() {
```
mkolesnik, line 917: This tests that the MCSB survives when one mesh is deleted.
There should be a companion test verifying the MCSB IS deleted when the last mesh is deleted.

The `expectManagedClusterSetBindingDeleted` helper already exists for this.

Something like:

```go
It("should delete the ManagedClusterSetBinding when all meshes are deleted", func() {
	expectManagedClusterSetBinding(testNs, testClusterSet)
	util.DeleteResource(ctx, k8sClient, &meshv1alpha1.MultiClusterMesh{}, otherMesh, testNs)
	expectManagedClusterSetBindingDeleted(testNs)
})
```

test/integration/controller_test.go
```diff
@@ -785,6 +786,21 @@ var _ = Describe("MultiClusterMesh Controller", func() {
...
 
+			It("should create ManagedClusterSetBinding for the ManagedClusterSet", func() {
+				util.CreateMultiClusterMesh(ctx, k8sClient, meshName, testNs, testClusterSet)
+				clusterSetBinding := expectManagedClusterSetBinding(testNs, testClusterSet)
+				Expect(clusterSetBinding.Spec.ClusterSet).To(Equal(testClusterSet))
+			})
+
+			It("should create Placement for the ManagedClusterSet", func() {
+				util.CreateMultiClusterMesh(ctx, k8sClient, meshName, testNs, testClusterSet)
+				placement := expectPlacement(meshName, testNs)
```
mkolesnik, line 797: The owner reference is the mechanism that enables GC-based cleanup for Placement (there's no explicit cleanup in `handleDeletion`).
Verifying it here documents the intent and catches regressions.
```suggestion
				placement := expectPlacement(meshName, testNs)
				Expect(placement.OwnerReferences).To(HaveLen(1))
				Expect(placement.OwnerReferences[0].Name).To(Equal(meshName))
```

pkg/hub/mesh/controller.go
```diff
@@ -872,3 +893,83 @@ func meshOwnedLabels(mesh *meshv1alpha1.MultiClusterMesh, clusterName string) ma
...
+			},
+		},
+		Spec: clusterv1beta2.ManagedClusterSetBindingSpec{
+			ClusterSet: mesh.Spec.ClusterSet,
+		},
+	}
+	return r.Create(ctx, clusterSetBinding)
+}
+
+// cleanupManagedClusterSetBinding deletes ManagedClusterSetBinding when the given ClusterSet is deleted or no mesh targets it
```
mkolesnik: nit: This comment says "when the given ClusterSet is deleted or no mesh targets it" but the ClusterSet-deletion path was removed. 

The function is now only called from `handleDeletion`.
```suggestion
// cleanupManagedClusterSetBinding deletes ManagedClusterSetBinding when no other mesh in the namespace targets the same ClusterSet
```

pkg/hub/mesh/controller.go
```diff
@@ -303,11 +307,24 @@ func (r *Reconciler) doReconcile(ctx context.Context, mesh *meshv1alpha1.MultiCl
...
+	if err != nil {
+		return fmt.Errorf("failed to check ManagedClusterSet %s: %w", mesh.Spec.ClusterSet, err)
+	}
+
+	if clusterSetExists {
+		if err := r.ensureManagedClusterSetBinding(ctx, mesh); err != nil {
+			return fmt.Errorf("failed to ensure ManagedClusterSetBinding for mesh %s binding %s: %w", mesh.Name, mesh.Spec.ClusterSet, err)
+		}
+		if err := r.ensurePlacement(ctx, mesh); err != nil {
+			return fmt.Errorf("failed to ensure Placement for mesh %s binding %s: %w", mesh.Name, mesh.Spec.ClusterSet, err)
```
mkolesnik: nit: "binding %s" is copy-pasted from the MCSB line above.

For Placement, `mesh.Name` alone is sufficient since the Placement name equals the mesh name:
```suggestion
			return fmt.Errorf("failed to ensure Placement for mesh %s/%s: %w", mesh.Namespace, mesh.Name, err)
```

mkolesnik (approved):

PR #223, opened 2026-07-16
Delete kind clusters sequentially as part of dev-clean

sridhargaddam:
kind delete cluster uses a lockfile when updating ~/.kube/config. 
If multiple clusters are deleted in parallel, they can race for the lock
and fail. Deleting clusters one at a time is fast enough, so there's little
value in parallelizing this step.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -337,7 +337,7 @@ dev-clean-meshes: ## Delete all mesh resources, cert-manager trust chain, and me
 
 .PHONY: dev-clean
 dev-clean: ## Destroy dev clusters and remove .kube/ folder
-	$(MAKE) --no-print-directory -j$(PARALLEL) --output-sync=line $(addprefix delete-,$(CLUSTERS))
+	$(MAKE) --no-print-directory -j1 $(addprefix delete-,$(CLUSTERS))
 	rm -rf $(DEV_KUBE_DIR)
 	$(call log,Dev environment cleaned)
 
--- a/Makefile
+++ b/Makefile
@@ -337,10 +337,10 @@ dev-clean-meshes: ## Delete all mesh resources, cert-manager trust chain, and me
 
 .PHONY: dev-clean
 dev-clean: ## Destroy dev clusters and remove .kube/ folder
-	$(MAKE) --no-print-directory -j1 $(addprefix delete-,$(CLUSTERS))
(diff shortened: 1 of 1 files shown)


sridhargaddam:
I was running into the following error without the fix:

```
$: make dev-clean
make --no-print-directory -j8 --output-sync=line delete-hub delete-cluster1 delete-cluster2
/home/sgaddam/goprojects/downstream/multicluster-mesh-addon/bin/kind delete cluster --name hub
Deleting cluster "hub" ...
Deleted nodes: ["hub-control-plane"]
/home/sgaddam/goprojects/downstream/multicluster-mesh-addon/bin/kind delete cluster --name cluster2
Deleting cluster "cluster2" ...
failed to update kubeconfig: failed to lock config file: open /home/sgaddam/.kube/config.lock: file exists
Deleted nodes: ["cluster2-control-plane"]
ERROR: failed to delete cluster "cluster2": failed to lock config file: open /home/sgaddam/.kube/config.lock: file exists
make[1]: *** [Makefile:346: delete-cluster2] Error 1
make[1]: *** Waiting for unfinished jobs....
/home/sgaddam/goprojects/downstream/multicluster-mesh-addon/bin/kind delete cluster --name cluster1
Deleting cluster "cluster1" ...
Deleted nodes: ["cluster1-control-plane"]
make: *** [Makefile:340: dev-clean] Error 2
```

mkolesnik (changes requested):
Actually if you pass kubeconfig file explicitly (via `--kubeconfig`) then it won't contend, so we can just pass `--kubeconfig $(DEV_KUBE_DIR)/$*.config` in the actual delete command and it should work fine with parallel runs.

sridhargaddam:
> Actually if you pass kubeconfig file explicitly (via `--kubeconfig`) then it won't contend, so we can just pass `--kubeconfig $(DEV_KUBE_DIR)/$*.config` in the actual delete command and it should work fine with parallel runs.

Honestly, the time difference when deleting the clusters is negligible whether we do it sequentially or in parallel. Anyway, let me update it.

mkolesnik (approved):

mkolesnik:
> Honestly, the time difference when deleting the clusters is negligible whether we do it sequentially or in parallel. Anyway, let me update it.

Yes I know its not much but 10s is still better than 30s if it's "free" :)

PR #224, opened 2026-07-17
Block dependent PRs until dependencies merge

sridhargaddam:
This PR adds a GitHub Action (using similar approach as [submariner](https://github.com/submariner-io/submariner/blob/devel/.github/workflows/dependent-issues.yml))
that fails when a PR lists unmerged Depends-On targets, so Tide
won't merge it early.

### DIFF
--- a/.github/workflows/dependent-issues.yml
+++ b/.github/workflows/dependent-issues.yml
@@ -1,0 +1,24 @@
+---
+name: PR Dependencies
+
+on:
+  pull_request_target:
+    types:
+      - opened
+      - edited
+      - reopened
+      - synchronize
+
+permissions:
+  pull-requests: write
+
+jobs:
+  check:
+    name: Check Dependencies
+    if: github.repository_owner == 'stolostron'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: depends-on/depends-on-action@c7ac9a6932a7069e83aef80c202d9f60f91e43fd # v0.17.0
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          check-unmerged-pr: true
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -41,6 +41,16 @@ Contact the org [owners](https://github.com/orgs/stolostron/people?query=role%3A
 - Avoid force-pushing during review if possible. Use merge commits to resolve conflicts.
 - Use the `/hold` and `/unhold` comments to control when a PR is eligible for merge. Contributors may use `/hold` to prevent a PR from merging while additional review or discussion is in progress. This is particularly useful for complex changes that require multiple approvals. If a PR is not on hold, it becomes eligible for merge after receiving a single approving review from a maintainer.
 
+### Dependent Pull Requests
+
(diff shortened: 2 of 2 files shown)


.github/workflows/dependent-issues.yml
```diff
@@ -0,0 +1,24 @@
+---
+name: PR Dependencies
+
+on:
+  pull_request_target:
+    types:
+      - opened
+      - edited
+      - reopened
+      - synchronize
```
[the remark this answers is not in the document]
sridhargaddam: Let me add a comment to the CONTRIBUTING.md file

CONTRIBUTING.md
```diff
@@ -41,6 +41,18 @@ Contact the org [owners](https://github.com/orgs/stolostron/people?query=role%3A
 - Avoid force-pushing during review if possible. Use merge commits to resolve conflicts.
 - Use the `/hold` and `/unhold` comments to control when a PR is eligible for merge. Contributors may use `/hold` to prevent a PR from merging while additional review or discussion is in progress. This is particularly useful for complex changes that require multiple approvals. If a PR is not on hold, it becomes eligible for merge after receiving a single approving review from a maintainer.
 
+### Dependent Pull Requests
+
+If a pull request depends on another open pull request in this repository, declare the dependency in the PR description using a full PR URL:
+
+```text
+Depends-On: https://github.com/stolostron/multicluster-mesh-addon/pull/123
```
mkolesnik, line 49: Do you know if short pr links are supported (eg #123)?

If so I'd use that as the example (and remote link only for external repo, though I'm not sure if this is even supported or if it's the intended use)

CONTRIBUTING.md
```diff
@@ -41,6 +41,18 @@ Contact the org [owners](https://github.com/orgs/stolostron/people?query=role%3A
...
+```
+
+You can list multiple dependencies, one `Depends-On:` line each. The **PR Dependencies / Check Dependencies** GitHub Action fails until every listed pull request is merged, which keeps Tide from merging the dependent PR early. Use `/hold` for other reasons to block merge (WIP, discussion); dependency gating is automatic when `Depends-On:` is present.
+
+The check re-runs when the dependent PR is updated (new commits, editing description etc). It does **not** refresh automatically when a listed dependency merges. After dependencies land, re-run **PR Dependencies / Check Dependencies** on the waiting PR (or push a commit / edit the description) so the check can turn green.
```
mkolesnik: This is partially correct, but if its a local dependency theres a good chance the parent pr gets squash merged so the dependent PR would need to be updated anyway.

I'd recommend to merge main to the PR once the parent lands, this should solve both concerns AFAICT.

mkolesnik (commented):
Thanks Sridhar, a couple of comments

CONTRIBUTING.md
sridhargaddam: I looked into the [GitHub action](https://github.com/depends-on/depends-on-action), and it appears that it only supports full PR URLs, not short PR links.

FilipB (approved):

PR #227, opened 2026-08-03
Add approach comparison table to design doc

mkolesnik:
Compare three approaches to multi-cluster Istio mesh management (manual, OCM policies, addon controller) across setup complexity, trust/credential management, safety, scaling, flexibility, and user experience.

The table reflects each approach at its full intended scope, with a phased scope note clarifying current Phase 1 coverage.

### DIFF
--- a/docs/design.md
+++ b/docs/design.md
@@ -3,6 +3,7 @@
 ## Table of Contents
 
 - [Overview](#overview)
+- [Approach Comparison](#approach-comparison)
 - [Architecture](#architecture)
 - [Scope](#scope)
 - [Supported Topologies](#supported-topologies)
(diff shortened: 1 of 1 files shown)


docs/design.md
```diff
@@ -24,6 +25,96 @@ The OCM Service Mesh Add-on automates multi-cluster Istio service mesh setup via
...
+| **Debugging / troubleshooting** | **High effort.** User debugs across N clusters independently with no centralized view. | **Medium effort.** Compliance reporting aggregates state and identifies non-compliant clusters. If deeper debugging is needed, the user traces through the policy stack. | **Low effort.** Per-cluster status conditions on a single CR with specific error reasons. If the user needs to dig deeper into the distribution mechanism, the complexity is similar to manual policies. |
+| **Prerequisites** | **Minimal.** `kubectl`, `openssl`/`cfssl`, `istioctl`. Tools most users already have. | **OCM policy framework** + `ManagedServiceAccount` addon. cert-manager optional (needed for automated per-cluster CA generation, otherwise user provides CA material manually). | **cert-manager** + `ManagedServiceAccount` addon + the addon itself. |
+
+### Summary
```
FilipB, line 99: I don't have much context why we need this comparison but it's obvious that dedicated add-on will be always better than some general usage tool like OCM Policies. The main question is probably if the amount of time required to develop and maintain dedicated add-on is worth the advantages over OCM policies. Or what was the reason to create this?

longmuir: A dedicated add-on would always be better if it can cover every possible use case and is cheap to extend and maintain. I don't think that's going to be the case though. We need to consider the cost to build, maintain and extend the addon over time - as both ACM and OSSM continue to evolve. I don't see these factors captured here.

In practice, I don't think this should be viewed as three separate options - users may already be doing some of the first two option, and an addon should aim to help with the things that are most challenging at scale, while giving users flexibility to customize for the cases the addon can't handle. 

It is true that you don't need an addon if you are comfortable rolling a lot of your own configuration - ACM helps a lot, and we should take advantage of that, but the thinking with the addon is that it would provide more native functions (make it easier) to handle the "mesh at scale" use cases - which customers have asked OSSM to provide more assistance with (because Solo/Tetrate do provide this). It was always meant to be a thin layer over OSSM/ACM though - with a primary consideration that it is easy for us to maintain.

mkolesnik: The comparison came up because we need to articulate why a dedicated addon vs OCM policies for mesh management.
The table documents the tradeoffs for anyone evaluating the approaches, **including us**.

You're right that a dedicated addon will always look better in a feature comparison, that's kind of the point of building one.
The table tries to be fair about where policies are genuinely strong (compliance reporting, drift detection, flexibility for non-standard topologies).

The question how much is this going to cost us is not part of this because this captures the cost to users (e.g. complexity, effort, risk, etc).

Obviously, implementing an addon costs time and resources to do.
How much is it going to be is hard to estimate, and really depends on how much effort we put in and where.

Also I don't think they necessarily exclude each other, obviously users can do the other two options and there's a summary what approach fits where.
I believe the addon does bring considerable value to the users, and that was what I was trying to capture here.

docs/design.md
```diff
@@ -24,6 +25,96 @@ The OCM Service Mesh Add-on automates multi-cluster Istio service mesh setup via
 
 Without this add-on, multi-cluster mesh setup is a manual process involving certificate management, O(N^2) secret exchanges, and per-cluster operator configuration.
 
+## Approach Comparison
```
sridhargaddam, line 28: As discussed during the call, we have three broad use-cases

1. Single multi-cluster mesh.
2. Independent mesh per cluster (fleet stamping).
3. Multiple multi-cluster meshes (HA/DR pairs)

The comparison evaluates the three approaches in the abstract, but different business use-cases favor different approaches. Lets consider adding a "Which approach fits which scenario?" subsection that maps to concrete use cases. For example...

- Single multi-cluster mesh (subset of fleet, one mesh): The addon's trust and discovery automation provide the clearest value here.

- Fleet stamping (independent mesh per cluster, no cross-cluster connectivity): OCM Policies are the natural fit for this use-case. The addon's core differentiators (shared trust, N-to-N discovery) aren't needed, and requiring a ClusterSet per single-cluster mesh adds overhead without any benefit.

- Multiple multi-cluster meshes at scale (e.g., 20 HA/DR pairs across 100 clusters): Exposes the ClusterSet-per-mesh limitation (20 ClusterSets for 20 pairs is organizational overhead that Placement-based selection would avoid). 

Without this mapping, the comparison reads as though the addon is universally better, but Use Case 2 is better served by policies alone IMO.

docs/design.md
sridhargaddam: The summary (at L103) recommends the addon for "production and semi-production environments" and relegates policies to "single-mesh deployments." IMO, this framing is too broad.

For fleet stamping (independent mesh per cluster with no cross-cluster connectivity), OCM Policies are the better tool and the addon's core value (shared trust, N-to-N discovery) doesn't apply, and requiring a ClusterSet per single-cluster mesh creates unnecessary overhead.

Consider framing the summary around use-case fit rather than a blanket recommendation. 

Something like:

- Cross-cluster mesh (shared trust domain, N-to-N discovery): the addon provides the most value
- Fleet stamping (independent, per-cluster mesh): OCM Policies are the natural fit
- Multiple multi-cluster meshes at fleet scale: It could be a combination of both.

This will also be aligned with @longmuir point that these approaches are complementary, not competing.

docs/design.md
```diff
@@ -24,6 +25,96 @@ The OCM Service Mesh Add-on automates multi-cluster Istio service mesh setup via
...
+
+### Scaling and Multi-Tenancy
+
+| Concern | Manual | OCM Policies | Addon Controller |
+|---------|--------|--------------|------------------|
+| **Operational complexity at scale** | **Untenable.** Istio docs say "managing configuration across multiple clusters at scale is challenging." Manageable for a handful of clusters, impractical beyond that. | **Manageable for single mesh.** User effort stays constant per mesh regardless of cluster count. Multi-mesh multiplies the policy surface area the user maintains and debugs. | **Constant.** One CR per mesh regardless of cluster count or mesh count. |
```
sridhargaddam, line 80: The "Constant. One CR per mesh regardless of cluster count or mesh count" claim for the addon is accurate per-mesh, but it doesn't account for the ClusterSet overhead. For the HA/DR pairs use case (e.g., 20 pairs across 100 clusters), the user needs 20 ClusterSets + 20 CRs. The operational complexity isn't constant with respect to mesh count, it's linear in ClusterSets, which require their own lifecycle management.

docs/design.md
rcernich: I'm going to ask a stupid question: Is there any reason the add-on couldn't leverage these other resources as well, to help ensure things are configured correctly for the user?  Just because we're leveraging existing mechanisms within ACM doesn't mean we can't enhance how they are rolled out.  We should also be considering how different features are exposed to the user.  For example, Istio doesn't reimplement Pod and Service, but makes use of them when configuring the mesh.  (Not sure if this analogy works, but...)

sridhargaddam: Yes, I agree that the addon should leverage ACM capabilities where feasible and provide a unified API that delivers the best possible user experience for the scenarios it is intended to address.

At the same time, I feel that if ACM supports a capability that is independently well-suited to a particular use case but isn’t aligned with the primary goals of the addon, it may be better to keep it outside the scope of the addon. Fleet management (without any cross-cluster connectivity) is one example that comes to my mind, although I’m happy to discuss this aspect further and hear others’ perspectives.

mkolesnik: @rcernich yes that's exactly the plan, and in fact what we're doing.

The addon manages the resources for the users automatically, it doesn't reinvent things, it abstracts a commong configuration API that allows the users to do the same work in a much more controlled way which also makes sure to avoid common problems - that is the gist of the comparison.

@sridhargaddam I agree that some scenarios, as outlined by @longmuir, are better served by direct OCM/ACM tooling and not by the addon.
The same as, as a K8s user, I wouldn't deploy a service just to end up accessing pod IPs directly, I also wouldn't deploy a single-cluster mesh with the addon (no matter how many copies of it I choose to deploy).

docs/design.md
mkolesnik: It's the same in this regard for mesh/policy both will need to target the clusters somehow.
I'll update this to clarify the point.

docs/design.md
mkolesnik: Fleet stamping is out of scope for this, it should be clear enough but I'll add explanation to clarify that if multi-cluster mesh is not required this is out of scope for the addon.

> 20 ClusterSets for 20 pairs is organizational overhead that Placement-based selection would avoid

This is not correct, you still have the operational complexity of managing the "deployment" whether you use Placements or ClusterSets.

> Without this mapping, the comparison reads as though the addon is universally better, but Use Case 2 is better served by policies alone IMO.

Usually yes, it is better. The policies are a good fit for small single-mesh (multi-cluster) deployments. Any large scale deployments become harder to manage with the policy alone.

mkolesnik:
Pushed updates addressing the feedback:

- Added explicit scope statement at the top of the comparison: the addon is for multi-cluster mesh (2+ clusters) only, single-cluster mesh is out of scope regardless of how many copies you deploy.
- Clarified "single mesh" references throughout to say "single multi-cluster mesh" so there's no ambiguity.
- Updated the scaling row to note policy collision risk for multi-mesh and clarified the addon's per-mesh overhead.
- Simplified the remote secret formula to 2N.

@sridhargaddam agreed that fleet stamping is not what the addon is for. The scoping statement should make that clear now. On the ClusterSet overhead point, the targeting cost is the same for any OCM-based approach, updated the table to reflect that.

Also merged upstream/main, conflict resolved cleanly (design.md moved to docs/dev/).

PR #229, opened 2026-08-06
Collect e2e test artifacts before cleanup

FilipB:
Add artifact collection to both e2e test suites so pod state, events, and container logs are captured before AfterAll cleanup runs. This preserves diagnostic info for debugging failures in CI where the cluster is torn down after the run.


.github/workflows/pr-e2e.yml
```diff
@@ -50,3 +50,14 @@ jobs:
...
+          ARTIFACT_DIR: ${{ github.workspace }}/_artifacts
+
+      - name: Upload test artifacts
+        if: always()
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
+        with:
+          name: e2e-artifacts
+          path: _artifacts/
+          if-no-files-found: ignore
+          retention-days: 14
```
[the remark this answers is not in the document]
FilipB: If we miss artifacts in this corner case, no big deal.

test/util/e2e_client.go
```diff
@@ -66,6 +67,55 @@ func (c *E2EClient) Cleanup(ctx context.Context) {
...
+		"-o", "jsonpath={.items[*].metadata.name}")
+	out, err := cmd.Output()
+	if err != nil || len(out) == 0 {
+		return
+	}
+	for pod := range strings.FieldsSeq(string(out)) {
+		c.kubectlToFile(ctx, dir, pod+".log",
+			"logs", "--all-containers", "-n", namespace, pod)
+	}
+}
```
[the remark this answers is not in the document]
FilipB: No need to have unit tests for test code.

FilipB:
@coderabbitai help

FilipB:
@coderabbitai run pre-merge checks

test/e2e/mesh_lifecycle_test.go
```diff
@@ -86,6 +87,17 @@ var _ = Describe("MultiClusterMesh lifecycle", Ordered, func() {
 	})
 
 	AfterAll(func(ctx SpecContext) {
+		dir := artifactDir("mesh-lifecycle")
+		Step("Collecting artifacts to %s", dir)
+		hubDir := filepath.Join(dir, "hub")
+		hubClient.CollectArtifacts(ctx, hubDir, ns, controllerNamespace)
+		// ManifestWorks are excluded — they can embed Secrets (e.g. cacerts)
```
sridhargaddam: Since not all ManifestWorks have the secrets, how about dumping the manifestWorks related to installation of the operator and controlPlane namespace?

test/e2e/mesh_lifecycle_test.go
```diff
@@ -86,6 +87,17 @@ var _ = Describe("MultiClusterMesh lifecycle", Ordered, func() {
...
+		hubClient.DumpResource(ctx, hubDir, "multiclustermeshes")
```
sridhargaddam: As you know, the manifestWorks are created on the Hub cluster in the cluster named namespaces (i.e., cluster1, cluster2 etc). It might be useful to collect them.

test/e2e/multi_primary_test.go
```diff
@@ -101,6 +101,17 @@ var _ = Describe("Multi-primary data plane", Ordered, Serial, func() {
...
+		Step("Collecting artifacts to %s", dir)
+		hubDir := filepath.Join(dir, "hub")
+		hubClient.CollectArtifacts(ctx, hubDir, meshNS, "multicluster-mesh-system")
+		// ManifestWorks are excluded — they can embed Secrets (e.g. cacerts)
+		hubClient.DumpResource(ctx, hubDir, "multiclustermeshes")
+		for name, spokeClient := range spokeClients {
+			spokeClient.CollectArtifacts(ctx, filepath.Join(dir, name),
+				cpNamespace, testOperatorNamespace, sampleNS, "istio-cni")
+		}
+
```
sridhargaddam: Since this is identical to the one in the test/e2e/mesh_lifecycle_test.go, we can probably create an API in the test/e2e/suite_test.go and pass necessary arguments like `artifactDir`

test/e2e/mesh_lifecycle_test.go
mkolesnik: I'd reconsider excluding MWs entirely.
This is a test environment, the secrets are ephemeral, and MW state is one of the most useful things to have when debugging a failed e2e run.

Two options:
1. Just collect them as-is.
   These are dev-env secrets, not production credentials, and the artifacts are only accessible to people with repo access anyway.
2. If you want to be cautious, sanitize the secret data fields inside MW manifests before writing them out.
   Hash the values (SHA256 prefix or similar) so you can still tell whether two clusters got the same secret or not, without leaking the actual bytes.

Option 1 is simpler and probably fine.
Option 2 is more work but gives you useful debug info (same/different secret) without any leakage concern.

test/e2e/multi_primary_test.go
```diff
@@ -101,6 +101,17 @@ var _ = Describe("Multi-primary data plane", Ordered, Serial, func() {
 	}, NodeTimeout(5*time.Minute))
 
 	AfterAll(func(ctx SpecContext) {
+		dir := artifactDir("multi-primary")
+		Step("Collecting artifacts to %s", dir)
+		hubDir := filepath.Join(dir, "hub")
+		hubClient.CollectArtifacts(ctx, hubDir, meshNS, "multicluster-mesh-system")
```
mkolesnik: nit: `controllerNamespace` constant is available here:
```suggestion
		hubClient.CollectArtifacts(ctx, hubDir, meshNS, controllerNamespace)
```

test/e2e/suite_test.go
```diff
@@ -111,6 +112,16 @@ func Success(format string, args ...any) {
...
 
+func artifactDir(testName string) string {
+	base := os.Getenv("ARTIFACT_DIR")
+	if base == "" {
+		base = filepath.Join(os.TempDir(), "multicluster-mesh-e2e")
+	}
+	dir := filepath.Join(base, testName)
+	GinkgoWriter.Printf("Artifact directory: %s\n", dir)
+	return dir
+}
```
mkolesnik: If `ARTIFACT_DIR` is unset, nobody is going to look at the artifacts.
Collecting to `/tmp` just wastes time, especially in a local dev-env run.

Return early (or return empty string) when unset, and skip collection in the callers:

```go
func artifactDir(testName string) string {
	base := os.Getenv("ARTIFACT_DIR")
	if base == "" {
		return ""
	}
	dir := filepath.Join(base, testName)
	GinkgoWriter.Printf("Artifact directory: %s\n", dir)
	return dir
}
```

Then in AfterAll:
```go
if dir := artifactDir("mesh-lifecycle"); dir != "" {
	// collect artifacts
}
```

CI sets `ARTIFACT_DIR` in the workflow, so collection still happens there.
Locally, you skip it unless you explicitly opt in.

test/util/e2e_client.go
mkolesnik: Two issues here:

1. **Collection orchestration belongs in the test harness, not the client.**
   `E2EClient` methods should be straightforward "do one task" functions (like `ApplyFile`, `DeleteFile`, `Exec`).
   `CollectArtifacts` is a multi-step workflow (iterate namespaces, decide what to collect, manage directories, write files).
   That's test harness logic, not client logic.

2. **Reinventing pod/event listing via kubectl.**
   The embedded `client.Client` already has `List` for pods and events.
   No reason to shell out to kubectl for these.
   kubectl is only needed for operations the client API doesn't cover, like pod logs.

Suggested split:
- Add `SaveFile(ctx, dir, filename, resource string, ns ...string)` to `E2EClient`.
  Complements the existing `*File` pattern (`ApplyFile`, `DeleteFile`).
- Add `GetLogs(ctx, ns, pod string) ([]byte, error)` to `E2EClient` (or `SaveLogs` that writes directly).
- Use the embedded `client.Client` for pod and event listing.
- Move the collection orchestration (which namespaces, which resources, directory structure) to standalone functions in the test package or a separate collector type.

test/e2e/mesh_lifecycle_test.go
FilipB: MWs are collected too now and secrets are hashed with SHA256 prefix

test/e2e/mesh_lifecycle_test.go
FilipB: MWs are collected too now and secrets are hashed with SHA256 prefix

test/e2e/multi_primary_test.go
FilipB: Added

test/e2e/suite_test.go
FilipB: Done

test/util/e2e_client.go
FilipB: Done

FilipB:
Collected artifacts for both suites are visible in https://jenkins-csb-servicemesh-master.dno.corp.redhat.com/job/main-pipelines/job/acm-ossm-add-on-e2e/13/artifact/e2e-artifacts/

test/e2e/suite_test.go
```diff
@@ -132,6 +142,143 @@ func detectPlatform() {
 		platform, testOperatorNamespace, testOperatorName, testCatalogNamespace, testCatalogSource)
 }
 
+func artifactDir(testName string) string {
+	base := os.Getenv("ARTIFACT_DIR")
+	if base == "" {
+		base = filepath.Join(os.TempDir(), "multicluster-mesh-e2e")
```
sridhargaddam: The previous comment from @mkolesnik (which you marked as Done) was to return early when ARTIFACT_DIR is unset and skip collection, but the implementation does not seem to match the request. Is this intentional?

FilipB: Thank you for catching this. I was too quick with the rebase and did it incorrectly.

test/e2e/suite_test.go
```diff
@@ -132,6 +142,133 @@ func detectPlatform() {
...
+	if !ok {
+		return
+	}
+	for k, v := range m {
+		h := sha256.Sum256([]byte(fmt.Sprint(v)))
+		m[k] = "sha256:" + hex.EncodeToString(h[:])
+	}
+}
+
+func collectNamespaceArtifacts(ctx context.Context, c *util.E2EClient, dir string, namespaces []string) {
```
mkolesnik, line 202: Would it be useful to also dump full pod/deployment YAMLs here?

Pod summary gives name/phase/IP but when debugging you often need container specs, resource requests, volume mounts, restart counts, etc.

If you want to keep it minimal for now that's fine, just checking.

test/e2e/suite_test.go
```diff
@@ -132,6 +142,133 @@ func detectPlatform() {
...
+
+func hashStringValues(obj map[string]any, field string) {
+	m, ok := obj[field].(map[string]any)
+	if !ok {
+		return
+	}
+	for k, v := range m {
+		h := sha256.Sum256([]byte(fmt.Sprint(v)))
+		m[k] = "sha256:" + hex.EncodeToString(h[:])
+	}
```
mkolesnik: nit: you could unmarshal directly into `corev1.Secret` and check `s.Kind`.
Saves the `map[string]any` type assertion dance in `hashStringValues`:


```suggestion
		var s corev1.Secret
		if err := json.Unmarshal(raw, &s); err != nil || s.Kind != "Secret" {
			continue
		}
		for k, v := range s.Data {
			h := sha256.Sum256(v)
			s.StringData[k] = "sha256:" + hex.EncodeToString(h[:])
		}
		for k, v := range s.StringData {
			h := sha256.Sum256([]byte(v))
			s.StringData[k] = "sha256:" + hex.EncodeToString(h[:])
		}
		if updated, err := json.Marshal(s); err == nil {
			mw.Spec.Workload.Manifests[i].Raw = updated
		}
	}
}
```

mkolesnik (commented):
LGTM overall just one suggestion about dumping pod/deployment yamls and a nit to improve secret sanitization.

Let me know what you think?

test/e2e/suite_test.go
FilipB: Added in https://github.com/stolostron/multicluster-mesh-addon/pull/229/changes/674ba0e751a0ec4f1dc152c8a2ced90018e84f6a

FilipB:
@coderabbitai full review

mkolesnik (commented):
/lgtm
/unhold

FilipB:
Adding hold, running jenkins job against this PR to test it

FilipB:
Artifacts collected: https://jenkins-csb-servicemesh-master.dno.corp.redhat.com/job/main-pipelines/job/acm-ossm-add-on-e2e/46/artifact/e2e-artifacts/

PR #230, opened 2026-08-06
Auto-detect OCP vs kind platform for e2e test config

FilipB:
Use OCM-idiomatic product.open-cluster-management.io ClusterClaim from ManagedCluster status to detect a platform.

OpenShift: servicemeshoperator3 in sail-operator, redhat-operators catalog.
kind: sailoperator in sail-operator, operatorhubio-catalog.

### DIFF
--- a/test/e2e/suite_test.go
+++ b/test/e2e/suite_test.go
@@ -73,7 +73,7 @@ var _ = BeforeSuite(func(ctx context.Context) {
 	}
 
 	Step("Detecting platform (kind vs OCP)")
-	detectPlatform(ctx, spokeClients["cluster1"])
+	detectPlatform(ctx, hubClient, "cluster1")
 
 	Step("Verifying cluster connectivity")
 	verifyConnection(ctx, hubClient, "hub")
@@ -112,8 +112,8 @@ func Success(format string, args ...any) {
 	GinkgoWriter.Println("* " + fmt.Sprintf(format, args...))
 }
 
-func detectPlatform(ctx context.Context, spokeClient client.Client) {
-	cfg, err := util.DetectPlatform(ctx, spokeClient)
+func detectPlatform(ctx context.Context, hubClient client.Client, clusterName string) {
+	cfg, err := util.DetectPlatform(ctx, hubClient, clusterName)
 	Expect(err).NotTo(HaveOccurred(), "failed to detect platform")
 	GinkgoWriter.Printf("Detected platform: catalog=%s/%s, operator=%s/%s\n",
 		cfg.CatalogNamespace, cfg.CatalogSource, cfg.OperatorNamespace, cfg.OperatorName)
--- a/test/e2e/suite_test.go
+++ b/test/e2e/suite_test.go
@@ -27,19 +27,16 @@ import (
 	"github.com/stolostron/multicluster-mesh-addon/test/util"
 )
 
-const (
-	testOperatorName      = "sailoperator"
-	testOperatorNamespace = "sail-operator"
-)
-
 var (
 	clusters = []string{"cluster1", "cluster2"}
 
 	hubClient    *util.E2EClient
 	spokeClients map[string]*util.E2EClient
 
-	testCatalogSource    string
-	testCatalogNamespace string
+	testOperatorName      string
+	testOperatorNamespace string
+	testCatalogSource     string
+	testCatalogNamespace  string
 )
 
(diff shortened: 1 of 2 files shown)


test/e2e/suite_test.go
```diff
@@ -111,6 +112,17 @@ func Success(format string, args ...any) {
 	GinkgoWriter.Println("* " + fmt.Sprintf(format, args...))
 }
 
+func detectPlatform(ctx context.Context, spokeClient client.Client) {
```
sridhargaddam: While this works, I'm curious to know if you considered using the "cluster.Status.ClusterClaims" which is a standard way for the Hub to know the type of ManagedCluster?

The controller code used to depend on the clusterClaim, but was removed in one of the previous PRs - https://github.com/stolostron/multicluster-mesh-addon/pull/115

FilipB: It's better approach, thanks. Maybe in the future we will need to differentiate types of OpenShift and it would not be possible with original approach.

test/util/platform.go
```diff
@@ -0,0 +1,51 @@
...
+	}
+
+	if isOpenShift(cluster) {
+		return PlatformConfig{
+			OperatorName:      "servicemeshoperator3",
+			OperatorNamespace: "sail-operator",
+			CatalogSource:     "redhat-operators",
+			CatalogNamespace:  "openshift-marketplace",
+		}, nil
+	}
```
sridhargaddam: nit: The addon controller code assumes OpenShift platform by default with an option to override if required (like for KIND clusters). So this part of code is optional.

test/util/platform.go
```diff
@@ -0,0 +1,51 @@
...
+		OperatorName:      "sailoperator",
+		OperatorNamespace: "sail-operator",
+		CatalogSource:     "operatorhubio-catalog",
+		CatalogNamespace:  "olm",
+	}, nil
+}
+
+func isOpenShift(cluster *clusterv1.ManagedCluster) bool {
+	for _, claim := range cluster.Status.ClusterClaims {
+		if claim.Name == "product.open-cluster-management.io" {
```
sridhargaddam: Personally, I like this approach, but @mkolesnik had some reservations on this. Let`s hear from Mike if he is okay with this.

sridhargaddam (commented):
LGTM. I've added an "/hold" so that we can hear from @mkolesnik about the product claims.

test/util/platform.go
FilipB: IIUC it's required to set this explicitly as tests are using those vars later to check expected values.

test/e2e/suite_test.go
```diff
@@ -74,6 +72,9 @@ var _ = BeforeSuite(func(ctx context.Context) {
 		spokeClients[name] = util.NewE2EClient(clientFrom(kc), kc)
 	}
 
+	Step("Detecting platform (kind vs OCP)")
+	detectPlatform(ctx, hubClient, "cluster1")
+
```
mkolesnik, line 77: This whole approach (new file, API call, ClusterClaim parsing) is over-engineered for what it does: picking between two sets of constants.

The CRD defaults are already OCP-focused (`servicemeshoperator3`, `redhat-operators`, `openshift-marketplace`).
The Makefile already has `PLATFORM` wired through.

Keep the original kind constants, apply them only when needed:

```go
var _ = BeforeSuite(func(ctx context.Context) {
	if os.Getenv("PLATFORM") == "kind" {
		testOperatorName = "sailoperator"
		testCatalogSource = "operatorhubio-catalog"
		testCatalogNamespace = "olm"
	}
	// ...
})
```

The vars default to OCP values (matching the CRD defaults), kind overrides them.
No new file, no API calls, no ClusterClaim parsing.
`test/util/platform.go` can be dropped entirely.

FilipB: Tests are updated to use PLATFORM env variable but I think it's better to set var values explicitly even for openshift for clarity. Those values are later used for assertions. Also we have to update CI to correctly set PLATFORM env variable

FilipB:
I'm also running tests on OCP via jenkins job against this PR. Please keep the hold until the tests are finished.

FilipB:
Tests are passing on both Kind and OCP

FilipB:
@sridhargaddam @mkolesnik The PR still requires lgtm from one of you. Thank you!

PR #231, opened 2026-08-06
Clean up ManagedCluster network labels in e2e AfterAll

FilipB:
The multi-primary test adds topology.istio.io/network labels to ManagedClusters in BeforeAll but never removes them, causing failures when re-running against the same clusters.

### DIFF
--- a/test/e2e/multi_primary_test.go
+++ b/test/e2e/multi_primary_test.go
@@ -123,12 +123,18 @@ var _ = Describe("Multi-primary data plane", Ordered, Serial, func() {
 
 		Step("Removing network labels from ManagedClusters")
 		for cluster := range spokeClients {
-			mc := &clusterv1.ManagedCluster{}
-			if err := hubClient.Get(ctx, key.Of(cluster), mc); err != nil {
-				continue
-			}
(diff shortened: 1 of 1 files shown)


test/e2e/multi_primary_test.go
```diff
@@ -112,6 +113,28 @@ var _ = Describe("Multi-primary data plane", Ordered, Serial, func() {
 		Step("Deleting test mesh %s", meshName)
 		if mesh != nil {
 			_ = client.IgnoreNotFound(hubClient.Delete(ctx, mesh))
+
+			Step("Waiting for mesh %s to be fully removed", meshName)
+			Eventually(func() bool {
+				err := hubClient.Get(ctx, key.For(mesh), mesh)
+				return apierrors.IsNotFound(err)
+			}).WithTimeout(2 * time.Minute).WithPolling(2 * time.Second).Should(BeTrue())
```
sridhargaddam, line 121: nit: The default timeout at the suite level is 2 minutes, so its optional to explicitly specify that here, but I see that you want to override the polling interval.

sridhargaddam (approved):

